PR #48, opened 2022-05-03
Feature/select wrong

qgk75:
[GSK-38](https://giskard.youtrack.cloud/issue/GSK-38) Give the possibility to select wrong / correct / bordeline /custom examples


giskard-server/src/test/java/tablesaw/TableSawIT.java
```diff
@@ -0,0 +1,40 @@
...
+     * @throws Exception
+     */
+    @Test
+    void filterStream() throws Exception {
+        String location = Paths.get("src/test/bucket/my-dataset.csv").toAbsolutePath().toString();
+        InputStreamReader reader = new InputStreamReader(
+            new FileInputStream(location));
+
+        Table table = Table.read()
+            .usingOptions(CsvReadOptions.builder(reader));
```
andreybavt, line 32: FYI https://github.com/jtablesaw/tablesaw/issues/290

when you have an input stream as an input TableSaw stores it all in memory, but if you pass a `File` it doesn't

andreybavt: you can try running this test with something like 5GB csv

PR #79, opened 2022-06-07
scroll labels in prediction explanation starting from 5 labels

qgk75:
Scroll labels when more than 5


giskard-frontend/src/views/main/project/PredictionExplanations.vue
```diff
@@ -198,13 +198,15 @@ export default class PredictionExplanations extends Vue {
         textStyle: {
           fontSize: "10",
         },
+        type: this.classificationLabels.length > 5 ? 'scroll' : 'plain',
```
andreybavt, line 201: good finding of 'type' property 👍

andreybavt (approved):

PR #82, opened 2022-06-07
hotfix : Solve Inspector tab and feedbacks creation

qgk75:
use query params from router for the inspection page


giskard-frontend/src/router.ts
```diff
@@ -74,11 +74,11 @@ export default new Router({
                   props: (route) => { return {projectId: Number(route.params.id)} }
                 },
                 {
-                  path: 'inspection/:inspectionId',
+                  path: 'inspection',
                   name: 'project-inspector',
                   component: () => import('./views/main/project/InspectorWrapper.vue'),
                   props: route => ({
-                    inspectionId: Number(route.params.inspectionId)
+                    inspectionId: Number(route.query.inspectionId)
```
andreybavt, line 81: 🤔 what was a problem with passing inspection id as a path param?

qgk75: Started with that, the issue is the inspectionId param is mandatory for the router, and we don't have this id at first .
[vue-router] missing param for named route "project-inspector": Expected "inspectionId" to be defined
I can dig more if needed no pb

andreybavt: what do you mean "at first"? When we create an inspection we receive a response with the newly created inspection id back, so it's available.
To me the problem was really because of `to.path.endsWith('inspect')` that didn't catch the new path name `inspection`

PR #84, opened 2022-06-09
GSK-139 : solve UI project creation 

qgk75:
Copy the name for the key in project creation


andreybavt (approved):

giskard-frontend/src/views/main/project/ProjectsHome.vue
```diff
@@ -154,5 +162,10 @@ export default class ProjectsHome extends Vue {
     }
   }
 
+  @Watch("newProjectName", { deep: true })
```
andreybavt, line 165: no need for deep here

PR #194, opened 2022-08-02
GSK-232 target column not available validation added for performance …

princyiakov:
…test

## Description
Added validation for target column availability for performance tests 

## Type of Change

- [ x] 🥂 Improvement (non-breaking change which improves an existing feature)

### DIFF
--- a/giskard-ml-worker/ml_worker/testing/performance_tests.py
+++ b/giskard-ml-worker/ml_worker/testing/performance_tests.py
@@ -34,6 +34,8 @@ class PerformanceTests(AbstractTestCollection):
                 TRUE if AUC metrics > threshold
 
         """
+        if not actual_slice.target:
+            raise ValueError("Target column is not available")
         if len(model.classification_labels) == 2:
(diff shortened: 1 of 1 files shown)


giskard-ml-worker/ml_worker/testing/performance_tests.py
```diff
@@ -49,6 +49,8 @@ def test_auc(self, actual_slice: GiskardDataset, model: GiskardModel, threshold=
             ))
 
     def _test_classification_score(self, score_fn, gsk_dataset: GiskardDataset, model: GiskardModel, threshold=1):
+        if not gsk_dataset.target:
```
andreybavt, line 52: Could add the same check for the rest of the performance tests that are missing it:
- test_auc
- _test_regression_score

andreybavt (changes requested):

giskard-ml-worker/ml_worker/testing/performance_tests.py
princyiakov: done the needful

andreybavt (approved):

PR #249, opened 2022-09-19
GSK-236 Icon info tooltip name model dataset

thomasabou01:
Added an info icon near the filter button on the inspection page.
When hovered, it displays the name and the model of the dataset

Screenshot 2022-09-19 at 11 06 36

Screenshot 2022-09-19 at 11 06 16


andreybavt:
@thomasabou01 , there's a weird overlay of tooltip and left vertical bar if we remove padding. So I restored it back. No other comments on my side

andreybavt (approved):

PR #292, opened 2022-10-10
GSK-302 Add id information on the model, dataset and inspection pages.

thomasabou01:
Finished


andreybavt:
image

@thomasabou01 , a tooltip's default width isn't enough. Could you increase it?

andreybavt (changes requested):
see comment

andreybavt (approved):

PR #310, opened 2022-10-20
git container for testing

thomasabou01:
First push


giskard-git/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    && yarn global add git-http-server \
+    && adduser -D -g git git
+
+USER git
+WORKDIR /home/git
+
+RUN git init --bare repository.git
+
+RUN cd repository.git && git symbolic-ref HEAD refs/heads/main
+RUN git config --global user.email "giskard@example.com" && git config --global user.name "Giskard"
```
andreybavt, line 13: actually this container should only be responsible for bare repo, so no need for cloning it. We may need to clone it in the "backend" container since it's the backend who will read/modify files in the "bare" repo

andreybavt (changes requested):

andreybavt:
@kevinmessiaen , this is a small POC we did with Thomas to check how we could use internal git server. In the future we could use [git's http server](https://git-scm.com/docs/git-http-backend) instead of a third party one

PR #337, opened 2022-11-09
send error tests to the ml_worker

thomasabou01:
fronted: add icon for error tests
backend: return a testexecution even if there is an error.


giskard-frontend/src/views/main/tests/test-utils.ts
```diff
@@ -19,6 +19,8 @@ function testStatusToIcon(status: TestResult){
             return 'fa-check';
         case TestResult.FAILED:
             return 'fa-xmark';
+        case TestResult.ERROR:
+            return 'fa-exclamation-triangle';
```
andreybavt, line 23: @thomasabou01 I slightly modified the icon

andreybavt (changes requested):
@thomasabou01 
There are 2 issues:
- when you execute 1 test from test list and it fails you see an error icon ⚠️ but if you reload the test list page the failed tests don't have the icon anymore
- when you run a test from the test page the error isn't displayed

image

andreybavt (approved):

PR #377, opened 2022-11-29
Bug/gsk 355 get rid of lime html generation

thomasabou01:
I found a simplest way to correct this bug.
I found in the codebase of the eli5 package a function that returns a way simpler html representation of the results that doesn't use IPython. I checked on many examples, the final result is exactly the same.
To be sure that I corrected the bug, I tried to delete IPython from my ml-worker environment.
- Before the fix 
Screenshot 2022-11-29 at 17 15 38

- After the fix 
Screenshot 2022-11-29 at 17 16 13
 The only issue is that we have still not a lot of control on the output but I think this might be not a big problem, we could discuss about it

Thanks to that I had not to touch at the frontend

TODO:

- [ ] Slider to filter highlighted words based on threshold


andreybavt:
@thomasabou01 Could you get rid of the horizontal scrolling?
image

Another visual remarque, when changing the importance level the circular selector is cropped at the top
image

backend/src/main/java/ai/giskard/web/dto/ExplainTextResponseDTO.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@UIModel
+public class ExplainTextResponseDTO {
+    List<String> words = new ArrayList<String>();
```
andreybavt, line 20: you can skip type in the constructor if it's already specified in the variable type
```suggestion
    List<String> words = new ArrayList<>();
```

python-client/giskard/ml_worker/core/model_explanation.py
```diff
@@ -64,13 +64,41 @@ def explain_text(model: GiskardModel, input_df: pd.DataFrame,
         model.prediction_function, input_df, text_column
     )
     try:
-        text_explainer.fit(text_document, prediction_function)
-        return text_explainer.show_prediction(target_names=model.classification_labels)
+        for i in range(10):
```
andreybavt, line 67: could you extract 10 into a constant with a self explanatory name like `TEXT_EXPLAIN_ATTEMPTS` ?

frontend/src/views/main/project/TextExplanationParagraph.vue
```diff
@@ -0,0 +1,126 @@
...
+                label="Importance"
+            ></v-slider>
+        </div>
+        <VirtualCollection
+            :cellSizeAndPositionGetter="cellSizeAndPositionGetter"
+            :collection="items"
+            :height="200"
+            :width="widthParagraph"
+        >
+            <span slot="cell" slot-scope="props"  :style="{ backgroundColor: props.data.color }">{{props.data.text}}</span>
```
andreybavt, line 17: `slot-scope` is deprecated in Vue 2.7

andreybavt: `data` is not defined in `defineProps`

frontend/src/views/main/project/TextExplanationParagraph.vue
```diff
@@ -0,0 +1,126 @@
...
+        return color
+    }
+}
+
+
+onMounted(() => {
+    window.addEventListener('resize', () => {
+        x.value = 0;
+        y.value = 0;
+        widthParagraph.value = paragraph.value.clientWidth;
```
andreybavt: `paragraph.value` is possibly null

frontend/src/views/main/project/TextExplanationParagraph.vue
```diff
@@ -0,0 +1,126 @@
...
+        widthParagraph.value = paragraph.value.clientWidth; 
+        items.value = createItems()
+    });
+})
+
+// Called when mounted and when window resized 
+function createItems(){
+    return Array.from({length: props.words.length}, (_, index) => {
+        let c = document.createElement("canvas") as HTMLCanvasElement;
+        let widthText = getTextWidth(props.words[index], c) + 3;
```
andreybavt, line 92: could you extract `3` and `20` into variables with names like `wordHorizontalPadding` and `lineHeight`?

frontend/src/views/main/project/TextExplanationParagraph.vue
```diff
@@ -0,0 +1,126 @@
...
+        y.value = 0;
+        widthParagraph.value = paragraph.value.clientWidth; 
+        items.value = createItems()
+    });
+})
+
+// Called when mounted and when window resized 
+function createItems(){
+    return Array.from({length: props.words.length}, (_, index) => {
+        let c = document.createElement("canvas") as HTMLCanvasElement;
```
andreybavt, line 91: you can create a single global canvas and use it for all words instead of creating a new one for each word can't you? In this case you won't even need to pass it as an argument to `getTextWidth` as it can be taken from component level.

frontend/src/views/main/project/TextExplanationParagraph.vue
```diff
@@ -0,0 +1,126 @@
...
+            x: x.value,
+            y: y.value
+        }
+        x.value += widthText;
+        return res;
+    })
+}
+
+function getTextWidth(text, c : HTMLCanvasElement) {
+    var ctx = c.getContext("2d")!;
```
andreybavt, line 114: replace var by let

andreybavt (changes requested):

frontend/src/views/main/project/TextExplanationParagraph.vue
thomasabou01: The problem is that the createItems function is called before every element is mounted on the page so we can't access a canvas for example. I modified so that the canvas is created only once and not for every element but it has to stay in the createItems function.

andreybavt:
Custom LIME results UI

Issue #588, opened 2022-12-08
Project creation automatic slug uses dashes (-) which aren't accepted by the backend


Googleton:
## 🐛 Bug Report

When creating a new project, as I type the name of that new project, the autogenerated slug uses dashes, which aren't valid.
It should use underscores, which are.

## 🔬 How To Reproduce

1. Create a new project
2. Type in the name of a project with a space

Issue #432, opened 2022-12-05
Migrate vue component declaration from @Component to composition API


andreybavt:
[GSK-351](https://giskard.youtrack.cloud/issue/GSK-351)

### Migration:

`vue-property-decorator` library doesn't support Vue 3, so in order to migrate to Vue 3 we have to migrate component definition from @Component to composition API

* [ ] Admin.vue
* [x] AdminUsers.vue
* [x] App.vue
* [ ] AuthPortal.vue
* [x] ButtonModalConfirmation.vue
* [ ] CreateUser.vue
* [ ] CustomInspectionFilter.vue
* [x] Dashboard.vue
* [ ] DatasetSelector.vue
* [x] Datasets.vue
* [x] EditUser.vue
* [ ] EditableLabel.vue
* [ ] ErrorToast.vue
* [x] FeedbackDetail.vue
* [x] FeedbackList.vue
* [ ] FeedbackPopover.vue
* [ ] InspectionFilter.vue
* [ ] Inspector.vue
* [x] InspectorLauncher.vue
* [ ] InspectorWrapper.vue
* [x] InviteUsers.vue
* [x] Login.vue
* [ ] Main.vue
* [x] MessageReply.vue
* [ ] ModelSelector.vue
* [x] Models.vue
* [ ] MultiSelector.vue
* [x] OverlayLoader.vue
* [x] PasswordRecovery.vue
* [ ] PredictionExplanations.vue
* [ ] PredictionResults.vue
* [x] Project.vue
* [ ] ProjectOverview.vue
* [x] ProjectsHome.vue
* [x] ResetPassword.vue
* [ ] ResultPopover.vue
* [ ] RouterComponent.vue
* [x] Signup.vue
* [ ] Start.vue
* [ ] TestCreateModal.vue
* [ ] TestEditor.vue
* [ ] TestSuite.vue
* [ ] TestSuiteCreateModal.vue
* [ ] TestSuiteSettings.vue
* [ ] TestSuites.vue
* [ ] Tests.vue
* [x] TextExplanation.vue
* [ ] UploadButton.vue
* [x] UserProfile.vue
* [x] UserProfileEditPassword.vue

Project: Giskard
Reporter: aavtomonov
Created: 2022-11-13

PR #589, opened 2022-12-08
Use underscores instead of dashes in auto generated project slugs

Googleton:
## Description

- [x] Replace `-` with `_` in `toSlug`
- [x] Migrate ProjectsHome.vue to Composition API 

## Related Issue

#588 
#432 - ProjectsHome.vue migration

## Type of Change

### DIFF
--- a/frontend/src/views/main/project/ProjectsHome.vue
+++ b/frontend/src/views/main/project/ProjectsHome.vue
@@ -92,90 +92,96 @@
   </div>
 </template>
 
-<script lang="ts">
-import {Component, Vue, Watch} from 'vue-property-decorator';
-import {ValidationObserver} from 'vee-validate'
-import {readAllProjects, readHasAdminAccess, readUserProfile} from '@/store/main/getters';
-import {Role} from '@/enums';
-import {dispatchCreateProject, dispatchGetProjects} from '@/store/main/actions';
-import {ProjectPostDTO} from '@/generated-sources';
+<script setup lang="ts">
+import {computed, onMounted, ref, watch} from "vue";
+import {ValidationObserver} from "vee-validate";
+import {dispatchCreateProject, dispatchGetProjects} from "@/store/main/actions";
+import {readAllProjects, readHasAdminAccess, readUserProfile} from "@/store/main/getters";
+import {Role} from "@/enums";
+import {ProjectPostDTO} from "@/generated-sources";
+import {toSlug} from "@/utils";
+import store from "@/store";
+import {useRoute, useRouter} from "vue-router/composables";
 import moment from "moment";
-import {toSlug} from '@/utils';
 
-@Component
-export default class ProjectsHome extends Vue {
+const route = useRoute();
+const router = useRouter();
 
-  private openCreateDialog = false; // toggle for edit or create dialog
-  private newProjectName = "";
-  private newProjectKey = "";
-  private newProjectDesc = "";
-  private creatorFilter = 0;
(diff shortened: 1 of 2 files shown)


frontend/src/views/main/project/ProjectsHome.vue
```diff
@@ -92,90 +92,96 @@
...
-    dialogForm: InstanceType<typeof ValidationObserver>;
-  };
 
-  public async mounted() {
-    const f = this.$route.query.f ? this.$route.query.f[0] || "" : ""
-    this.creatorFilter = parseInt(f) || 0;
-    await this.loadProjects();
-  }
+// template ref
+const dialogForm = ref<ValidationObserver | null>(null);
```
andreybavt, line 119: This one is a bit more tricky since `ValidationObserver` is not a type but a component, so this ref should be declared as:

```
const dialogForm = ref<InstanceType<typeof ValidationObserver> | null>(null);
```

frontend/src/views/main/project/ProjectsHome.vue
```diff
@@ -92,90 +92,96 @@
...
+  creatorFilter.value = parseInt(f) || 0;
+  await loadProjects();
+})
 
-  get userProfile() {
-    return readUserProfile(this.$store);
-  }
+// computed
+const userProfile = computed(() => {
+  return readUserProfile(store);
```
andreybavt, line 129: `readUserProfile` returns `AdminUserDTO | null` so `userProfile` can be null too.
So either you declare computed as `computed<AdminUserDTO | null>...` or handle the null inside the computed method.
If we do declare computed as nullable, then in the template we'd have to add `?` everywhere where 
In the current context null user profile doesn't make sense. If you reached projects home page you there should be user profile already since the user is logged in. So I suggest we do something like this:

```

// computed
const userProfile = computed<AdminUserDTO>(() => {
  const userProfile = readUserProfile(store);
  if (userProfile === null){
    throw new Error("User is not defined");
  }
  return userProfile;
});
```

andreybavt (changes requested):

frontend/src/views/main/project/ProjectsHome.vue
```diff
@@ -92,90 +92,100 @@
...
+    clearAndCloseDialog();
+  } catch (e) {
+    console.error(e.message);
+    projectCreateError.value = e.message;
   }
-
 }
+
+// watchers
+watch(() => newProjectName.value, (value, oldValue, onCleanup) => {
```
andreybavt, line 187: nitpick: oldValue, onCleanup can be removed since they're not used

frontend/src/views/main/project/ProjectsHome.vue
```diff
@@ -92,90 +92,100 @@
...
-  }
+const isAdmin = computed(() => {
+  return readHasAdminAccess(store);
+});
 
-  public get isCreator() {
-    return this.userProfile?.roles?.includes(Role.AICREATOR);
-  }
+const isCreator = computed(() => {
+  return this.userProfile?.roles?.includes(Role.AICREATOR);
```
andreybavt, line 141: there's no `this` anymore, plus a computable is a `ref` so to refer to its value you'd need to use
`userProfile.value`:

```suggestion
  return userProfile.value.roles?.includes(Role.AICREATOR);
```

andreybavt (changes requested):
forgot 2 other points, they should be the last ones

andreybavt (approved):

Issue #393, opened 2022-12-05
test_diff* tests fail with zero division if there are no correct predictions for the dataset


andreybavt:
[GSK-391](https://giskard.youtrack.cloud/issue/GSK-391)

instead of throwing an exception a test should fail with a human-readable message

Project: Giskard
Reporter: aavtomonov
Created: 2022-11-29

PR #594, opened 2022-12-09
#393 Add always wrong model & data, make test fail if metric is invalid

Googleton:
## Description

- [x] Add `always_wrong_*` model and dataset to tests
- [x] Change error handling from exception to failing the test if `metric_1 == 0`

## TODO while draft

I do not like how the `always_wrong_model` is done. It uses a dummy classifier to get a dummy prediction function that returns "0", but the `GiskardModel` exported is marked as "regression" to avoid having to pass in a list of label types and classification labels. This feels hacky, but since it is a dummy model, I'm not sure how to better handle it.

Also, the always_wrong dataset has a "sex" column because the test was made with the german credit scoring dataset in mind. To avoid editing more code than necessary, I added it to the dataset but I feel like the labels over which we want to define the actual/reference slices could be specified via the test parameters?


python-client/tests/test_performance.py
```diff
@@ -233,7 +234,7 @@ def test_diff_accuracy(data, model, threshold, expected_metric, request):
     )
     assert round(result.metric, 2) == expected_metric
     assert type(result.output_df) is bytes
-    assert result.passed
+    assert result.passed == should_pass
```
andreybavt, line 237: it's worth also asserting messages since we know there should be a message about zero metric

python-client/tests/fixtures/always_wrong.py
```diff
@@ -0,0 +1,41 @@
...
+    return GiskardDataset(
+        df=df,
+        column_types=df.dtypes.apply(lambda x: x.name).to_dict(),
+        target="output",
+        feature_types=input_types
+    )
+
+
+@pytest.fixture()
+def always_wrong_model() -> GiskardModel:
```
andreybavt, line 28: I'd put a dummy model fixture to german_credit_scoring.py since "always wrong" is a classification concept, in regression there's no binary distinction between "wrong" and "right".
Also, since we're in the context of german_credit_scoring dataset, instead of `always_wrong` I'd call it `always_default`.
Then in your test you could pass a reference slice as:
```
reference_slice = data.slice(lambda df: df[(df.sex == "female") & (df.default == "Not default")])
```

in this case you'll know that the true value of every reference slice row is "Not default" but your dummy model always returns "Default" so the metric will be 0

python-client/tests/test_data/always_wrong.csv
```diff
@@ -0,0 +1,3 @@
+input,sex,output
```
andreybavt, line 1: with the comment above in place, you won't need a new dataset, you'll be able to reuse the standard german credit scoring one

python-client/tests/fixtures/german_credit_scoring.py
```diff
@@ -142,3 +144,18 @@ def german_credit_model(german_credit_data) -> GiskardModel:
...
+def german_credit_always_default_model() -> GiskardModel:
+    X = np.array([0])
+    y = np.array(["Default"])
+
+    dummy = DummyClassifier(strategy="constant", constant=np.array(["Default"]))
+    dummy.fit(X, y)
+
+    return GiskardModel(
+        prediction_function=dummy.predict,
+        model_type='regression',
```
andreybavt, line 159: this one is a 'classification' model


actually to build a dummy model that would behave simmilar to a `german_credit_model` but return 'Default' all the time you can just do more or less the same as the `german_credit_model` does including the X_train, Y_train creation:


```
@pytest.fixture()
def german_credit_always_default_model(german_credit_data) -> GiskardModel:
    Y = german_credit_data.df["default"]
    X = german_credit_data.df.drop(columns="default")

    dummy = DummyClassifier(strategy="constant", constant="Default")
    dummy.fit(X, Y)

    return GiskardModel(
        prediction_function=dummy.predict_proba,
        model_type="classification",
        feature_names=list(input_types),
        classification_threshold=0.5,
        classification_labels=dummy.classes_,
    )

```

Like this your dummy model will be exactly like a "normal" one except that it'll return Default label all the time

python-client/tests/test_performance.py
```diff
@@ -236,6 +236,25 @@ def test_diff_accuracy(data, model, threshold, expected_metric, request):
...
+    result = tests.performance.test_diff_accuracy(
+        actual_slice=data.slice(lambda df: df[df.sex == "male"]),
+        reference_slice=data.slice(lambda df: df[(df.sex == "female") & (df.default == "Not default")]),
+        model=request.getfixturevalue(model),
+        threshold=threshold,
+    )
+    assert round(result.metric, 2) == expected_metric
+    assert type(result.output_df) is bytes
+    assert not result.passed
+    assert result.messages
```
andreybavt, line 255: being precise in the assertions helps with increasing test sensitivity. Here we'd need to assert the actual message too (and maybe the size of the messages array).
you can use `re` module to assert the message content with `re.match`

Also could you test the rest of the `test_diff_*` functions in the same way? Since they have the same signature, you can run tests in a loop like:

```
    for test_fn in [
        tests.performance.test_diff_accuracy,
        tests.performance.test_diff_f1,
        tests.performance.test_diff_precision,
        tests.performance.test_diff_recall,
        tests.performance.test_diff_reference_actual_f1
    ] :
        result = test_fn(
            actual_slice=data.slice(lambda df: df[df.sex == "male"]),
            reference_slice=data.slice(lambda df: df[(df.sex == "female") & (df.default == "Not default")]),
            model=request.getfixturevalue(model),
            threshold=threshold,
        )
     ...   
        
```

or maybe even parameterize a test name, but you'd have to then convert a string with a function name to an actual test function before running it

andreybavt (changes requested):

andreybavt (approved):

PR #606, opened 2022-12-14
Slice management: Create slices in UI and apply them over inspections

Googleton:
## Description

Slices are now in Giskard!

The UI is currently only located inside of the Inspector. This is what it looks like:
Screenshot 2023-01-10 at 10 37 11
Screenshot 2023-01-10 at 10 37 29
Screenshot 2023-01-10 at 10 37 45
Screenshot 2023-01-10 at 10 38 29


backend/src/main/java/ai/giskard/service/ml/MLWorkerService.java
```diff
@@ -119,6 +124,44 @@ public synchronized UploadStatus upload(MLWorkerClient client, ProjectFile file)
         return result.get();
     }
 
+    public synchronized FilterDatasetResponse filterDataset(MLWorkerClient client, Dataset dataset, String function) throws IOException {
```
andreybavt, line 127: TODO: remove `synchronized`

backend/src/main/java/ai/giskard/web/rest/controllers/SliceController.java
```diff
@@ -0,0 +1,54 @@
...
+    @Transactional
+    @PreAuthorize("@permissionEvaluator.canWriteProject(#dto.projectId)")
+    public SliceDTO updateSlice(@Valid @RequestBody SlicePutDTO dto) {
+        return giskardMapper.sliceToSliceDTO(sliceService.updateSlice(dto));
+    }
+
+
+    @DeleteMapping("slices/{sliceId}")
+    @Transactional
+    public void deleteSlice(@PathVariable @NotNull Long sliceId) {
```
kevinmessiaen, line 50: I think that we should add `@PreAuthorize("@permissionEvaluator.canWriteProject(#dto.projectId)")` to only allow authorised users to delete the slice

frontend/src/components/SliceDropdown.vue
```diff
@@ -0,0 +1,189 @@
...
+
+const autocomplete = ref<any | null>(null);
+
+onMounted(() => {
+  loadSlices();
+})
+
+const items = computed(() => {
+  // TODO: Find a solution for this typing issue
+  let result: any[] = slices.value.map(slice => ({
```
kevinmessiaen, line 107: Maybe have a type `AutocompleteItem` defined as following

```
{
  text: string,
  name: string,
  index: number | null
}
```

For each item in the `slices` array have either the index in the array or ID of the slice to retrieve it latter on and for the `__NEW_SLICE__` item have `-1`

I think it would be cleaner.

andreybavt:
There's a buggy behaviour when switching from less selective slice to more selective one when we're outside of a new range:

https://user-images.githubusercontent.com/7889985/211676018-2179da0a-c2c5-4a72-a3b1-692398007c73.mov

andreybavt:
Slice gets unselected after being edited:

https://user-images.githubusercontent.com/7889985/211676385-c050c00d-d93a-49ed-9656-aff27c7f6fb5.mov

andreybavt:
Slices change order in the dropdown. I had `test` first and `all` after, then modified `test` and it became second
image

Should we sort them by name?

backend/src/main/resources/config/liquibase/changelog/20221215151953_changelog.xml
```diff
@@ -0,0 +1,37 @@
...
+            <column name="created_date" type="TIMESTAMP"/>
+            <column name="last_modified_by" type="VARCHAR(50)"/>
+            <column name="last_modified_date" type="TIMESTAMP"/>
+            <column name="code" type="VARCHAR(255)"/>
+            <column name="name" type="VARCHAR(255)"/>
+            <column name="project_id" type="BIGINT"/>
+        </createTable>
+    </changeSet>
+    <changeSet author="hugoviallon (generated)" id="1671113997239-6">
+        <addUniqueConstraint columnNames="key" constraintName="UC_PROJECTSKEY_COL" tableName="projects"/>
```
andreybavt, line 20: watch out for the non related changesets that get generated automatically. I'm not entirely sure why liquibase tries to drop `projects` constraints and recreate them, but these changes don't make any sense here. Generally speaking you should go though the generated xml to validate it manually (liquibase suggests it too)

backend/src/main/java/ai/giskard/service/ml/MLWorkerService.java
```diff
@@ -16,26 +17,34 @@
...
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
```
andreybavt, line 36: unused import

Googleton:
Thanks for the functional feedback, it was an oversight on my part.

I've solved the first 2 by:
- Resetting the dataset row number to 0 (well, 1) when applying a slice
- Selecting the new edited slice instance if it was selected previously

As for the sorting, I'm open to any solution (name or creation date). Sorting by name would still cause a move when editing the name, which we may want to avoid ?

I'll take a look at the code feedback tomorrow :)

andreybavt:
Another general feedback: let's have a friendly message when you apply a filter that's too strict so that there are no matching rows instead of showing empty forms.

image

P.S.
no need to do it at midnight ;)

andreybavt:
1. Let's limit the max height of a name element in the slice creation modal so that the update/cancel buttons don't disappear

image


2. Could you remove the autocomplete (control+space feature)from the code editor that results into a black horizontal line? Have a look at the monaco editor on test creation screen. Also I don't think we need an overview widget neither since the slicing functions will usually be small.

image

3. Minor improvement: could you add autofocus on "name" field when the modal opens? It could also be more intelligent and focus on `name` when you create a slice and `code` when you edit it

4. I'd make a distinction between an "Edit slice" modal title and the slice name itself, maybe print the slice name on a new line?
image

5. Long slice names break saving operation. There could be a character limit on the UI and validation on the backend to prevent it.
image

6. When playing with the feature I found it a bit long to navigate to slice editing, since I had to adapt my slice function frequently and see the results. I think we could remove 1 click by placing edit/delete icons next to the slice name in the list. Maybe in order to make the UI lighter we could only show these buttons for the slice that you've currently hovering over
image


## General comment

I feel that it's a bit cumbersome to iterate while creating a slicing function. My workflow was:
- write a simple function
- check that it works (no syntax error, the resulting rows are the ones that I expected)
- modify the condition to make it more sophisticated

on each step, I had to close the modal and reopen it again. It's a bit disappointing, especially if there's a silly syntax error. 

It could be great if we could quickly validate the slice function with a chosen dataset without leaving the editing mode.That was the idea behind the "Run" button on [this mockup](https://www.notion.so/giskard/Slice-management-8e047ca91b6240bea699301f239fcbfc#51c7a441d4e54229b19a19eb69dd48b7)


Secondly, I might have missed some discussions last week, but I think slice creation being part of an inspection makes slices too hidden for users. It also unnecessarily requires a model to exist (without which you won't be able to create an inspection). In principle, a slice function should only require a dataset, so I think the previous location of the slice functions list (next to datasets) made more sense.

Googleton:
Hey Andrey, thanks for the feedback again.

I'll reply here to some of the points you mentionned:

4. Modal titles will be modified to "Editing slice" and "Creating slice", and the name will not be displayed. Since the name field is directly underneath, it just doubles the same information so it can be left out.

6. I agree with this too, using it yesterday made me realise it's not a good workflow currently. 
I believe we could have a "Run" or "Validate" button that checks the code against a dataset, but I think having a small pen icon next to the "X" in the slice dropdown would allow for quicker access to editing it. 
This way, a user will work on their code, validate it. When they want to test to see how many rows return, they can save. But then they have to open the dropdown, open the menu, open the modal.. Having a shortcut to the modal for the current slice would be nicer IMO.



As for your general feedback, we discussed it with Jean Marie and since Slices are (for now) only relevant for inspections we decided to put them there. With a future version having a test catalogue, we imagined we could have a "function catalogue" that works in a similar way, with slicing functions, perturbation functions and other types as the product evolves. 

There is an in-between then and now that we need to figure out if the dropdown solution isn't good enough for now, that's for sure

backend/src/main/java/ai/giskard/web/rest/controllers/InspectionController.java
```diff
@@ -36,6 +38,8 @@ public class InspectionController {
     private final InspectionRepository inspectionRepository;
     private final GiskardMapper giskardMapper;
     private final PermissionEvaluator permissionEvaluator;
+    private final SliceRepository sliceRepository;
```
andreybavt, line 41: not used

backend/src/main/java/ai/giskard/web/rest/controllers/InspectionController.java
```diff
@@ -74,6 +78,32 @@ public JsonNode getRowsFiltered(@PathVariable @NotNull Long inspectionId, @Reque
...
+    /**
+     * This function asks for a slice ID and an inspection ID and returns the list of rows filtered thru that slice.
+     * It streams data to the ML worker and is a bit more complex in its execution
+     * @param inspectionId
+     * @param sliceId
+     * @return
+     */
+    @PostMapping("/inspection/{inspectionId}/slice/{sliceId}")
+    @Transactional
+    public JsonNode getRowsSliced(@PathVariable @NotNull Long inspectionId, @PathVariable @NotNull Long sliceId) throws IOException {
```
andreybavt, line 90: Instead of adding a new method I think we should add sliceId to the existing `getRowsFiltered` otherwise slicing + filtering doesn't work

andreybavt (changes requested):

andreybavt:
> ack, we discussed it with Jean Marie and since Slices are (for now) only relevant for inspections we decided to put them there. With a future version having a test catalogue, we imagined we could have a "function catalogue" that works in a similar

I agree with both solutions you proposed for 4 and 6. As for the definitive place for the slices list, lets keep it inside inspection for now, but keep the "General comment" in mind for the moment when we start using slices in tests too.

I'd still enhance the slice editor to be able to try to run it on a dataset directly without leaving the editor

backend/src/main/java/ai/giskard/service/DatasetService.java
```diff
@@ -19,6 +21,7 @@
 import tech.tablesaw.io.csv.CsvReadOptions;
 
 import javax.validation.constraints.NotNull;
+import java.io.IOException;
```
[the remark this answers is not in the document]
Googleton: I see a lot of unused imports in my code, I'm used to having my editor do it for me, not sure why it's not applying automatically...

Is there a recommended code formatter for Giskard? I used google-java-format previously in a git precommit hook which helped keep all pushed code clean

backend/src/main/java/ai/giskard/domain/ml/Slice.java
```diff
@@ -0,0 +1,30 @@
...
+    private String name;
+
+    // Later on this will probably depend on the slice type: From a file, from the interface, from the API ?
+
+    @NotNull
+    private String code;
+
+    @ManyToOne
+    @NotNull
+    private Project project;
```
andreybavt, line 29: There needs to be a back reference in Project.java with `cascade = CascadeType.ALL` so that the project deletion works

andreybavt: like for ai.giskard.domain.Project#testSuites

Googleton:
Small todo for myself, increase slice name length limit, too restrictive atm

andreybavt:
@Googleton , could you have a last look at Sonar results? I don't think everything's coming from your changes though...

PR #617, opened 2022-12-21
Bug/change error message external ml worker not connected

thomasabou01:
quick fix


andreybavt:
@thomasabou01 , could you get rid of unrelated H2O and Tensorflow tests in this PR?

andreybavt (approved):

Issue #614, opened 2022-12-20
Remove user access management from open-source version


alexcombessie:

Googleton:
Changes to settings UI. Data removed from the User page:
<img width="1512" alt="Screenshot 2023-01-03 at 10 16 43" src="https://user-images.githubusercontent.com/1703741/210328828-32dfc6a3-59a0-4743-b86b-f4a0d259c8cf.png">

And added tabs to the settings tab, with General containing the missing info: 
<img width="1512" alt="Screenshot 2023-01-03 at 10 18 24" src="https://user-images.githubusercontent.com/1703741/210329103-df7da5f7-143d-413f-91e0-db6f1d77343f.png">

PR #635, opened 2023-01-02
Remove user access management from open source version

Googleton:
## Description

- [x] Migrate Vuex => Pinia
- [ ] Add proper feature management 
- [x] Rework the settings view to better split between user settings and general settings
- [x] Make the frontend use feature flags to hide user management entirely

## Related Issue

#614 
#432 ==> Components Migrated: 
- PasswordRecovery.vue
- ResetPassword.vue
- Signup.vue
- Dashboard.vue
- AdminUsers.vue
- CreateUser.vue
- EditUser.vue
- InviteUsers.vue
- UserProfile.vue
- UserProfileEditPassword.vue
- Project.vue

## Type of Change

- [ ] 🔐 Security fix


backend/src/main/java/ai/giskard/security/ee/NoAuthConfigurer.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private final UserRepository userRepository;
+
+    public NoAuthConfigurer(UserRepository userRepository) {
+        this.userRepository = userRepository;
+    }
+
+    @Override
+    public void configure(HttpSecurity http) {
+        User admin = this.userRepository.getOneByLogin("admin");
```
andreybavt, line 20: The problem is that `NoAuthConfigurer` needs to be created in order to start the backend, but `InitService` that creates default users (including `admin`) is executed once the app starts. So it's a chicken and egg problem that you can reproduce if you remove giskard-home and start from scratch. 
We shouldn't call `userRepository` in the constructor, rather at the incoming request time (`doFilter`)

andreybavt (changes requested):

backend/src/main/java/ai/giskard/security/ee/NoAuthConfigurer.java
Googleton: As we discussed => Try to use the login instead of the ID in most cases. This removes the issue altogether

andreybavt:
TODO: remove dashboard and arrive directly on projects screen when opening Giskard

frontend/src/views/main/profile/UserProfile.vue
```diff
@@ -97,373 +97,87 @@
...
+      resetFormData();
     }
-    this.appSettings!.generalSettings = await api.saveGeneralSettings(settings);
-  }
+  });
+}
 
-  private epochToDate(epoch: number) {
-    return moment.unix(epoch).format('DD/MM/YYYY HH:mm:ss');
+async function generateToken() {
```
andreybavt, line 163: we lost the possibility to generate API tokens in the free version since UserProfile vue isn't available

andreybavt (changes requested):

backend/build.gradle.kts
```diff
@@ -191,6 +191,7 @@ dependencies {
     }
     implementation("org.springframework.boot:spring-boot-starter-undertow")
     implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
+    implementation("org.springframework.boot:spring-boot-starter-webflux")
```
andreybavt, line 194: @Googleton , what part of webflux did you use?

andreybavt:
Could you have a look at sonar results too?

andreybavt:
An unhandled corner case:
- enable enterprise mode
- delete admin user
- switch to free mode
- no user settings are available

image

backend/build.gradle.kts
Googleton: I used it early on while testing something but forgot to remove it. Will fix!

Googleton:
So this issue actually happens even with authentication enabled. If you delete the admin user while logged in as admin and refresh, it will make the settings button disappear.

I think these checks should be in place:

Either:
A. A user cannot delete itself
B. A user can delete itself but also gets logged out. If there is only a single user, deletion cannot happen (we can never have no users)

I think A is the best scenario because it prevents Admin deleting the only Admin user. To delete admin with the A scenario, you would first need to create another admin user. 

And in any case, upon startup, InitService should check for the Auth state and if it's disabled, ensure the 'admin' user exists (and if not create it).

andreybavt:
@Googleton, I agree with 

> And in any case, upon startup, InitService should check for the Auth state and if it's disabled, ensure the 'admin' user exists (and if not create it).

Currently we don't remove users from the DB, we disable them, but to handle all of the possible cases in the "free" version we should ensure that the "admin" user exists and it's enabled and if not create/enable it at startup.

As for the "enterprise" user management, I think we should combine both A and B, by applying the following logic:
- If the current user is the only admin, he can't delete itself without assigning admin rights to some other user
- when a user deletes itself we should redirect him to the login page
- as a consequence of the first point, if you end up removing all of the users except for one, the last one must be an admin => he can't delete himself without creating another user and assigning admin rights to him

PR #639, opened 2023-01-08
Task/export import project

thomasabou01:


thomasabou01:
To export a project there is this new button.
It will download a zip file which name is the the key of the project.
It will also add a metadata folder in `giskard-home/projects/{project_key}`

Screenshot 2023-01-09 at 19 44 32

To import a project, there is this new button at the bottom right. When hovered, it shows 2 buttons. One **NEW** and one **IMPORT**

Screenshot 2023-01-09 at 19 47 44

The default name of the project will be the filename. (the frontend ensure that there is only alphanum and _)
Screenshot 2023-01-09 at 19 49 14

backend/src/main/java/ai/giskard/service/ProjectService.java
```diff
@@ -93,7 +123,124 @@ public void delete(Long id) {
...
+
+        return Files.readAllBytes(zipPath);
+    }
+
+
+    public Project importProject(MultipartFile zipMultipartFile, String userName) throws IOException {
+        Path pathToTmp = locationService.resolvedTmpPath();
+        String projectKey;
+        try{
+            projectKey = zipMultipartFile.getOriginalFilename().split("[.]")[0];
```
andreybavt, line 164: it's best to read project key from a project.yaml since users may rename the zip file itself

andreybavt (changes requested):

backend/src/main/java/ai/giskard/service/FileUploadService.java
```diff
@@ -40,21 +40,20 @@ public class FileUploadService {
...
-        Project project = projectRepository.getOneByKey(modelParams.getProjectKey());
-        Path projectModelsPath = locationService.modelsDirectory(project.getKey());
-        createOrEnsureOutputDirectory(projectModelsPath);
-
         ProjectModel model = createProjectModel(modelParams);
+        return uploadModel2(model, modelStream, requirementsStream);
+    }
 
-        // first save to get ID that will be used in the filenames
+    @Transactional public ProjectModel uploadModel2(ProjectModel model, InputStream modelStream, InputStream requirementsStream) throws IOException{
```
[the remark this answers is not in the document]
thomasabou01: oh yes, I forgot to change the name while developing ... :/

andreybavt:
image
We should actually allow user to map an old project key to a new one in case we detect a conflict

andreybavt:
We should also open an imported project after it's successful import

backend/src/main/java/ai/giskard/service/ProjectService.java
```diff
@@ -93,7 +127,134 @@ public void delete(Long id) {
...
+                 reply.setFeedback(savedFeedback);
+                 feedbackReplyRepository.save(reply);
+             });
+             feedbackRepository.save(savedFeedback);
+         });
+
+         testSuites.forEach(testSuite -> {
+             testSuite.setProject(savedProject);
+             testSuite.setActualDataset(datasetRepository.getById(mapFormerNewIdModelDataset.get(testSuite.getActualDataset().getId())));
+             testSuite.setReferenceDataset(datasetRepository.getById(mapFormerNewIdModelDataset.get(testSuite.getReferenceDataset().getId())));
```
andreybavt, line 247: `getReferenceDataset()` and `getActualDataset()` can return null resulting into NullPointerException

frontend/src/views/main/project/Project.vue
```diff
@@ -209,6 +213,17 @@ export default class Project extends Vue {
 		}
 	}
 
+	public async exportProject(id: number){
+		api.exportProjet(id).then((response) => {
+		const url = window.URL.createObjectURL(new Blob([response]))
```
andreybavt, line 218: there's already a method `downloadURL` in api.ts

andreybavt (changes requested):

backend/src/main/java/ai/giskard/web/rest/controllers/PrepareImportProjectDTO.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.Set;
+
+@UIModel
+@Getter
+@Setter
+@AllArgsConstructor
+public class PrepareImportProjectDTO {
```
Googleton, line 14: This should be in ai.giskard.web.dto instead of the controllers folder

Googleton:
Minor, but alignment here is not correct
Screenshot 2023-01-23 at 13 56 32

Googleton (commented):
Nothing left to say on my end 👍 

Just need to fix the sonarcloud issues

backend/src/main/java/ai/giskard/service/ProjectService.java
```diff
@@ -93,7 +120,173 @@ public void delete(Long id) {
...
+        List<Feedback> feedbacks = mapper.readValue(locationService.resolvedMetadataPath(pathToTimestampDirectory, Feedback.class.getSimpleName()).toFile(), new TypeReference<>(){} );
+        List<TestSuite> testSuites = mapper.readValue(locationService.resolvedMetadataPath(pathToTimestampDirectory, TestSuite.class.getSimpleName()).toFile(), new TypeReference<>(){} );
 
+        Map<Long, Long> mapFormerNewIdModelDataset = new HashMap<>();
+
+        project.setOwner(userRepository.getOneByLogin(userNameOwner));
+        project.setKey(projectKey);
+        Set<User> guestList = new HashSet<>();
+        importedUsersToCurrent.forEach((key, value) -> {
+            if (!value.equals(userNameOwner) && !value.equals("admin"))
```
andreybavt, line 219: instead of comparing to "admin" you should rather check if the user has admin right (check SecurityUtils), because the login can be different

backend/src/main/java/ai/giskard/web/rest/controllers/ProjectController.java
```diff
@@ -132,4 +142,29 @@ public ProjectDTO invite(@PathVariable("id") Long id, @PathVariable("userId") Lo
...
+    @PostMapping(value = "/project/import/{timestampDirectory}/{projectKey}")
+    @Transactional
+    public ProjectDTO importProject(@RequestBody Map<String, String> mappedUsers, @PathVariable("projectKey") @NotNull String projectKey, @PathVariable("timestampDirectory") @NotNull String timestampDirectory, @AuthenticationPrincipal final UserDetails userDetails) throws IOException {
+        permissionEvaluator.canWrite();
+        try {
+            return projectService.importProject(mappedUsers, timestampDirectory, projectKey, userDetails.getUsername());
+        } catch (IOException e){
+            throw new GiskardRuntimeException("Error while importing the project", e);
+        }finally {
+            FileSystemUtils.deleteRecursively(locationService.resolvedTmpPath());
```
andreybavt, line 167: the temporary directory cleaning should be part of the service, not the controller. Ideally controllers should be responsible for triggering actions and formatting responses according to the chosen format (DTO serialized to JSON in our case)

andreybavt (changes requested):

backend/src/main/java/ai/giskard/service/ProjectService.java
```diff
@@ -93,7 +120,173 @@ public void delete(Long id) {
...
+        if (!Files.exists(pathToTimestampDirectory)){
+            ZipUtils.unzipProjectFile(zipFile, pathToTimestampDirectory);
+        }
+        return pathToTimestampDirectory;
+    }
+
+    public PrepareImportProjectDTO prepareImport(Path pathToTimestampDirectory, String projectKey) throws IOException{
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory())
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        if (projectKey.equals("null")){
```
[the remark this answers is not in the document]
thomasabou01: TBH I didn't found a safe way to do this

backend/src/main/java/ai/giskard/web/rest/controllers/ProjectController.java
thomasabou01: my change is better now or still not what you expect ?

backend/src/main/java/ai/giskard/web/rest/controllers/ProjectController.java
```diff
@@ -132,4 +140,19 @@ public ProjectDTO invite(@PathVariable("id") Long id, @PathVariable("userId") Lo
...
 
+    @PostMapping(value = "/project/import/{timestampDirectory}/{projectKey}/prepare", consumes = "multipart/form-data")
+    public PrepareImportProjectDTO prepareImport(@RequestParam("file") MultipartFile zipFile, @PathVariable("timestampDirectory") @NotNull String timestampDirectory, @PathVariable("projectKey") String projectKey, @AuthenticationPrincipal final UserDetails userDetails) throws IOException {
+        Path pathToTimestampDirectory;
+        permissionEvaluator.canWrite();
+        pathToTimestampDirectory = projectService.unzip(timestampDirectory, zipFile);
+        return projectService.prepareImport(pathToTimestampDirectory, projectKey);
+    }
+
+    @PostMapping(value = "/project/import/{timestampDirectory}/{projectKey}")
```
andreybavt, line 151: After having another look at this code, I find it cleaner if frontend didn't have to be responsible for sending neither `timestampDirectory` nor `projectKey`. What we should only get from a user is a zip archive, for the rest:
- temporary directory name should be generated on the backend and sent to the frontend since it's the backend's responsibility to create that directory
- project key is taken from the archive directly and sent back to the frontend to autofill the project key text input

backend/src/main/java/ai/giskard/service/ProjectService.java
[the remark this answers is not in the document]
andreybavt: I left a comment about it in the Controller. The downside of the current approach is that frontend is responsible for something it shouldn't be responsible for - chosing a temporary directory name and "faking" a project key when it's no yet available. As an indicator - we can't use "null" as a project key, even if it's a corner case.

backend/src/main/java/ai/giskard/service/ProjectService.java
andreybavt: It's the `value` user that should be checked for admin rights, not the current user, isn't it?

backend/src/main/java/ai/giskard/web/rest/controllers/ProjectController.java
andreybavt: in JAVA it's not typical to use the `__` method prefix. Private methods should just have a `private` modifier. Plus I guess you could have a feeling that the original `importProject` method was too large to be wrapped with additional try-catch-finally. A better solution instead of wrapping one `importProject` with another would be to only keep a single `importProject` with `try` block but extract each entity creation into a separate submethod like `importModels`, `importDatasets`. By doing so you'll also keep the main `importProject` method on a higher level of abstraction than the methods it'll be calling which is a good practice.

andreybavt (changes requested):
see comments

andreybavt (approved):

PR #640, opened 2023-01-09
436 users who are ai creator and ai tester do not have the rights to create a test suite

kevinmessiaen:
## Description

Allow guests AI creators and AI testers to create testing suites

## Related Issue

N/A

## Type of Change

- [ ] 🔐 Security fix

## Checklist


backend/src/main/java/ai/giskard/security/PermissionEvaluator.java
```diff
@@ -26,14 +28,18 @@ public boolean isCurrentUser(String login) {
     }
 
     /**
-     * Determine if a user can write a project, i.e. is admin or project's owner
+     * Determine if a user can write a project, i.e. is admin, is project's owner or is guest with any role
```
Googleton, line 31: This is not correct and grants too many permissions to guests

"canWriteProject" is a very global setting and impact way more than test suites. This change effectively gives admin rights over a project to any guest inside it.

The correct change in this case would be to have a specific check on test suite creation. In that place, we would make sure that the guest user has either the AI Tester or AI Creator role

Googleton (changes requested):

Googleton (commented):
This is good now.

I'm not a huge fan of having such a complex condition in the `PreAuthorize` annotation but I'm not used to spring so maybe this is fine. I would tend to prefer a utility method.

@andreybavt do you think the condition is fine as it is now ? If yes, then I'm fine with merging this

Googleton (approved):

PR #652, opened 2023-01-10
Task/feature name change

rabah-khalek:
Father PR: https://github.com/Giskard-AI/giskard/pull/618

- [x] Changed the name of `feature_types` into `column_meanings`
- [x] remove `column_types` as input to `DatasetMeta`


andreybavt (approved):

backend/src/main/resources/config/liquibase/changelog/initial_changelog.xml
```diff
@@ -18,7 +18,7 @@
             <column name="size" type="BIGINT">
                 <constraints nullable="false"/>
             </column>
-            <column name="feature_types" type="VARCHAR"/>
+            <column name="column_meanings" type="VARCHAR"/>
```
andreybavt, line 21: you shouldn't be modifying existing database patches as they are already applied on users' servers. We have to generate a new patch instead. you can do it with `../gradlew liquibaseCreateDiff` from `backend` directory

andreybavt (changes requested):
I left a few comments. I'll generate a liquibase migration file myself and push it

PR #662, opened 2023-01-12
Bug/target none

rabah-khalek:
daughter of https://github.com/Giskard-AI/giskard/pull/618 and https://github.com/Giskard-AI/giskard/pull/661

if `ds.target=None` or if `ds.target not in ds.df.columns` in:
```python
    if ds.target in ds.df.columns:
        df = ds.df.drop(ds.target, axis=1)

    df_columns = set(df.columns)
    columns_with_types.discard(ds.target)
```
the `df_columns = set(df.columns)` will trigger an error as `df` won't be defined. The solution is:
```python
    df = ds.df
    if ds.target in ds.df.columns:
        df = ds.df.drop(ds.target, axis=1)

    df_columns = set(df.columns)
    columns_with_types.discard(ds.target)
```


andreybavt:
@rabah-khalek did you push everything you wanted to this branch? There's only 1 changed file with a single row modified

rabah-khalek:
Yes, I did.

andreybavt (dismissed):

PR #666, opened 2023-01-16
encrypt data sent from/to ML Worker

andreybavt:
External worker - ML Worker process diagram
https://miro.com/app/board/uXjVPd1XXws=/?moveToWidget=3458764543811138917&cot=14

- [x] moved from TCP sockets to UNIX socket and Netty's LocalAddress (IPS mechanism) to make it less accessible from other processes on local machines
- [x] Added AES GCM encryption with encryption key passed through a regular http**s** API call. 
- [x] Reading ML Worker port from remote server, not specifying it manually


andreybavt:
@Googleton , If you have time to check out this PR it'd be great, it can be also a chance for me to share knowledge about the Backend <--> ML Worker data flow

Googleton:
I've actually spent yesterday afternoon on it, we can discuss it whenever
you're available

Le mer. 25 janv. 2023, 23:59, Andrey Avtomonov ***@***.***> a
écrit :

Googleton:
Internal ML Worker doesn't work currently, looks like some flag checks are missing

backend/src/main/java/ai/giskard/web/rest/controllers/SettingsController.java
```diff
@@ -106,10 +110,22 @@ public AppConfigDTO getApplicationSettings(@AuthenticationPrincipal final UserDe
             .build();
     }
 
+    @GetMapping("/ml-worker-connect")
+    public MLWorkerConnectionInfoDTO getMLWorkerConnectionInfo() throws NoSuchAlgorithmException {
```
Googleton, line 114: The exception is never thrown

backend/src/main/java/ai/giskard/ml/tunnel/OuterChannelHandler.java
```diff
@@ -1,175 +1,179 @@
...
+        log.debug("Outer: Writing {} bytes from {}", in.readableBytes(), outerChannel.id());
+        if (channelRegistry.isDataChannel(outerChannel.id())) {
+            passDataDirectly(in, outerChannel);
+        } else {
+            handleServiceChannelInput(ctx, outerChannel, in);
         }
-
     }
 
     private void handleServiceChannelInput(ChannelHandlerContext ctx, SocketChannel outerChannel, ByteBuf in) {
```
Googleton, line 90: ctx is now unused. i think Sonar will catch other things like that but it hasn't ran on the PR yet

PR #678, opened 2023-01-24
GSK-746 Add proper license validation with Keygen

Googleton:
## Description

Add license upload on the Settings page. This then makes the backend validate the license, if it is valid, it will save it and apply it. License is checked on startup for subsequent restarts of the giskard instance.

The license contains entitlements which are used as feature flags. These entitlements are bound to policies in Keygen's interface.


backend/src/main/java/ai/giskard/security/jwt/JWTConfigurer.java
```diff
@@ -5,6 +5,7 @@
 import org.springframework.security.web.DefaultSecurityFilterChain;
 import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
 
+//TODO: No longer used
```
andreybavt, line 8: Let's remove it then

backend/src/main/java/ai/giskard/service/ee/LicenseService.java
```diff
@@ -0,0 +1,156 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class LicenseService {
+
+    private final Logger log = LoggerFactory.getLogger(LicenseService.class);
+
+    private final FileLocationService fileLocationService;
+
+    private static final String SIGNATURE_KEY = "c947f66224d465b50004c327fc831cff672fc07b540b0613d6f661d0e72d455d";
```
andreybavt, line 30: Could you move it to application.yml and read using `ApplicationProperties` bean?

backend/src/main/java/ai/giskard/service/ee/LicenseService.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        public boolean hasFeature(FeatureFlagService.FeatureFlag flag) {
+            return this.getFeatures().get(flag);
+        }
+    }
+
+    private License license;
+
+    public License getCurrentLicense() {
```
andreybavt, line 62: getCurrentLicense and uploadLicense should be either `synchronized` or rely on double checked locking in order to prevent race condition where 1 request is modifying a license and another one is reading it at the same time

andreybavt (changes requested):

backend/src/main/java/ai/giskard/service/ee/LicenseService.java
```diff
@@ -0,0 +1,156 @@
...
+        decodeLicense(licenseFile);
+        Files.write(fileLocationService.licensePath(), licenseFile.getBytes());
+    }
+
+    private License decodeLicense(String lic) throws IOException {
+        // 1. Remove start/end decorators
+        String encodedPayload = lic.replaceAll("(^-----BEGIN LICENSE FILE-----\\n|\\n|-----END LICENSE FILE-----\\n$)", "");
+
+        // 2. Decode base64 => json
+        byte[] payloadBytes = Base64.getDecoder().decode(encodedPayload);
```
andreybavt, line 95: I'd extract the different steps into separate methods, it'll be easied to read the main decodeLicense logic on a higher level while being able to dive into submethods if needed

backend/src/main/java/ai/giskard/service/ee/LicenseService.java
```diff
@@ -0,0 +1,156 @@
...
+        String decodedLicense = new String(Base64.getDecoder().decode(encodedData));
+        JsonNode licenseJson = mapper.readTree(decodedLicense);
+
+        JsonNode attributes = licenseJson.get("data").get("attributes");
+        JsonNode included = licenseJson.get("included");
+        if (!"ACTIVE".equals(attributes.get("status").asText())) {
+            throw new GiskardRuntimeException("License file is invalid.");
+        }
+
+        // TODO: Check expiration
```
andreybavt, line 134: From keygen docs:

You should assert the following:

- That issued is not greater than the current time, indicating the user has set their system clock to the past, also known as clock tampering.
- That expiry is not less than the current time, indicating an expired license file.

backend/src/main/java/ai/giskard/service/ee/LicenseService.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Takes a license file content, parses it, validates that it is a proper license.
+     * If it's a real license, saves it in giskard home and updates available feature flags
+     *
+     * @param licenseFile
+     */
+    public void uploadLicense(String licenseFile) throws IOException {
+        synchronized (this) {
```
andreybavt, line 71: since the whole method scope is synchronized, it's equivalent to just adding `synchronized` modifier to the method

backend/src/main/java/ai/giskard/service/InitService.java
```diff
@@ -269,8 +270,8 @@ public void initUsers() {
             }
         });
 
-        if (!featureFlagService.hasFlag(FeatureFlagService.FeatureFlag.AUTH)) {
-            // Given the loop above, we can safely assume that the user at least exists.
+        if (!licenseService.getCurrentLicense().hasFeature(FeatureFlag.AUTH)) {
```
andreybavt, line 273: I feel that we'll be using this construction a lot, so perhaps it's worth just adding a `hasFeature` method to the `LicenseService` directly

backend/src/main/java/ai/giskard/service/ProjectService.java
```diff
@@ -89,6 +92,11 @@ private void validateProjectKey(String projectKey) {
     public Project create(Project project, String ownerLogin) {
         String projectKey = project.getKey();
         validateProjectKey(projectKey);
+
+        if (LicenseUtils.isLimitReached(licenseService.getCurrentLicense().getProjectLimit(), (int) projectRepository.count())) {
+            throw new GiskardRuntimeException("Project limit is reached. You can upgrade your plan to create more.");
```
andreybavt, line 97: Let's create a separate `LicenseValidationException` to handle these kind of errors. It'll allow us to handle it separately in `ExceptionTranslator` if needed

andreybavt (changes requested):

andreybavt:
There needs to be a validation to make sure the form isn't submitted without first name, last name and email.

There should also be error handling to make sure there's no infinite spinner in case Make pipeline fails
image

frontend/src/views/setup/Setup.vue
```diff
@@ -0,0 +1,110 @@
...
+            <v-btn small color="primary" @click="openFileInput">Upload license file</v-btn>
+            <input type="file" ref="fileInput" style="display: none;" @change="onFileUpdate"/>
+          </v-card-actions>
+        </v-card>
+      </v-col>
+    </v-row>
+    <v-row v-if="done">
+      <v-col>
+        <v-card>
+          <v-card-title>Thank you for installing Giskard</v-card-title>
```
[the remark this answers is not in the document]
Googleton: I added some things to ensure it doesn't happen, but it didn't happen very consistently for me. If you manage to make it happen again with my fixes now I'd love to know, because now I can't manage to reproduce it.

Googleton:
I added validation and a `finally` clause to prevent the spinner from getting stuck if an error happens on the make side, in response to your comment. Can't mark it as resolved since it's not a review comment :)

PR #679, opened 2023-01-24
Task/explicit error message get project wrong key

thomasabou01:


python-client/giskard/client/giskard_client.py
```diff
@@ -96,7 +96,11 @@ def get_project(self, project_key: str):
                 The giskard project that belongs to the project key
         """
         self.analytics.track("Get Project", {"project_key": anonymize(project_key)})
-        response = self._session.get(f"project", params={"key": project_key}).json()
+        try:
+            response = self._session.get(f"project", params={"key": project_key}).json()
+        except GiskardError as e:
+            warning("No project found with the provided key")
```
andreybavt, line 102: Since the error is raised anyway, instead of showing a warning you could chain two errors with
```
raise ValueError(f"No project found with the provided key: {project_key}") from e
``` 

It's also a good practice to be specific in the error message, so adding a project_key would help

andreybavt (changes requested):
S

andreybavt:
Closed after checking the current message that is already pretty clear: 
```
Project not found by key: test: no details
```

PR #682, opened 2023-01-26
moved proba calculation from relative to absolute

rabah-khalek:
## Description

I had the disparate impact as ratio of average probabilities, while I believe it should be a sum of them instead.

## Type of Change

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
     branches:
       - main
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


andreybavt:
@rabah-khalek apparently some tests don't pass anymore

andreybavt (changes requested):
there are currently failing unit tests

rabah-khalek:
Yes, I had to change the unit test, because the metric has changed. As for the new metric, I already benchmarked it yet again with the plots in the new article on disparate_impact. Thanks @andreybavt

rabah-khalek:
Upon reflection, i will close this pr, keeping the DI relative.

I thought first that the relative version is not a good representation for small statistical sample. At the same time, the absolute version deviates from the actual definition of disparate impact, i.e ratio of probabilities.

PR #710, opened 2023-02-02
GSK-753 Feature/ai test v2 suite creation

kevinmessiaen:
## Description

- Added possibility to create test suite with optionals inputs (each input from Giskard that doesn't have a default value are dynamically calculated)
image

- Automatic selection of tests based on the suite inputs
image

- defined more tag for test to improve the automatic test selection
image

## Type of Change

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Giskard lets you instantly see your model's prediction for a given set of featur
 
 Saw anything strange? Leave a feedback directly within Giskard, so that your team can explore the query that generated the faulty result. Designed for both tech and business users, Giskard is super intuitive to use! 👌
 
-And of course, Giskard works with any model, any environment and integrates seemlessly with your favorite tools ⤵️ <br/>
+And of course, Giskard works with any model, any environment and integrates seamlessly with your favorite tools ⤵️ <br/>
 <p>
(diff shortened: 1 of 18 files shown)


andreybavt:
@kevinmessiaen , when creating a test suite, unlike in AI Test v1 you should be able to specify whichever combination of inputs you like, not just model, actual ds, reference ds. Let's discuss when you're available

andreybavt (changes requested):

andreybavt:
image


@kevinmessiaen , actually the idea for the suite creation design was different compared to the current implementation. Here's what we had in mind:

When adding a new input user sets a name for it and choses a type
image
At this moment you don't specify concrete inputs, it's like a "test suite template" that only has placeholders for concrete inputs (actual models and datasets)
This is a major difference compared to AI Test v1 where we filled in the inputs at the suite creation time.

If the "Auto-generate tests" toggle is on the tests whose signature fits the set of provided inputs (and based on some additional heuristics) get created and inputs are mapped between suite inputs (provided by a user) and test inputs (provided by a test developer). By default all test inputs are "exposed", meaning that they're mapped to suite inputs. Then in needed a user may remap inputs in the auto-generated tests or "unexpose" them by assigning a fixed input.

PR #724, opened 2023-02-08
Fixed issue where selected tab was not highlighted by default

kevinmessiaen:
## Description

Fixed issue where selected tab was not highlighted by default

## Type of Change

## Checklist

### DIFF
--- a/backend/src/test/java/ai/giskard/web/rest/testing/TestControllerIT.java
+++ b/backend/src/test/java/ai/giskard/web/rest/testing/TestControllerIT.java
@@ -74,9 +74,15 @@ class TestControllerIT {
     @Transactional
     void testDatasetMetadata() throws Exception {
         Project project = projectRepository.getOneByName(initService.getProjectByCreatorLogin("admin"));
-        Optional<Dataset> dataset = project.getDatasets().stream().findFirst();
-        Assertions.assertFalse(dataset.isEmpty(), "demo dataset not found");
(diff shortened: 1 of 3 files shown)


Googleton (commented):
Code is good for me but it seems like integration tests fail

kevinmessiaen:
> Code is good for me but it seems like integration tests fail

I fixed it, I think it was an issue coming from `main`

Googleton:
I see your fix, I'm not sure if changing the test is proper in this case. Shouldn't we look at the test data instead and figure out why it breaks now ? Maybe it's something trivial that doesn't require changing the test itself ?

Googleton:
> > I see your fix, I'm not sure if changing the test is proper in this case. Shouldn't we look at the test data instead and figure out why it breaks now ? Maybe it's something trivial that doesn't require changing the test itself ?
> 
> Actually the test isn't deterministic: `project.getDatasets().stream().findFirst();` will always return any of available dataset randomly and some does not have target by default

Yea that's true. Should we consider getting a specific demo project for this test?

kevinmessiaen:
> > > I see your fix, I'm not sure if changing the test is proper in this case. Shouldn't we look at the test data instead and figure out why it breaks now ? Maybe it's something trivial that doesn't require changing the test itself ?
> > 
> > 
> > Actually the test isn't deterministic: `project.getDatasets().stream().findFirst();` will always return any of available dataset randomly and some does not have target by default
> 
> Yea that's true. Should we consider getting a specific demo project for this test?

I've selected one with a feature by default.

andreybavt (approved):

Issue #725, opened 2023-02-08
Filtering doesn't work on texts containing `\n` caracters


andreybavt:
Currently, we iterate over the lines read by BufferedReader:

https://github.com/Giskard-AI/giskard/blob/e57911a8f8a85c9c2c0441451e93b7838cfec47b/backend/src/main/java/ai/giskard/service/ml/MLWorkerService.java#L239

This doesn't handle the case when the text inside a table cell also contains `\n` characters.

To reproduce:
- Create an inspection in ENRON project
- use a filter like:
```
def filter_row(row):
    return len(row.Content) > 0
```

The results are the following 
![image](https://user-images.githubusercontent.com/7889985/217525010-732ef8f6-26e0-4ed7-8912-b7502b977486.png)

We should probably rely on Apache commons CSV to read records:
```java
final CSVFormat csvFormat = CSVFormat.Builder.create()
        .setHeader(HEADERS)
        .setAllowMissingColumnNames(true)
        .build();
final Iterable<CSVRecord> records = csvFormat.parse(reader);

```

Related issue:
Backend errors aren't reported to the frontend when they occur during the filtering

Issue #722, opened 2023-02-08
Possible to run slice validation without selecting a dataset


andreybavt:
![image](https://user-images.githubusercontent.com/7889985/217493038-6d726d70-2f96-4cff-befb-cd952fee86eb.png)

Issue #720, opened 2023-02-08
Possibility to create slices without name


kevinmessiaen:
### Issue Type

Bug

### Source

docker

### Giskard Library Version

1.7.0.

### Giskard Server Version

1.3.2

### Current Behaviour?

```shell
It is possible to create a slice without specifying its name. The name should be mandatory.
```

### Standalone code OR list down the steps to reproduce the issue

```shell
- Open a project and go to the model tab
- Inspect the model
- Click on slice to apply and then select create new slice
- Save while letting the slice name field blank
```

Googleton:
Good catch, will fix asap!

PR #734, opened 2023-02-10
Various feedback for slicing

Googleton:
## Description

- [x] Slice name should not be empty #720 
- [x] Default slice code should be more descriptive, have examples 
- [x] Validation should be disabled when no dataset is picked #722 
- [x] The currently inspected dataset should be selected by default 
- [x] Datasets containing newline characters do not work #725 
- [x] Add mixpanel tracking for slice creation & application

### DIFF
--- a/backend/src/main/java/ai/giskard/service/SliceService.java
+++ b/backend/src/main/java/ai/giskard/service/SliceService.java
@@ -91,8 +91,11 @@ public class SliceService {
     }
 
     private void validateSliceName(String sliceName) {
-        if (sliceName.length() > 30) {
-            throw new IllegalArgumentException(String.format("Slice name %s is too long. Slice names should not be longer than 30 characters.", sliceName));
+        if (sliceName.length() < 1) {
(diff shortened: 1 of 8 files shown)


backend/src/main/java/ai/giskard/service/SliceService.java
```diff
@@ -91,8 +91,11 @@ private static void writeSliceResultToCacheFile(Path cachedSliceFile, List<Integ
     }
 
     private void validateSliceName(String sliceName) {
-        if (sliceName.length() > 30) {
-            throw new IllegalArgumentException(String.format("Slice name %s is too long. Slice names should not be longer than 30 characters.", sliceName));
+        if (sliceName.length() < 1) {
```
kevinmessiaen, line 94: We should check if the sliceName is not blank to avoid potential validation of name like `  `:

```suggestion
        if (sliceName == null || sliceName.trim().isEmpty()) {
```

kevinmessiaen (approved):

andreybavt (approved):

PR #837, opened 2023-02-22
Fixed settings display issue

kevinmessiaen:
## Description

Fixed issue when setting page disappears when overflowing on the page

## Type of Change

- [ ] 🔐 Security fix


andreybavt (approved):

frontend/src/views/main/admin/settings/Settings.vue
```diff
@@ -1,15 +1,18 @@
 <template>
-  <v-container fluid id="container-project-tab" class="vertical-container">
+  <div>
```
andreybavt, line 2: there's still `#container-project-tab` style in Project.vue

we should either apply again or get rid of it

andreybavt (changes requested):

frontend/src/views/main/admin/settings/Settings.vue
kevinmessiaen: Yes actually it's used in `project.vue`. I think it was inside `setting.vue` as an error since the id is not really related to the content itself

PR #842, opened 2023-02-24
Fix Giskard dev installation on Windows (no WSL)

Googleton:
This branch is meant to make a Giskard dev install work seamlessly when installing on Windows.

Here are the fixes it implements:

- Change giskard-home to be equal to `${user.home}/giskard-home` instead of `${HOME}/giskard-home` => user.home is provided by Spring and is compatible with all operating systems, HOME does not exist by default on Windows
- Add some documentation for common Python errors (working on making it happen by default instead of be in documentation at the moment....)


Googleton:
@andreybavt Do you see an issue with using `user.home` instead of the `HOME` env var?

andreybavt:
> @andreybavt Do you see an issue with using `user.home` instead of the `HOME` env var?

yes, Spring should handle it fine

PR #865, opened 2023-03-06
AI Test v2

andreybavt:


Googleton:
Where could I find example notebooks to properly test this ? Or would it be better to try and do self-discovery ?

PR #869, opened 2023-03-09
Task/gsk 863 handle batch dimension

rabah-khalek:
## Description

Issues with handling the shape of Output tensor in PyTorch

### DIFF
--- a/python-client/giskard/models/pytorch/__init__.py
+++ b/python-client/giskard/models/pytorch/__init__.py
@@ -119,18 +119,13 @@ class PyTorchModel(MLFlowBasedModel):
         if to_unpack:
             with torch.no_grad():
                 for entry in data:
-                    predictions.append(self.clf(*entry).detach().numpy())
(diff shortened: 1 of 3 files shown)


python-client/giskard/models/pytorch/__init__.py
```diff
@@ -94,69 +94,39 @@ def save_with_mlflow(self, local_path, mlflow_meta: mlflow.models.Model):
...
-        self.check_if_one_entry_prediction_is_valid(data, to_unpack)
-        predictions = []
-        if to_unpack:
-            for entry in data:
-                predictions.append(self.get_torch_prediction_unpack_input(entry))
-        else:
-            for entry in data:
-                predictions.append(self.get_torch_prediction(entry))
-        return predictions
+            return [self.clf(*entry).detach().squeeze(0).numpy() for entry in _map_to_tuple(data)]
```
rabah-khalek, line 99: I still prefer to check if the data is tupled once instead of every iteration.

PR #882, opened 2023-03-27
implemented automodel

rabah-khalek:
https://linear.app/giskard/issue/GSK-886/automodel-class
to solve: https://github.com/Giskard-AI/giskard/pull/882
Inferring the `clf` module to define the Giskard model.
Screenshot 2023-03-27 at 12 49 29


rabah-khalek:
Re: 

>  the default behavior is that we try to guess the model kind, and the user doesn't have to think about it. Then if their model is not supported we write a small bit of doc. 
that is kind of what i implemented so far. the doc is just a message listing what we currently support as models.

As for the name, I followed a bit the nomenclature of HuggingFace (like `TFAutoModel`) which I thought could be intuitive to users. But I am not opposed to your 2 suggested names as well.

@jmsquare , @andreybavt , @mattbit WDYT?

mattbit:
```
isinstance(custom_model, sklearn.base.BaseEstimator)
```
might be better

rabah-khalek:
@mattbit I generalised the class to cover the points we discussed.
I also added a unit-test to check if all models are "discovered" correctly. Have a look!

python-client/giskard/models/automodel/__init__.py
```diff
@@ -0,0 +1,30 @@
+from importlib import import_module
+
+
+class AutoModel:
+    def __new__(cls, *args, **kw):
```
mattbit, line 5: I have some reservations on the design of this. In my opinion it is a bit tricky to have a fake class `AutoModel` which can never be instantiated, leading to things like:
```py
model = AutoModel(my_classifier)
assert isinstance(model, AutoModel)  # fail!
```
This would be counterintuitive for me.

Maybe we should just put this code in the parent `Model` class.

Actually, I would rather go for a standard [factory pattern](https://en.wikipedia.org/wiki/Factory_method_pattern), like Huggingface does.

Something like
```py
from giskard import Model

model = Model.from_classifier(my_classifier)
```
?

mattbit (changes requested):

python-client/giskard/models/automodel/__init__.py
rabah-khalek: done

mattbit (approved):
Looks good to me.
For the helper method, maybe calling it just `make` is ok, its usage would be:
```py
model = Model.make(clf)
```
which is expressive and semantically makes sense as it is just an alias for `make_model`. But I have no strong preference.

mattbit:
Along this line, I was thinking that we could expose similar helpers to make life easier for the user.
While it would break a bit SOLID principles, users will have a single entrypoint instead of having to discover the ad-hoc classes for torch, tensorflow, etc.

I’m thinking about something like
```py
from giskard import Model

model = Model.from_torch(my_torch_model)

# or
model = Model.from_sklearn(my_sklearn_model)
```
Feels more natural to me and it is quite idiomatic, the user will just type `Model.from_` and get a list of available options from the IDE autocomplete:

Screenshot 2023-03-31 at 09 56 11

I think it could be worth the exception, and since it would be simple aliases I see no significant increase of complexity in code and maintainability.

What do you think @rabah-khalek @andreybavt @jmsquare?

andreybavt:
> Along this line, I was thinking that we could expose similar helpers to make life easier for the user. While it would break a bit SOLID principles, users will have a single entrypoint instead of having to discover the ad-hoc classes for torch, tensorflow, etc.
> 
> I’m thinking about something like
> 
> ```python
> from giskard import Model
> 
> model = Model.from_torch(my_torch_model)
> 
> # or
> model = Model.from_sklearn(my_sklearn_model)
> ```
> 
> Feels more natural to me and it is quite idiomatic, the user will just type `Model.from_` and get a list of available options from the IDE autocomplete:
> 
> Screenshot 2023-03-31 at 09 56 11
> 
> I think it could be worth the exception, and since it would be simple aliases I see no significant increase of complexity in code and maintainability.
> 
> What do you think @rabah-khalek @andreybavt @jmsquare?

Having factory methods in addition to class constructors is a nice idea. I just think that adding them on `Model` class could be not so clean. For example it wouldn't make sense to have these methods on model objects that were created using the subclass constructors, e.g.:

```python
m = SKLearnModel(...)
m.from_catboost(...) # this methods would be suggested by autocomplete
SKLearnModel.from_torch(...) # same here
```


So I think we could simply add them to the `models` module, in this case the workflow would be:
```python
from giskard import model

model.from_sklearn(...) # One way to use it
model.SKLearnModel(...) # If you know that this class exists
model.create(...) # or `make` for autodetection

```

rabah-khalek:
to be honest I'm not super happy about aliasing the factory method in the base model class. It is not hierarchically clean. 
I think we have these two options:
- a factory method (and I'm not of `make_model`) -- maybe `create_model` instead?
- or a factory class -- we can call it `Model` and change the name of our current base model to `BaseModel` in which we can implement your suggestions @mattbit ( `.from_XXX` method) and maybe `Model.autocreate(*args, **kwargs)` for the automatic case.

I am personally in favour of the latter.

mattbit:
I think that @andreybavt’s proposal of putting helper methods at the module level is still the cleanest option.

rabah-khalek:
Let's stick then to a simple factory method on the module level, called `create_model` or `autocreate_model`, no aliasing, and gather feedbacks from user to see if it's intuitive enough. Maybe @jmsquare you can help with the naming?

andreybavt:
> I completely missed @andreybavt's comment... somehow. My bad. Just to understand.. in this:
> 
> ```python
> from giskard import model
> 
> model.from_sklearn(...) # One way to use it
> model.SKLearnModel(...) # If you know that this class exists
> model.create(...) # or `make` for autodetection
> ```
> 
> `model` is the module? (it is in fact `models`) or a factory class defined in `models/__init__.py`?

yeah, my bad, it's the `models` module. Now that I made that typo I'm thinking that it might be worth dropping the "s" in the end of it's name, but no strong opinion on it

jmsquare:
If I understood well, there are 2 upload possibilities : 

- **Case A**:
```
from giskard import model, Dataset

my_dataset = Dataset(my_df, ...)
my_model = model.create(clf, ...)
```
- **Case A**:
```
from giskard import Model, Dataset

my_dataset = Dataset(my_df, ...)
my_model = Model(clf, ...)
```
**Case A** enables auto-completion of the `.fromXXX` method. But **Case B** is much simpler since it's symmetric with the way we implement Dataset and do not require this additional method `create`

If we are enough confident on our model inference process, we absolutely should not advertize the `.fromXXX` methods: either in the documentation, **nor** in the auto-completion. Indeed, if users see that `.fromXXX` exist, they will tend to use the specific model object instead of our generic model class. We saw that many users don't even know the ML library they are using, they only know the ML methods they are using. For instance, Nicolas knew he was using RandomForest, but not scikit learn. So **Case B** adds an additional step to the upload process.

However, I do agree, that if one day, we use `my_cv_dataset = dataset.ComputerVision(object_to_spec, etc.)`, it would be better to have **Case A**.

For the moment, I prefer to stay with **Case B**. This is simpler for the user (our key requirement) and it does not need to implement all the `.fromXXX`  methods that can be misleading.

rabah-khalek:
no we don't have Case B, JM. 
there's no class `Model` that does the `auto-inference` rather a module (file) called `model` that contains a function `create` that does it.

jmsquare:
Ah ok, I understand the need to create a  function (create) instead of an object to be more Pythonic. So if I understood well, we gonna have:
```
 from giskard import model
 from giskard import dataset
 
 my_model = model.create(my_clf, ...)
 my_dataset = dataset.create(my_df, ...)
```
To avoid importing many modules and using it in the code, I propose we use some wrapper function. The above code would become just

```
 my_model = create_model(my_clf, ...)
 my_dataset = create_dataset(my_df, ...)
```

mattbit:
I think @jmsquare raised a very important point regarding the inconsistency between `Model` and `Dataset` creation.

Also, it does not feel right to me to rename the module `models` to `model`, because of the confusion between `from giskard import Model` and `from giskard import model`, and the inconsistency with how we import `Dataset`.

It’s tricky and I don’t have a well defined opinion on this. Maybe let’s stick to `Model.create(...)` or `Model.make(...)` for now and see what happens in user tests?

andreybavt:
I think exposing `create_model` and `create_dataset` from the root of `giskard` module is a good idea, it'll be a quick start for some users by just doing `from giskard import create_model, create_dataset`

At the same time it won't be as powerful as `from_X` factory methods or dedicated constructors since we can't add model specific arguments to the generic `create_model` method

So I think we should keep all three:
1. `create_model` - highest level method that detects which model should be created and creates it
2. `from_X` methods (can be also renamed to `create_X_model`) to be consistent with the first method. This will increase discoverability, because if you have autocompletion and you type `from giskard import create_` you get all of the possible models that we support. In this case each `create_X_model` will have parameters that are only specific to a given model type
3. Just submodels class constructors like SKLearnModel, this will be the lowest level that we'll be using anyway, but if a user is familiar with our classes, why not creating models directly

andreybavt:
> ```python
> giskard_model = model(my_custom_model, ...)  # instance of giskard.models.Model
> ```

it's also kind of renaming `create_model` from proposal above to just `model` isn't it (same for datasets)?

If we do it, then indeed we'll have to add `s` to the module name to distinguish between module and method name

rabah-khalek:
Yeah I still prefer this one. Ok good, I'll take care of it.

mattbit:
> it's also kind of renaming `create_model` from proposal above to just `model` isn't it (same for datasets)?

Yes, not focusing on the name (the two would work fine in my opinion). Just keeping the modules with their current name (`models`), and exposing an helper at the root level allowing to `from giskard import model` or `from giskard import create_model` and ensuring that this is symmetric with dataset creation as JM suggested.

rabah-khalek:
here're the main files that were refactored:
```
model/ -> models/
giskard/core/model.py -> giskard/models/base/__init__.py
giskard/ml_worker/core/model.py (empty) -> deleted
giskard/ml_worker/core/dataset.py -> giskard/datasets/base/__init__.py
```
more details here:
<details>
  <summary>Previous structure (branch: new-upload-API)</summary>

```
python-client/giskard
├── __init__.py
├── cli.py
├── cli_utils.py
├── core
│   ├── __init__.py
│   ├── core.py
│   ├── dataset_validation.py
│   ├── model.py
│   ├── model_validation.py
│   └── validation.py
├── ml_worker
│   ├── __init__.py
│   ├── core
│   │   ├── dataset.py
│   │   ├── model.py
│   │   ├── model_explanation.py
│   │   ├── suite.py
│   │   ├── test_result.py
│   │   └── test_runner.py
│   └── ml_worker.py
├── models
│   ├── __init__.py
│   ├── catboost
│   │   └── __init__.py
│   ├── huggingface
│   │   └── __init__.py
│   ├── pytorch
│   │   └── __init__.py
│   ├── sklearn
│   │   └── __init__.py
│   ├── tensorflow
│   │   └── __init__.py
│   └── utils.py
├── path_utils.py
├── settings.py
└── temp.py
```
</details>

<details>
  <summary>New structure (this branch)</summary>

```
python-client/giskard
├── __init__.py
├── cli.py
├── cli_utils.py
├── core
│   ├── __init__.py
│   ├── core.py
│   ├── dataset_validation.py
│   ├── model_validation.py
│   └── validation.py
├── datasets
│   ├── __init__.py
│   └── base
│       └── __init__.py
├── ml_worker
│   ├── __init__.py
│   ├── core
│   │   ├── model_explanation.py
│   │   ├── suite.py
│   │   ├── test_result.py
│   │   └── test_runner.py
│   └── ml_worker.py
├── models
│   ├── __init__.py
│   ├── base
│   │   └── __init__.py
│   ├── catboost
│   │   └── __init__.py
│   ├── huggingface
│   │   └── __init__.py
│   ├── pytorch
│   │   └── __init__.py
│   ├── sklearn
│   │   └── __init__.py
│   ├── tensorflow
│   │   └── __init__.py
│   └── utils.py
├── path_utils.py
└── settings.py
```
</details>

Now a user can create a model and dataset as follows:
```py
from giskard import model, dataset
my_model = model(my_clf, ...) # instance of SKLearnModel, PyTorchModel, etc.
my_dataset = dataset(my_df, ...) # instance of Dataset
```
but also:
Screenshot 2023-04-03 at 13 17 34

with auto-completion also for the factory methods:
Screenshot 2023-04-03 at 13 18 34

rabah-khalek:
further refactorings in `python-client` to be performed according to https://github.com/Giskard-AI/giskard/issues/909

python-client/giskard/client/giskard_client.py
```diff
@@ -177,7 +177,7 @@ def save_model_meta(self, project_key: str, model_id: UUID, meta: ModelMeta, pyt
             },
         )
         self.analytics.track(
-            "Upload Model",
+            "Upload BaseModel",
```
andreybavt, line 180: this should be just Model

python-client/giskard/core/model_validation.py
```diff
@@ -67,20 +67,20 @@ def validate_model_execution(model: Model, dataset: Dataset) -> None:
...
+def validate_deterministic_model(model: BaseModel, validate_ds: Dataset, prev_prediction):
     """
     Asserts if the model is deterministic by asserting previous and current prediction on same data
     """
     new_prediction = model.predict(validate_ds)
 
     if not np.allclose(prev_prediction.raw, new_prediction.raw):
         warning(
-            "Model is stochastic and not deterministic. Prediction function returns different results"
+            "BaseModel is stochastic and not deterministic. Prediction function returns different results"
```
andreybavt, line 78: Model

andreybavt, line 78: Model

andreybavt, line 78: Model

python-client/giskard/ml_worker/testing/newtestlib/test.py
```diff
@@ -1,19 +1,19 @@
...
+from giskard.ml_worker.testing.registry.lib import BaseModel, Dataset
 
 
 # @test(inputs={
 #     'test_dataset': {
 #         'requirements': {TARGET_DEFINED}
 #     }
 # })
-# def my_test(model: Model, train_dataset: Dataset, test_dataset: Dataset):
+# def my_test(model: BaseModel, train_dataset: Dataset, test_dataset: Dataset):
```
andreybavt, line 10: this can be removed alltogether

python-client/giskard/ml_worker/testing/tests/performance.py
```diff
@@ -98,14 +98,14 @@ class AucTest(GiskardTest):
...
+    model: BaseModel
     threshold: float
 
-    def __init__(self, actual_slice: Dataset = None, model: Model = None, threshold: float = None):
+    def __init__(self, actual_slice: Dataset = None, model: BaseModel = None, threshold: float = None):
         """
         Init an AucTest instance, can be passed in Suite.add_test method
         :param actual_slice: Slice of the actual dataset
-        :param model: Model used to compute the test
+        :param model: BaseModel used to compute the test
```
andreybavt, line 108: Model

python-client/giskard/ml_worker/testing/tests/performance.py
```diff
@@ -137,8 +137,8 @@ def test_auc(actual_slice: Dataset, model: Model, threshold: float = 1.0):
     Args:
         actual_slice(Dataset):
           Slice of the actual dataset
-        model(Model):
-          Model used to compute the test
+        model(BaseModel):
+          BaseModel used to compute the test
```
andreybavt, line 141: Model

python-client/giskard/ml_worker/testing/tests/performance.py
```diff
@@ -176,8 +176,8 @@ def test_f1(actual_slice: Dataset, model: Model, threshold: float = 1.0):
...
```
andreybavt, line 180: Model

python-client/giskard/ml_worker/testing/tests/performance.py
```diff
@@ -596,8 +596,8 @@ def test_diff_rmse(actual_slice: Dataset, reference_slice: Dataset, model: Model
           Slice of the actual dataset
         reference_slice(Dataset):
           Slice of the actual dataset
-        model(Model):
-          Model used to compute the test
+        model(BaseModel):
+          BaseModel used to compute the test
```
andreybavt, line 600: everywhere in docstrings in this file "Model used" referred just to a "model" word rather than to a "Model" class

python-client/giskard/ml_worker/testing/performance_tests.py
```diff
@@ -572,8 +572,8 @@ def test_diff_reference_actual_rmse(self, reference_slice, actual_slice, model,
               Slice of actual dataset
             reference_slice(Dataset):
               Slice of reference dataset
-            model(Model):
-              Model used to compute the test
+            model(BaseModel):
+              BaseModel used to compute the test
```
andreybavt, line 576: same comment as for the rest of the "*_tests.py" where "Model used" meant just "Model used", not "BaseModel used"

python-client/giskard/models/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+from giskard.models.sklearn import SKLearnModel
+from giskard.models.catboost import CatboostModel
+from giskard.models.pytorch import PyTorchModel
+from giskard.models.tensorflow import TensorFlowModel
+from giskard.models.huggingface import HuggingFaceModel
+
+# format: dict[GiskardModel: list(tuple(module, base_class))]
+_libraries = {HuggingFaceModel: [("transformers", "PreTrainedModel")],
+              SKLearnModel: [("sklearn.base", "BaseEstimator")],
+              CatboostModel: [("catboost", "CatBoost"),
```
andreybavt, line 14: why not just checking for the parent class `CatBoost` directly? In either case all of the catboost models inherit from it and pass by its `save_model` method when being saved

andreybavt (changes requested):
it's generally safer to do renaming with `Shift+F6`. Intellij is pretty smart to be able to find the relevant usages and not word occurences in texts

rabah-khalek:
you're right, I used `Shift+F6` for everything except when I renamed `Model` to `BaseModel`. I'll take care of it!

python-client/giskard/models/__init__.py
andreybavt: Indeed it's weird that mlflow check each subtype separately, looking at the code, they all inherit from `CatBoost`

https://github.com/catboost/catboost/blob/d84d92c6757c71132b5d700c4e09cafe81f13422/catboost/python-package/catboost/core.py#L1754

rabah-khalek: Oh, sorry. You're right! I didn't check their code actually, only their doc, then mlflow which made me a bit biased. Thanks for this catch!

python-client/giskard/ml_worker/testing/tests/performance.py
rabah-khalek: I think it can be kept "Model" as it's a new line, so a non-biased user who doesn't know about our initial naming of `Model` the class, wouldn't consider it as a class.
```            
            actual_slice(Dataset):
              Slice of the actual dataset
            model(BaseModel):
              Model used to compute the test
            threshold(float):
              Threshold value for Accuracy
 ```

andreybavt (approved):

Issue #885, opened 2023-03-30
[GSK-896] Better error handling


rabah-khalek:
28/03/2023 (webedia)

* \[x\] **solved in** [**https://github.com/Giskard-AI/giskard/pull/911/commits/915e8a7a5936082a9e2068fab9b57f22f44614e9**](https://github.com/Giskard-AI/giskard/pull/911/commits/915e8a7a5936082a9e2068fab9b57f22f44614e9) --  `Dataset` initialization, the function signature just has one mandatory argument, but it actually requires at least one of the optional parameters (e.g. `cat_columns`, `column_types`). There should be a `ValueError` that was not raised and the user had no idea why he could not create a dataset with just `ds = Dataset(df)`. **This error is caused because one of the dataframe columns is of type** `list` (currently not supported by Giskard!) The user has a list column but they transforms it in the preprocessing before passing it to the model. User has columns of these types:
  * `list[int]`
  * `list[dict]`
* \[x\] **solved in** [**https://github.com/Giskard-AI/giskard/pull/911/commits/c172d7e4836506e813cdfd29b25d01bc40109dad**](https://github.com/Giskard-AI/giskard/pull/911/commits/c172d7e4836506e813cdfd29b25d01bc40109dad) -- `validate_model` user passes their model and not the Giskard model (→ should add an error message)
* \[x\] **solved with** [**https://github.com/Giskard-AI/giskard/pull/882**](https://github.com/Giskard-AI/giskard/pull/882) -- add an error message by checking the clf.**module** if it unmatch the wrapper class.
* \[x\] solved in [https://github.com/Giskard-AI/giskard/pull/911/commits/3644c49443540c9d1806c4bedbade7917f3a9513](https://github.com/Giskard-AI/giskard/pull/911/commits/3644c49443540c9d1806c4bedbade7917f3a9513) -- Preprocessing function taking 2 arguments (*df, blacklist* in this case): When the data preprocessing function is wrong there is no error message
  * check number of inputs of the preproc func and postproc func

30/03/2023 (webedia)

* [ ] The message error of the test is not clear -- @jmsquare do you have more info on this one?, @kevinmessiaen could you take care of it?
* [ ] unclear error when type is not good for `add_test`: It seems that you passed a `TestResult` to the test, please pass a test function reference and the parameters separately instead of calling the test function directly -- @kevinmessiaen could you take care of it?
[...]

Issue #884, opened 2023-03-30
[GSK-894] wrap modules' import inside methods that need them


rabah-khalek:
wrap the imports so that we print a nice error message if the user tries to use certain features without having installed the proper package. For example, if the user tries to setup a sklearn model but the `sklearn` library is missing, we may show a message "Please install `giskard[sklearn]` to use this functionality" or similar.

<sub>From [SyncLinear.com](https://synclinear.com) | [GSK-894](https://linear.app/giskard/issue/GSK-894/wrap-modules-import-inside-methods-that-need-them)</sub>

PR #911, opened 2023-04-03
Implementing Webedia's user-test feedback (part 1)

rabah-khalek:
To solve:

- [x] https://github.com/Giskard-AI/giskard/issues/884 -- solved in https://github.com/Giskard-AI/giskard/pull/911/commits/291d4921c59922139baeadc2bc78baf0adb61493
- [x] https://github.com/Giskard-AI/giskard/issues/885


python-client/giskard/models/huggingface/__init__.py
```diff
@@ -1,15 +1,17 @@
...
 from giskard.core.core import SupportedModelTypes
 from giskard.models.base import WrapperModel
 
-from transformers import pipeline, pipelines
+try:
+    import torch
+    from transformers import pipeline, pipelines
+except ImportError:
+    pass
+
```
rabah-khalek, line 14: Although python caches the package on import ([ref](https://docs.python.org/3.6/reference/import.html#the-import-system)), I didn't like the idea of importing `torch` and `transformers` anytime they're used inside a function. Another reason is that they're sometimes used in the signature of a function. 

With `try; except` the user won't receive an error if he doesn't have `mlflow`, `torch` or `transformers` when they `import giskard` but only get the error if they use the corresponding class.

@andreybavt is there a better way to do this?

rabah-khalek:
@andreybavt this can be reviewed now, I'll migrate the remaining three comments in https://github.com/Giskard-AI/giskard/issues/885 to a new issue/PR.

python-client/giskard/core/model_validation.py
```diff
@@ -191,12 +191,16 @@ def validate_label_with_target(classification_labels, target_values=None, target
                 'to make results more understandable in Giskard."'
             )
 
+        _to_append = ""
```
andreybavt, line 194: _to_append is already private to the method, so no real need to add `_` in front of the name

andreybavt, line 194: Not an urgent  improvement, but I think if we're going to add arbitrary_types_allowed every time, it's just worth wrapping validate_arguments with our own decorator that sets `{'arbitrary_types_allowed': True}` by default

python-client/giskard/models/base/__init__.py
```diff
@@ -25,6 +22,12 @@
 from giskard.path_utils import get_size
 from giskard.settings import settings
 
+try:
```
andreybavt, line 25: as discussed, we install mlflow-skinny by default, so no need for this try-except

python-client/giskard/models/base/__init__.py
```diff
@@ -281,21 +290,33 @@ class WrapperModel(BaseModel, ABC):
...
+            classification_labels=None,
     ) -> None:
         super().__init__(model_type, name, feature_names, classification_threshold, classification_labels)
         self.clf = clf
         self.data_preprocessing_function = data_preprocessing_function
         self.model_postprocessing_function = model_postprocessing_function
 
+        if self.data_preprocessing_function:
+            sign_len = len(signature(self.data_preprocessing_function).parameters)
+            print(sign_len)
```
andreybavt, line 310: do we need printing sign_len here?

python-client/giskard/models/base/__init__.py
```diff
@@ -281,21 +290,33 @@ class WrapperModel(BaseModel, ABC):
...
+            feature_names=None,
+            classification_threshold=0.5,
+            classification_labels=None,
     ) -> None:
         super().__init__(model_type, name, feature_names, classification_threshold, classification_labels)
         self.clf = clf
         self.data_preprocessing_function = data_preprocessing_function
         self.model_postprocessing_function = model_postprocessing_function
 
+        if self.data_preprocessing_function:
```
andreybavt, line 308: it can be cleaner if instead of `any` we declare a type of pre/post-processing functions using `Callable[...]` in case it's provided `@validate_arguments` might do the check itself

python-client/giskard/models/catboost/__init__.py
```diff
@@ -1,10 +1,12 @@
 from typing import Union
-
-import mlflow
-
 from giskard.core.core import SupportedModelTypes
 from giskard.models.base import MLFlowBasedModel
 
+try:
```
andreybavt, line 5: see similar comment

python-client/giskard/models/huggingface/__init__.py
```diff
@@ -1,15 +1,17 @@
...
 from scipy import special
-import torch
 from pathlib import Path
 import yaml
-
 from giskard.core.core import SupportedModelTypes
 from giskard.models.base import WrapperModel
 
-from transformers import pipeline, pipelines
+try:
```
andreybavt, line 9: can be removed since this case is already handled in the ai-test-v2-merged branch

python-client/giskard/models/pytorch/__init__.py
```diff
@@ -1,19 +1,21 @@
...
 import collections
-from torch.utils.data import DataLoader
-from torch.utils.data import Dataset as torch_dataset
-
 from giskard.core.core import SupportedModelTypes
 from giskard.models.base import MLFlowBasedModel
-
 from ..utils import map_to_tuples
 
+try:
```
andreybavt, line 11: same comment as in other classes

andreybavt (changes requested):

python-client/giskard/models/base/__init__.py
rabah-khalek: we said though that we're going to do that, and raise an error with a helpful tip about `pip install <missing-package>`  (https://linear.app/giskard/issue/GSK-944/add-nicer-error-message-if-user-tries-to-import-module-with-missing), I'll take care of it in this branch.

python-client/giskard/models/base/__init__.py
rabah-khalek: `validate_arguments` doesn't seem to do it, but we can later add these lines to the custom decorator `validate_args`
Screenshot 2023-04-07 at 13 19 15

python-client/giskard/core/model_validation.py
rabah-khalek: done here: https://github.com/Giskard-AI/giskard/blob/fddd931f062065e47db3876291c18a6fe4251963/python-client/giskard/core/validation.py#L6-L20

python-client/giskard/models/base/__init__.py
andreybavt: actually `mlflow-skinny` is our core dependency, without it `giskard` can't be installed, so we can be sure that it's available (which is indeed not the case for `torch`, `tensorflow` and `transformers` for example). So the check for `mlflow` isn't required

andreybavt (changes requested):

rabah-khalek:
I did it in case we need to separate giskard installation into a:
- programmatic (where we won't need mlflow at all) and 
- programmatic+UI

andreybavt:
Since it's not yet decided that we split the library in two (in fact last time we said we actually don't split it for the sake of simplicity, but rather relax the dependencies ranges), I'd remove it just to minimize the confusion as we always know that it's going to be installed at the moment. Plus there are still some 
```
try:
    import mlflow
except ImportError:
    pass
```
sections

rabah-khalek:
Ok, you're right. Fixed.

python-client/giskard/core/validation.py
```diff
@@ -1,4 +1,23 @@
 import pandas as pd
+from pydantic import validate_arguments
+import functools
+
+
+def validate_args(func):
```
andreybavt, line 6: since you're not modifying the logic of the underlying original function, you can skip the inner method creation (`def foncigured_decorator`) and just create your `validate_args` like
```
def validate_args(func):
    """
    Decorator to enforce a function args to be compatible with their type hints.
    :return: A wrapper function decorated by pydantic validate_arguments configured to allow arbitrary types check.
    """
    return functools.wraps(func)(
        validate_arguments(config=dict(arbitrary_types_allowed=True))(func)
    )
```

nitpick: I'd also name it after the original decorator - configured_decorator so that they're interchangeable (only the import will have to be changed in case you want to use one or another)

rabah-khalek: done!

PR #915, opened 2023-04-06
Fix Windows python client not working

Googleton:


kevinmessiaen (approved):

python-client/giskard/cli.py
```diff
@@ -144,6 +144,11 @@ def _start_command(is_server, url: AnyHttpUrl, api_key, is_daemon):
         if is_daemon:
             # Releasing the lock because it will be re-acquired by a daemon process
             pid_file.release()
+            # If on windows, throw error and exit
+            if sys.platform == "win32":
+                logger.error("Daemon mode is not supported on Windows.")
+                return
+
```
rabah-khalek, line 151: So external ML-worker won't work on windows right?

rabah-khalek: Okay clear. Maybe we can open an issue with the discussion about the Unix sockets? to include also @andreybavt's comment from slack.

PR #917, opened 2023-04-08
Enhancement/pdm migration

andreybavt:


andreybavt:
Switched from poetry to pdm

PR #920, opened 2023-04-11
Improve setuptools build (GSK-945, closes #919)

### LINKED ISSUE #919: [GSK-945] Generate proto stubs when installing giskard python library
Run whatever is executed in `python-client/build.gradle.kts:55` but during `pip install`

ideally this should be described in `pyproject.toml`

\*\*\*

Tested with @andreybavt : we can use `setuptools` for the build, migrate to standard PEP 621 for the `pyproject.toml`. We then drop poetry for development since it is not compatible with PEP 621.

### Consequences

* we will not need the gradle build to generate the whl, meaning that we can install directly from the git subdirectory
* this will drop the poetry dependency for the package build, hopefully giving less problems with installation (setuptools is often already installed)

<sub>From [SyncLinear.com](https://synclinear.com) | [GSK-945](https://linear.app/giskard/issue/GSK-945/generate-proto-stubs-when-installing-giskard-python-library)</sub>

mattbit:
## Description

Improves `setup.py` by implementing `build.SubCommand` and Windows path support.

## Related Issue

Issue #919 

## Type of Change

- [ ] 🔐 Security fix


python-client/setup.py
```diff
@@ -2,79 +2,91 @@
...
+from setuptools.command.build import SubCommand
 from setuptools.command.build_py import build_py
 
 
-class GrpcTool(Command):
-    user_options = []
-    out_path = 'giskard/ml_worker/generated'
+class GrpcTool(Command, SubCommand):
+    user_options = [
+        ("build-lib=", "d", 'directory to "build" (copy) to'),
```
andreybavt, line 13: do we actually need to parameterize it? I can't imagine when we'd want to use another destination directory

andreybavt (changes requested):
@mattbit , do you think the failing build is related to the `build_grpc` being called from `run` and not from `finalize_options` like before?

mattbit:
> @mattbit , do you think the failing build is related to the `build_grpc` being called from `run` and not from `finalize_options` like before?

No, I think that the build is failing because the output files get written in the build directory (instead of the sources directory). But when installing in dev mode, we exceptionally want the files to ignore the build dir parameter passed by setuptools and write in the current directory together with the sources.

This should be possible if the package is installed in editable mode, but apparently there is some bug in `setuptools` that prevents the editable attribute to be set correctly? I did not have the time to investigate this. That’s the reason I tried to expose the build dir to use `setup.py build_grpc --build-dir ./` when we want to install in dev mode.

Actually the best way would be to use `nox` or `tox` to make a separate installation of the package and run the tests on that environment. That is also what the setuptools docs recommend:
> Editable installs are not a perfect replacement for regular installs in a test environment. When in doubt, please test your projects as installed via a regular wheel. There are tools in the Python ecosystem, like [tox](https://pypi.org/project/tox) or [nox](https://pypi.org/project/nox), that can help you with that (when used with appropriate configuration).

python-client/setup.py
mattbit: See comment below

mattbit:
Will close this for now.

Issue #890, opened 2023-03-30
[GSK-900] issue with classification_labels


rabah-khalek:
![Screenshot 2023-03-30 at 16.20.27.png](https://uploads.linear.app/5cb5c438-2552-41c0-89ac-b08d19e80d42/672d51e4-8fc4-4424-bf02-296a39b68f33/9a29a851-3384-4d82-95b0-95333ec550c8)

* issue with classification_labels provided to the model

  ```
  posts['label'] = posts['label'].astype("int32")
  
  ```
* could be that the ones provided are `str` and the one inferred from `classes_` are int? → add a unit-test on this

<sub>From [SyncLinear.com](https://synclinear.com) | [GSK-900](https://linear.app/giskard/issue/GSK-900/issue-with-classification-labels)</sub>

rabah-khalek:
solved here: https://github.com/Giskard-AI/giskard/pull/924/commits/4148b2c19b9d95d9f64d07098e89e6ace3b11ef8

rabah-khalek:
I removed the `str` casting to classification label(s), and this issue is now solved in https://github.com/Giskard-AI/giskard/pull/924/commits/b4c7fe6c358a47662d992b2f4433ff6de26eb9b8

PR #924, opened 2023-04-12
user-test/part2

rabah-khalek:
- [ ] https://github.com/Giskard-AI/giskard/issues/890

### DIFF
--- a/python-client/giskard/models/base/__init__.py
+++ b/python-client/giskard/models/base/__init__.py
@@ -61,7 +61,7 @@ class BaseModel(ABC):
                 ) from e
 
         if classification_labels is not None:
-            classification_labels = list(classification_labels)
+            classification_labels = [str(i) for i in list(classification_labels)]
             if len(classification_labels) != len(set(classification_labels)):
                 raise ValueError(
                     "Duplicates are found in 'classification_labels', please only provide unique values."
--- a/python-client/tests/test_upload.py
+++ b/python-client/tests/test_upload.py
@@ -72,7 +72,7 @@ def _test_upload_model_exceptions(model: SKLearnModel, ds: Dataset):
                     name=model_name,
                     classification_labels=[0, 1],
                 ).upload(client, "test-project", ds)
-            assert e.match("Values .* in .* column are not declared in classification_labels parameter \[0, 1\] of the model: uploaded model") # noqa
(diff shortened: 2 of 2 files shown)


rabah-khalek:
The typing here: https://github.com/Giskard-AI/giskard/blob/4c0a86808f42c10998c030a0d6cedaf3ffb44069/python-client/giskard/ml_worker/testing/tests/drift.py#L589 will be resolved in https://github.com/Giskard-AI/giskard/pull/935

python-client/giskard/datasets/__init__.py
```diff
@@ -1,17 +1,23 @@
...
-            infer_column_types: Optional[bool] = False,
-            column_types: Optional[Dict[str, str]] = None):
-    return Dataset(df,
+def wrap_dataset(dataset: pd.DataFrame,
+                 name: Optional[str] = None,
+                 target: Optional[str] = None,
+                 cat_columns: Optional[List[str]] = None,
+                 infer_column_types: Optional[bool] = False,
+                 column_types: Optional[Dict[str, str]] = None):
+    console = Console()
```
andreybavt, line 15: I see that you're using a 3rd party library `rich` here. This library isn't directly required by Giskard, but installed as a transitive dependency of `bandit` which is a static analyser (so not installed as part of Giskard's prod dependencies either). There are 2 options:
1. We decide to keep `rich` in which case we should declare it as a depencency of GIskard
2. We stick to standard python print or logging. It's not as convenient to color things, but doesn't require an extra dependency

For now, I'd stick to option 2 to reduce the number of dependencies

python-client/giskard/core/model_validation.py
```diff
@@ -207,7 +207,7 @@ def validate_label_with_target(model_name: str, classification_labels: Union[np.
             )
 
         to_append = " of the model: " + model_name if model_name else ""
-        target_values = target_values if is_string_dtype(target_values) else [str(label) for label in target_values]
+        target_values = [label for label in target_values]
```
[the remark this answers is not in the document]
rabah-khalek: of course!

python-client/giskard/datasets/__init__.py
rabah-khalek: you're right. Done. Kept it simple with a `print` for now.

PR #937, opened 2023-04-17
Scan feature

mattbit:


python-client/giskard/scanner/performance.py
```diff
@@ -0,0 +1,320 @@
...
+        return len(self.issues) > 0
+
+    def __repr__(self):
+        if not self.has_issues():
+            return "<PerformanceScanResult (no issues)>"
+
+        return f"<PerformanceScanResult ({len(self.issues)} issue{'s' if len(self.issues) > 1 else ''})>"
+
+    def _repr_html_(self):
+        tab_header = f"""
```
andreybavt, line 199: as an improvement, moving it to a separate html template and using either a standard python string.Template or something like jinja could be a good idea

mattbit: This will definitely be rewritten with `jinja2` templates, but right now we are rushing. Also the `<iframe>` thing is a hack and will need to be changed. The best thing will be to wrap everything in a jupyter widget, but has been deprioritized for now.

mattbit:
@andreybavt should be ready for review

python-client/giskard/ml_worker/testing/tests/metamorphic.py
```diff
@@ -57,12 +57,16 @@ def _compare_prediction(results_df, prediction_task, direction, output_sensitivi
                 axis=1,
             )
             passed_idx = results_df.loc[results_df["predict_difference_ratio"] < output_sensitivity].index.values
+        else:
+            raise ValueError(f"Invalid prediction task: {prediction_task}")
```
andreybavt, line 61: any reason to prefer manual check over defining types and adding `@configured_validate_arguments`

python-client/giskard/scanner/performance.py
```diff
@@ -0,0 +1,235 @@
...
+                if test_name == "accuracy":
+                    tests.append(test_diff_accuracy)
+                if test_name == "recall":
+                    tests.append(test_diff_recall)
+                if test_name == "precision":
+                    tests.append(test_diff_precision)
+                if test_name == "rmse":
+                    tests.append(test_diff_rmse)
+
+        if model is None:
```
andreybavt, line 64: this can be moved up and be executed in the very beginning

python-client/giskard/scanner/performance.py
```diff
@@ -0,0 +1,235 @@
...
+        for s in slices:
+            for test in tests:
+                test_result = self._diff_test(s, precooked, dataset, test, threshold)
+                if not test_result.passed:
+                    issues.append(Issue(s, model, dataset, dataset_meta, test_result, self._get_test_name(test)))
+
+        return issues
+
+    def _get_test_name(self, test):
+        # @TODO: throw this away when we can
```
andreybavt, line 95: you can already get tests metadata from the registry:
`from giskard.ml_worker.testing.registry.registry import tests_registry`

python-client/giskard/scanner/performance.py
```diff
@@ -0,0 +1,235 @@
...
+        if test.meta.name.endswith("accuracy"):
+            return "Accuracy"
+        if test.meta.name.endswith("recall"):
+            return "Recall"
+        if test.meta.name.endswith("rmse"):
+            return "RMSE"
+        if test.meta.name.endswith("precision"):
+            return "Precision"
+
+    def _diff_test(self, slice_fn, model, dataset, test_fn, threshold):
```
andreybavt, line 107: nitpick: this method can be inlined, especially as it's only used once

python-client/giskard/scanner/performance.py
```diff
@@ -0,0 +1,235 @@
...
+        loss_values = [
+            metrics.log_loss([true_label], [probs], labels=model.meta.classification_labels)
+            for true_label, probs in zip(true_target, pred.raw)
+        ]
+
+        return pd.DataFrame({"__gsk__loss": loss_values}, index=dataset.df.index)
+
+    def _find_slices(self, dataset, model, meta: pd.DataFrame, slicer_name):
+        df_with_meta = dataset.df.join(meta)
+        target_col = "__gsk__loss"
```
andreybavt, line 134: `__gsk__loss` is a good candidate to be extracted as a constant

python-client/giskard/scanner/performance.py
```diff
@@ -0,0 +1,235 @@
...
+
+        # @TODO: FIX THIS
+        # Text features
+        slicer = TextSlicer(dataset_with_meta, target=target_col, slicer=slicer_name)
+        for col in cols_by_type["text"]:
+            slices.extend(slicer.find_slices([col]))
+
+        # @TODO: Should probably bind back to original dataset, but how to to pass
+        # cleanly also the metadata needed for the plots? Maybe need a wrapper object
+        # like Issue or similar.
```
andreybavt, line 167: sounds like a good idea, something like `DatasetSlices` class that combines the two objects: dataset and found slices?

python-client/giskard/scanner/result.py
```diff
@@ -0,0 +1,2 @@
+class ScanResult:
```
andreybavt, line 1: is it planned to be used?

python-client/giskard/slicing/filters.py
```diff
@@ -0,0 +1,106 @@
...
+
+class InternalTestFilter(DataSliceFilter):
+    def __init__(self, model, dataset, test_name=None, threshold=0.1):
+        self.dataset = dataset
+        self.model = model
+        self.threshold = threshold
+
+        if test_name == "f1":
+            self.test = test_diff_f1
+        if test_name == "accuracy":
```
andreybavt, line 29: let's use `elif` to skip the subsequent equality checks once the test is initialized

python-client/giskard/slicing/filters.py
```diff
@@ -0,0 +1,106 @@
...
+        if test_name == "recall":
+            self.test = test_diff_recall
+        if test_name == "precision":
+            self.test = test_diff_precision
+        if test_name == "rmse":
+            self.test = test_diff_rmse
+        if test_name is None:
+            self.test = test_diff_f1
+
+    def _diff_test(self, data_slice):
```
andreybavt, line 40: I'd call it `_test_not_pass` to be more explicit

python-client/giskard/slicing/filters.py
```diff
@@ -0,0 +1,106 @@
...
+from giskard.ml_worker.testing.tests.performance import (
+    test_diff_f1,
+    test_diff_rmse,
+    test_diff_accuracy,
+    test_diff_recall,
+    test_diff_precision,
+)
+
+
+class DataSliceFilter:
```
andreybavt, line 14: this calss should be abstract (inherit ABC)

python-client/giskard/slicing/filters.py
```diff
@@ -0,0 +1,106 @@
...
+    """Selects slices based on their significance."""
+
+    def filter(self, slices):
```
andreybavt, line 17: this method should be marked as abstract (`@abstractmethod`)

python-client/giskard/slicing/stop_words.py
```diff
@@ -0,0 +1,834 @@
+sw_en = ['hers',
```
andreybavt, line 1: I think it's cleaner to store them as csv

python-client/giskard/slicing/tree_slicer.py
```diff
@@ -0,0 +1,99 @@
...
+from typing import Optional, Sequence
+from sklearn.tree import DecisionTreeRegressor
+from sklearn.tree._tree import Tree as SklearnTree
+from sklearn.model_selection import GridSearchCV
+
+from .base import BaseSlicer
+from .slice import Query, LowerThan, GreaterThan, QueryBasedSliceFunction
+
+
+def make_slices_from_tree(data: pd.DataFrame, tree: SklearnTree, feature_names: Optional[Sequence] = None):
```
andreybavt, line 14: `data` isn't used

python-client/giskard/slicing/bruteforce_slicer.py
```diff
@@ -0,0 +1,39 @@
...
+from .slice import Query, LowerThan, GreaterThan, QueryBasedSliceFunction
+
+
+class BruteForceSlicer(BaseSlicer):
+    def __init__(self, dataset: Dataset, features=None, target=None):
+        self.dataset = dataset
+        self.features = features
+        self.target = target
+
+    def find_slices(self, features, target=None):
```
andreybavt, line 14: the signature doesn't match the parent class

andreybavt (changes requested):
I know it's still a POC stage but there are a lot of missing unit tests for scanners/slicers. They're implemented in a pretty modular way so should be easy to test

python-client/giskard/scanner/performance.py
mattbit: Actually this has been done but the comment stayed ;) It’s the `Issue` class.

python-client/giskard/scanner/performance.py
mattbit: will drop soon

Issue #909, opened 2023-04-03
[GSK-924] Refactorings in python-client


rabah-khalek:
<sub>From [SyncLinear.com](https://synclinear.com) | [GSK-924](https://linear.app/giskard/issue/GSK-924/refactorings-in-python-client)</sub>

rabah-khalek:
@andreybavt, since as you mentionned, we're moving things to the programmatic realm, it makes sense to elevate some scripts from `python-client/giskard/ml_worker` to `python-client/giskard`. 

I would mainly like to do the following refactorings:
- `python-client/tests` --> `python-client/unit_tests`
- `python-client/giskard/ml_worker/testing` --> `python-client/giskard/tests`

together with (this is the current structure in [feature/new-upload-api-automodel](https://github.com/Giskard-AI/giskard/tree/feature/new-upload-api-automodel)):
```
python-client/giskard
├── __init__.py
├── cli.py
├── cli_utils.py
├── core
│   ├── __init__.py
│   ├── core.py
│   ├── dataset_validation.py --> python-client/giskard/datasets/.
│   ├── model_validation.py --> python-client/giskard/models/.
│   └── validation.py
├── datasets
│   ├── __init__.py
│   └── base
│       └── __init__.py
├── ml_worker
│   ├── __init__.py
│   ├── core
│   │   ├── model_explanation.py --> depends on whether Matteo will use it in scan or not. 
│   │   ├── suite.py --> python-client/giskard/tests/.
│   │   ├── test_result.py --> python-client/giskard/tests/.
│   │   └── test_runner.py --> python-client/giskard/tests/.
│   └── ml_worker.py
├── models
│   ├── __init__.py
│   ├── base
│   │   └── __init__.py
│   ├── catboost
│   │   └── __init__.py
│   ├── huggingface
│   │   └── __init__.py
│   ├── pytorch
│   │   └── __init__.py
│   ├── sklearn
│   │   └── __init__.py
│   ├── tensorflow
│   │   └── __init__.py
│   └── utils.py
├── path_utils.py
└── settings.py
```
WDYT?

rabah-khalek:
No, I haven't worked on it

rabah-khalek:
semi-done with v2

PR #940, opened 2023-04-19
improve `import giskard` speed

andreybavt:
With a lot of root level imports (e.g. a large list of items in `__all__`  of `giskard.__init__.py`) the `import giskard`  command has degraded a lot its performance. For the moment it takes ~4-6 seconds to `import giskard`.

It’d be less of an issue if we didn’t have a CLI, but with CLI it’s an unacceptable delay (usually an expected response delay for an interactive app is ~200ms) since all of the interactions with `giskard server ...`  or `giskard worker ...`  become very painful.

The solution is to stop adding new items to the root `giskard.__init__.py`  unless it’s _really_ required.

As a first improvement, I moved test functions outside of the main `__init__`, 
**it reduced `import giskard` **from 4 seconds to 0.6 seconds.****

As a next step we can:
- improve the tests imports to make them intuitive too, something like `giskard.testing import test_rmse`
- check again what's taking the time in the root __init__.py and move it outside too


python-client/giskard/__init__.py
```diff
@@ -66,81 +57,10 @@ def get_version() -> str:
...
-    'test_diff_reference_actual_f1',
-    'test_metamorphic_invariance',
-    'test_metamorphic_increasing',
-    'test_metamorphic_decreasing',
-    'test_metamorphic_decreasing_t_test',
-    'test_metamorphic_increasing_t_test',
-    'test_metamorphic_invariance_t_test',
-    'test_metamorphic_increasing_wilcoxon',
-    'test_metamorphic_decreasing_wilcoxon',
-    'test_metamorphic_invariance_wilcoxon',
```
rabah-khalek, line 105: I think it's a good idea, but I don't like the fact that the user has to pull tests now from:
```py
from giskard.ml_worker.testing.tests.X import some_test
```

would it help to only expose at least:
```
giskard.ml_worker.testing.metamorphic
giskard.ml_worker.testing.statistic
giskard.ml_worker.testing.drift
giskard.ml_worker.testing.performance
```
so that a user can do:
```
from giskard.performance import test_f1
```

rabah-khalek:
Another solution is to refactor the tests to a better place from where are right now, on a similar level as `models` and `datasets` as suggested here: https://github.com/Giskard-AI/giskard/issues/909

python-client/giskard/__init__.py
andreybavt: > Another solution is to refactor the tests to a better place from where are right now, on a similar level as `models` and `datasets` as suggested here: #909
> 
> so we'd have something like:
> 
> ```
> from giskard.tests.performance import test_f1
> ```
> 
> or simply:
> 
> ```
> from giskard.tests import test_f1
> ```

yes, for sure, `giskard.tests`  LGTM

mattbit (changes requested):
Good for me if we expose the tests that were previously in the root from `giskard.tests` or `giskard.testing`. But could be part of a later refactoring where we move everything out of the `ml_worker` subpackage.

PR #958, opened 2023-04-21
Remove `reset_index` from tests

### LINKED ISSUE #957: [GSK-997] Remove side effect on dataframe index in Giskard tests
<sub>From [SyncLinear.com](https://synclinear.com) | [GSK-997](https://linear.app/giskard/issue/GSK-997/remove-side-effect-on-dataframe-index-in-giskard-tests)</sub>

mattbit:
Closes #957

## Description

Remove side effect on user `Dataset` in tests.


python-client/tests/testing/test_side_effects.py
```diff
@@ -0,0 +1,34 @@
...
+    dataset.df.set_index(pd.to_datetime(dataset.df.index), inplace=True)
+    original_idx = dataset.df.index.copy()
+
+    _ = test_accuracy(
+        dataset=dataset,
+        model=model,
+        threshold=0.8
+    ).execute()
+
+    assert (dataset.df.index == original_idx).all()
```
rabah-khalek, line 34: Shouldn't we also test if the metric is unchanged due to the omission of this `reset`? @mattbit

mattbit: I would keep this single-responsibility, we already have tests for the metric value.

rabah-khalek: you're right, we do checks on expected_metrics in the unit-test, and these have passed.

PR #962, opened 2023-04-21
[GSK-965] Task: Add turtles and change default green colour

henchaves:
## Description
This PR aims to improve some components, making them more graphically similar.
For example, we're introducing a new primary colour and therefore updated all primary buttons (colour and text colour).
Also, an illustration was added to the Test Suite tab to be displayed when there were no suites created before.

Some components may seem awkward now because of the light green colour, but they will be updated in the following PRs. 

## Related Issue

[GSK-965 (available on Linear)](https://linear.app/giskard/issue/GSK-965/put-turtles-and-green)

## Type of Change

- [ ] 🔐 Security fix


kevinmessiaen:
For me there is a contrast issue here:

It's rated as 'very poor': https://coolors.co/contrast-checker/91f7c0-ffffff

image

kevinmessiaen:
On small screen the left menu disappear but the content does not adjust:

image

Previously it was adjusting:

image

henchaves:
> For me there is a contrast issue here, @jmsquare we should adjust text color or background color to make it more readable, do you have some suggestion?
> 
> It's rated as 'very poor': https://coolors.co/contrast-checker/91f7c0-ffffff
> 
> This is the automatically suggested fix, however we should adapt to fit into Giskard branding: https://coolors.co/contrast-checker/064c26-ffffff
> 
> image
> 
> image

I talked to JM about this. We can use a darker colour (green or grey) for these elements that don't have a dark background. I liked this option #064C26. I'm going to add this as a dark-primary colour, ok?

kevinmessiaen (approved):

PR #964, opened 2023-04-24
Refactoring GiskardTestMethod to make it easier to understand

kevinmessiaen:
## Description

Allow to execute test directly instead of having to call `execute`. To add test in a Suite, the `set_params` method needs to be called instead.

## TODO list

- [x] Update tests
- [ ] Update slicing functions
- [ ] Update transformation functions
- [ ] Update ML worker
- [ ] Update test class to have same structure as annotated @test
- [ ] Update doc


python-client/giskard/ml_worker/core/suite.py
```diff
@@ -224,6 +224,12 @@ def add_test(self, test_fn: Test, test_name: Optional[Union[int, str]] = None, *
         :return: The current instance of the test suite to allow chained calls.
         :rtype: TestSuite
         """
+        if isinstance(test_fn, bool) or isinstance(test_fn, TestResult):
+            raise TypeError("""
+            It seems that you executed your test inside of the 'add_test' method.
+            
+            In order to add a test to a 'Suite', please use the 'set_params' method.
```
mattbit, line 231: In my opinion this message is not very clear. In particular "to add ad test to a Suite, please use set_params", I find it a bit confusing.

python-client/giskard/ml_worker/testing/registry/giskard_test.py
```diff
@@ -95,7 +95,7 @@ def get_builder(self):
 
 
 class GiskardTestMethod(GiskardTest):
-    params = None
+    params = {}
```
mattbit, line 98: Maybe `params = dict()` to be more explicit, because sometimes `{}` could be confused with a `set`.

python-client/giskard/ml_worker/testing/registry/giskard_test.py
```diff
@@ -109,14 +109,18 @@ def __init__(self, test_fn: Function) -> None:
             meta = tests_registry.get_test(test_uuid)
         super(GiskardTest, self).__init__(test_fn, meta)
 
-    def __call__(self, *args, **kwargs) -> 'GiskardTestMethod':
-        self.is_initialized = True
-        self.params = kwargs
+    def __call__(self, *args, **kwargs) -> Result:
+        return self.test_fn(*args, **kwargs)
+
+    def set_params(self, *args, **kwargs) -> 'GiskardTestMethod':
```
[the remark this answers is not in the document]
kevinmessiaen: Actually it's to avoid something like this:

```
test_1 = test_f1.set_param(threshold=0.8)
test_2 = test_f1.set_param(threshold=0.5)

print(test_1.params['threshold']) => 0.5
```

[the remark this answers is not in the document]
kevinmessiaen: For me we should keep `GiskardTestMethod` as immutable for this reason. Maybe we can rename `set_params` to `with_params` or `fixture` to make it less confusing.

python-client/giskard/ml_worker/testing/registry/giskard_test.py
```diff
@@ -109,14 +109,18 @@ def __init__(self, test_fn: Function) -> None:
...
+        new_instance = GiskardTestMethod(self.data)
+        new_instance.is_initialized = True
+        new_instance.params = kwargs
 
         for idx, arg in enumerate(args):
-            self.params[next(iter([arg.name for arg in self.meta.args.values() if arg.argOrder == idx]))] = arg
+            new_instance.params[next(iter([arg.name for arg in self.meta.args.values() if arg.argOrder == idx]))] = arg
```
mattbit, line 121: It’s a bit strange to manipulate the object attribute `params` from outside.

Also, this is not clear:
```
next(iter([arg.name for arg in self.meta.args.values() if arg.argOrder == idx]))
```
I think `iter` is not necessary, could just get the first item if that’s what we need. By why not getting the list of arg names (e.g. by inspection) and then just doing `arg_name[n]`?

andreybavt:
@kevinmessiaen did you plan to refresh this PR?
If not we can keep it closed

PR #967, opened 2023-04-25
Refactoring of scan feature

mattbit:
## Description

- Decouples performance metric from giskard tests
- Adds support for different types of issues
- Refactos the scanner to be arbitrarily extensible and configurable (detectors registry)
- Move the templating to the main `Scanner` object in preparation of additional scan types (e.g. robustness)

Closes GSK-1003, GSK-976.


python-client/giskard/scanner/scanner.py
```diff
@@ -0,0 +1,33 @@
...
+    def analyze(self, model: BaseModel, dataset: Dataset) -> ScanResult:
+        """Runs the analysis of a model and dataset, detecting issues."""
+        validate_model(model=model, validate_ds=dataset)
+
+        # Collect the detectors
+        detectors = self.get_detectors(tags=[model.meta.model_type.value])
+
+        issues = []
+        for detector in detectors:
+            issues.extend(detector.run(model, dataset))
```
andreybavt, line 22: should we parallelize it already?

python-client/giskard/scanner/result.py
```diff
@@ -0,0 +1,67 @@
...
+        tpl = env.get_template("scan_results.html")
+        html = tpl.render(
+            issues=self.issues,
+            num_major_issues=len([i for i in self.issues if i.is_major]),
+            num_medium_issues=len([i for i in self.issues if not i.is_major]),
+        )
+
+        escaped = html.replace('"', "&quot;")
+
+        return f'''<iframe srcdoc="{escaped}" style="width: 100%; border: none;" class="gsk-scan"></iframe>
```
andreybavt, line 39: nitpick: I'd maybe read it from file

python-client/giskard/scanner/templates/_issues_table.html
```diff
@@ -12,19 +12,20 @@
...
-                    <span class="{{'text-red-400' if issue.is_major else 'text-amber-200'}}">{{"{:.2f}".format(issue.test_results.metric * 100) }}% than
-                        global</span>
+                    <span class="{{'text-red-400' if issue.is_major else 'text-amber-200'}}">
+                        {{ issue.deviation }}
+                    </span>
                 </td>
                 <td class="p-3">
                     <span class="text-gray-400">
-                        {{issue.test_results.actual_slices_size[0]}} samples ({{ '{:.2f}'.format(100 * issue.test_results.actual_slices_size[0] / issue.test_results.reference_slices_size[0])}}%)
+                        {{ issue.description }}
```
andreybavt, line 28: we should be careful displaying user-provided content, for the XSS reasons. There's an escaping mechanism (`|e`) like : `{{ dangerous|e }}`

python-client/giskard/scanner/scanner.py
mattbit: Still wondering if it’s better to let the detectors manage the parallelization or to do it at this level. For now we only have a single detector so we can wait.

python-client/giskard/scanner/result.py
mattbit: What do you mean?

python-client/giskard/scanner/templates/_issues_table.html
mattbit: Autoescaping is enabled for HTML but a bug in the iframe generation prevented this from being effective. It is now fixed.

python-client/giskard/scanner/decorators.py
```diff
@@ -0,0 +1,16 @@
+import re
+
+from .registry import DetectorRegistry
+
+
+def detector(name=None, tags=None):
+    def inner(cls):
+        DetectorRegistry.register(name or _to_snake_case(cls.__name__), cls, tags)
+        return cls
```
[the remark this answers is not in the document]
mattbit: It’s a class decorator, not sure how to implement this…

python-client/giskard/scanner/issues.py
```diff
@@ -0,0 +1,36 @@
+from ..datasets import Dataset
+from ..models.base import BaseModel
+
+from typing import Optional
+
+
+class IssueInfo:
```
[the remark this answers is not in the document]
mattbit: Updated to be an abstract class

PR #968, opened 2023-04-25
feature/debug

rabah-khalek:
<details> 
<summary> ToDo </summary>

- [x] spec
- [x] check the outdated [branch](https://github.com/Giskard-AI/giskard/tree/GSK-106_Reduce_the_number_of_test_outputs) of this feature.
- [x] converge on the technical implementation
- [x] to be filled
</details>

<details>
<summary> comment (deprecated) </summary>

I found couple of issues with the previous implementation:
https://github.com/Giskard-AI/giskard/tree/GSK-106_Reduce_the_number_of_test_outputs

1. Our test should be agnostic to the type of `dataset` we use. Once we have CV, we might not have `pandas.DataFrame` anymore. Therefore, the slicing operation shouldn't be only valid for the latter.
2. We should return a `dataset` and not a df that can be ready to be rendered in the UI (with new name, uuid, etc.).
3. The `debug_filters` along which we define the `slices_to_debug` should not be duplicated and defined per test, rather collected somewhere for readability and better handling.

Here's my first preliminary proposal:
- a function that returns a fresh_copy of a dataset:
https://github.com/Giskard-AI/giskard/blob/3d9ae1e0b20185979b5fab8ec6d977147e2a23a1/python-client/giskard/ml_worker/testing/tests/debug_utils.py#L13-L17
- creating a filter catalogue where each filter has the name of the test it belongs to (valid for any type of dataset we support):
https://github.com/Giskard-AI/giskard/blob/3d9ae1e0b20185979b5fab8ec6d977147e2a23a1/python-client/giskard/ml_worker/testing/tests/debug_utils.py#L20-L26
- adding `slices_to_debug` to the `TestResult` by using the `debug_filters` catalogue:
https://github.com/Giskard-AI/giskard/blob/3d9ae1e0b20185979b5fab8ec6d977147e2a23a1/python-client/giskard/ml_worker/testing/tests/performance.py#L45-L49 
so basically:
  - `debug_filters.get(test_name)` will get the filter for the corresponding test.
  -  `(gsk_dataset, prediction)` are the parameters needed by this filter.
  - returns a list of slices (`Dataset`s) assigned to `slices_to_debug`. 

This will result in:
Screenshot 2023-04-12 at 14 40 44

@andreybavt and @jmsquare let me know what you think about this. 
</details>

<details> <summary> minutes of 12/04/2023 </summary>

After the 12/04/2023 discussion with @jmsquare and @andreybavt :

- [x] add `debug` arg to the tests 
- [x] The filter has to return a list of either booleans or ints
- [x] <s> The name slice for `Dataset` might not be ideal -- to spec </s> (deprecated)
- [x] <s> The `catalogue` idea to review by @andreybavt </s> (abandonned)
- [x] For next week we should have at least one representative `slice` per type of tests:
  - [x]   performance
  - [x]   statistical
  - [x]   metamorphic
  - [x]   drift

</details>

https://linear.app/giskard/project/test-debug-feature-0024fb6da8c5?filter=eyJhbmQiOlt7InN0YXRlIjp7Im5hbWUiOnsibmluIjpbIkRvbmUiXX19fV19


common/proto/ml-worker.proto
```diff
@@ -233,6 +234,7 @@ message SingleTestResult{
 
   repeated uint32 actual_slices_size = 21;
   repeated uint32 reference_slices_size = 22;
+  string output_df_id = 23;
```
andreybavt, line 237: where is it used?

frontend/src/views/main/project/SuiteTestExecutionList.vue
```diff
@@ -136,5 +140,54 @@ async function removeTest(suiteTest: SuiteTestDTO) {
...
+
+    inputs.push({
+      name: "dataset",
+      type: "Dataset",
+      isAlias: false,
+      value: result.inputs["dataset"],
+      params: []
+    });
+
+    // name: arg.name,
```
andreybavt, line 166: let's cleanup commented code

frontend/src/views/main/project/SuiteTestExecutionList.vue
```diff
@@ -136,5 +140,54 @@ async function removeTest(suiteTest: SuiteTestDTO) {
...
+  let inputs = [];
+  if (props.execution.inputs.length > 0) {
+    inputs = props.execution.inputs;
+  } else {
+    // This is a bit hacky, but there is no other way because the test suite input list is never guaranteed to be present :(
+    inputs.push({
+      name: "model",
+      type: "BaseModel",
+      isAlias: false,
+      value: result.inputs["model"],
```
andreybavt, line 154: we should actually be looking at argument type, not name. If I upload a test whose input is called "my_model" this won't work

frontend/src/views/main/project/SuiteTestExecutionList.vue
```diff
@@ -1,37 +1,37 @@
...
+              <span>{{ getTestName(suiteTest.test) }}</span>
+              <div>
+                <v-btn
+                    v-if="result !== undefined && !result.passed"
+                    text
+                    color="green"
+                    @click.stop="debugTest(result, suiteTest)"
+                >
+                  <v-icon>mdi-bug</v-icon>
+                  Debug
```
andreybavt, line 24: as we discussed, you should only be able to debug a test that has at least 1 model and 1 dataset as inputs. For therest of the cases we should disable the buttong with a tooltip explaining why it's disabled

frontend/src/views/main/project/SuiteTestExecutionList.vue
```diff
@@ -136,5 +140,54 @@ async function removeTest(suiteTest: SuiteTestDTO) {
...
+    // value: '',
+    // params: []
+  }
+
+  let model = inputs.filter(i => i.name == "model")[0].value;
+
+  let res = await api.runAdHocTest(testSuiteStore.projectId!, suiteTest.testUuid, inputs, true);
+  let dataset = res.result[0].result.outputDfUuid;
+
+  const debuggingSession = await api.prepareInspection({
```
andreybavt, line 178: you should also pass a project key. It'll become unreliable to take project id from either dataset or model as soon as we decide to share models/datasets across projects. Where possible we try to specify a project key explicitly

andreybavt (changes requested):

frontend/src/views/main/project/SuiteTestExecutionList.vue
andreybavt: not doing it ATM

frontend/src/views/main/project/SuiteTestExecutionList.vue
rabah-khalek: data drift tests do not need a model @Googleton @andreybavt

common/proto/ml-worker.proto
rabah-khalek: @Googleton could you answer this one?

rabah-khalek: According to @Googleton : 

> it’s what we use for the frontend to know which debugging session to open

python-client/giskard/ml_worker/testing/registry/giskard_test.py
```diff
@@ -129,7 +129,13 @@ def __call__(self, *args, **kwargs) -> "GiskardTestMethod":
     def execute(self) -> Result:
         analytics.track("test:execute", {"test_name": self.meta.full_name})
 
-        return self.test_fn(**self.params)
+        try:
+            return self.test_fn(**self.params)
```
andreybavt, line 133: you can inspect the `self.test_fn` and see if it has `debug` argument instead of trying to run it and catching and exception

andreybavt (changes requested):

PR #970, opened 2023-04-25
Task/fix data processor mutables

mattbit:
## Description

Fixing bug in data processor where mutables are initialized at a class level instead of per each instance.


mattbit:
@andreybavt just noticed that this is still not merged

andreybavt:
it slipped between the cracks, merged now

PR #981, opened 2023-04-28
Feature/new-custom-model

rabah-khalek:
A new class between `BaseModel` and `WrapperModel`:
- `BaseModel` don't have model as attribute.
- `WrapperModel` have `model`, `data_preprocessing_function` and `model_postprocessing_function`.

We need a new class that:
- takes only `model` as attribute in additional to all attributes of `BaseModel`.
- detects `model` library and uses the correct serialization methods

This new class will be used by the user similar to `CustomModel` but with an automatic inference on the method of serialisation.


rabah-khalek:
todo:

- [ ] Refactor: model_predict -> predict_proba
- [ ] Find better naming for `AutoSerializableModel` and change the abstract models name if necessary:
  - BaseModel
  - WrapperModel
  - MLFlowBasedModel
  - CustomModel
- [ ] Create a unit-test based on this notebook: https://colab.research.google.com/drive/1rAwTUqCrgaKrw9FblCr5KViI8mYv9rOq?usp=sharing
- [ ] Add it to the documentation.

rabah-khalek:
@andreybavt, I know have only one global class called `Model` which automatically infer the `model` library and uses: 
1. the correct `save_model` and `load_model` (which, most importantly, can be overriden).
2. the correct `model_predict` (which, most importantly, can be overriden).

I've put `should_save_model_class=True` by default, so we don't need to have another class called `CustomModel`.

Since no need for `wrap_X`, I also refactored the Dataset class to the way we had it, making symmetrical to `Model`.

rabah-khalek:
@andreybavt I just realised that in order for the default `model_predict` to work:
```py
    def model_predict(self, df: pd.DataFrame):
        giskard_class = getattr(importlib.import_module(self.meta.loader_module), self.meta.loader_class)
        return giskard_class.model_predict(df)
```
I need to pass additional meta information as attributes:
- when it's `pytorch`
```py
            torch_dtype: TorchDType = "float32",
            device="cpu",
```
- when it's `huggingface`:
```py
huggingface_module
pipeline_task
```

PR #977, opened 2023-04-27
[GSK-961] Feature: Create Project Properties

henchaves:
## Description

The idea of this PR is to create a new page that will aggregate three old tabs: Overview, Datasets and Models.
A button will always be displayed within the project so the user can navigate to this page whenever he wants. 
The idea is to reduce the number of existing tabs so that the user can focus on the most important functionalities.

## Related Issue

[GSK-961 (available on Linear)](https://linear.app/giskard/issue/GSK-961/make-the-overview-page-not-a-tab-but-a-project-properties-button) and [GSK-962(available on Linear)](https://linear.app/giskard/issue/GSK-962/dataset-and-model-tab-should-not-be-at-the-same-level-as-the-other)

## Type of Change

- [ ] 📚 Examples / docs / tutorials / dependencies update
- [ ] 🔧 Bug fix (non-breaking change which fixes an issue)
- [ ] 🥂 Improvement (non-breaking change which improves an existing feature)
- [x] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🔐 Security fix

PR #998, opened 2023-05-02
[GSK-963] Improvement: Reorder the tabs

henchaves:
## Description
This PR does basically two things: move tabs to the sidebar and reorder them.
The idea is to reduce the number of buttons on the container, and also make the tabs more visible.

## Related Issue

[GSK-963 (available on Linear)](https://linear.app/giskard/issue/GSK-963/reorder-the-tabs)
This PR depends on PR #977 to be merged before. 

## Type of Change

- [ ] 🔐 Security fix


kevinmessiaen:
For me it seems unintuitive that we have to click on Giskard logo to go back to the project list in this screen. Maybe we should add a button `Projects` 

I also think that the properties button should be on the left bar on not on the upper right corner (it took me time to find the screen) 

image

PR #1000, opened 2023-05-04
Improving the models module, introducing CloudPickleBasedModel

rabah-khalek:
Supersedes: 
- https://github.com/Giskard-AI/giskard/pull/981
- [composition-logic branch](https://github.com/Giskard-AI/giskard/blob/feature/new-custom-model-part1/python-client/giskard/models/automodel/__init__.py)

Todo:
- [x] Implement `CloudpickleBasedModel` to be used as fall-back.

Todo in another branch (https://github.com/Giskard-AI/giskard/pull/1004):
- [x] Consider the case of arbitrary model as unit-test (cloud-pickle)
- [x] Consider the case of sklearn model as unit-test (mlflow without additional meta)
- [x] Consider the case of pytorch model as unit-test (mlflow with additional meta)
- [x] Consider the case of huggingface model as unit-test (huggingface with additional meta)
- [ ] doc
- refactor (models/base/__init__.py is way too heavy) 
  - since `wrap_model` and `wrap_dataset` are no longer the main methods (in favour of the new `Model` and `Dataset`) -> to be revisited: `Dataset` takes `df` as arg while `Model` takes `model` as arg -- to symmetrise
  - `automodel` package that contains `Model` class
- find a way to allow auto-completion of the `WrapperModel` methods inside `Model`.
- currently `should_save_model_class=True` by default (since the class size is ~ 4 KB) It's a small price that is worth paying to avoid additional documentation about `should_save_model_class` class attribute. We can think of smart ways to infer it in the future (not urgent).


andreybavt (commented):
Generally speaking changes LGTM. The last thing to fix is the autocompletion of methods of Model class

PR #1002, opened 2023-05-05
GSK-895 Dataset sample & predict caching

kevinmessiaen:
## Description

- [x] Create a sample dataset whenever uploading the dataset (max 1000 rows)
- [x] Allow to choose between sample and whole data when creating debugging session
- [x] Run all catalog functions on sample dataset for quick preview
- [x] Allow to try a test suite on sample dataset (no persistance in database) 
- [x] Cache model predictions
- [x] Various UI improvement

## Type of Change


kevinmessiaen:
TODO: test on this dataset: https://github.com/Giskard-AI/giskard-examples/blob/task/test-fraud-detection-notebook/ieee_fraud_detection_adversarial_validation.ipynb

python-client/giskard/models/base/__init__.py
```diff
@@ -244,30 +261,40 @@ def predict(self, dataset: Dataset) -> ModelPredictionResults:
...
         """
         timer = Timer()
-        df = self.prepare_dataframe(dataset)
 
-        raw_prediction = self.predict_df(df)
+        # Read cache
+        cached_predictions = self.model_cache.read_from_cache(dataset.df[GISKARD_HASH_COLUMN])
+        missing = pd.isna(cached_predictions)
+        if len(missing.shape) > 1:
+            missing = np.any(missing, axis=1)
```
mattbit, line 269: It’s better if the cache returns a dataframe with a single column or a `pd.Series`, and your `isna()` would be sufficient. Otherwise, `cached_predictions.isna().any(axis=1)` will directly give you what you want.

python-client/giskard/models/base/__init__.py
```diff
@@ -244,30 +261,40 @@ def predict(self, dataset: Dataset) -> ModelPredictionResults:
...
+
+        df = self.prepare_dataframe(dataset.slice(lambda x: dataset.df[missing], row_level=False))
+
+        if len(df) > 0:
+            raw_prediction = self.predict_df(df)
+            self.model_cache.set_cache(dataset.df[missing][GISKARD_HASH_COLUMN], raw_prediction)
+            cached_predictions[missing] = raw_prediction
```
mattbit, line 277: Let’s make sure that the naming here goes back to `raw_predictions`, the code after this should not be aware of the cache and should behave as usual (before your modifications).

python-client/giskard/models/base/__init__.py
```diff
@@ -358,12 +396,25 @@ def download(cls, client: GiskardClient, project_key, model_id):
                     loader_class=file_meta["loader_class"],
                 )
 
+                try:
```
mattbit, line 399: This code is present twice (both in `if` and in `else`), maybe just pull it ouside of the `if else` blocks?

python-client/tests/automodel/test_automodel.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 import pandas as pd
 import pytest
+
```
mattbit, line 4: ?

python-client/tests/fixtures/german_credit_scoring.py
```diff
@@ -64,7 +64,7 @@ def german_credit_catboost_raw_model(german_credit_data):
 
     columns_to_encode = [key for key in column_types.keys() if column_types[key] == "category"]
 
-    credit = german_credit_data.df
+    credit = german_credit_data.df[german_credit_data.columns]
```
mattbit, line 67: This should not be needed

mattbit (changes requested):
I did a partial review of the caching (in Python). In my opinion it’s better to make some changes, most importantly:
- Keep the hashes separated from the user’s dataframe (currently a new column is added to `Dataset::df`), as we do with other metadata (e.g. `dataset/metadata`). This means not touching the data inside `df` but creating a separate `pd.Series` or `DataFrame` with the same index that we use to store the hashes.
- Add some tests for the caching: model prediction is an extremely sensible part of our workflow and we need to be sure there are no glitches.
- Only lazily evaluate the hashes when we need.
- Refactor the `ModelCache` class to make it simpler and faster (see comments in the file), isolating it from model logic (classification labels, model type, etc.)
- Allow disabling the cache

I’ll let @andreybavt comment on whether this should be addressed now or later, but I see the first two points as a hard requirement because they could cause bugs elsewhere in the code.

python-client/giskard/datasets/base/__init__.py
```diff
@@ -358,7 +393,7 @@ def extract_column_dtypes(df):
         Returns:
             dict: A dictionary where the keys are the column names and the values are the corresponding data types as strings.
         """
-        return df.dtypes.apply(lambda x: x.name).to_dict()
+        return {key: value for key, value in df.dtypes.apply(lambda x: x.name).to_dict().items()}
```
mattbit, line 396: Why do we need this?

python-client/giskard/datasets/base/__init__.py
```diff
@@ -204,13 +223,29 @@ def add_transformation_function(self, transformation_function: TransformationFun
         self.data_processor.add_step(transformation_function)
         return self
 
+    def dataset_hash(self):
+        if not self._dataset_hash_initialized:
```
mattbit, line 227: Do we need the `_dataset_hash_initialized`? Just checking if `self._hashes is None` would do, right?

python-client/giskard/models/base/__init__.py
```diff
@@ -286,6 +306,23 @@ def predict_df(self, df: pd.DataFrame):
         """
         ...
 
+    def _predict_from_cache(self, dataset: Dataset):
+        dataset_hash = dataset.dataset_hash()
+        cached_predictions = self.model_cache.read_from_cache(dataset_hash)
+        missing = cached_predictions.isna()
```
mattbit, line 312:
```py
missing = cached_predictions.isna().any(axis=1)
```

python-client/giskard/models/base/__init__.py
```diff
@@ -286,6 +306,23 @@ def predict_df(self, df: pd.DataFrame):
...
+        if len(missing.shape) > 1:
+            missing = missing.any(axis=1)
+
+        df = self.prepare_dataframe(dataset.slice(lambda x: dataset.df[missing], row_level=False))
+
+        if len(df) > 0:
+            raw_prediction = self.predict_df(df)
+            self.model_cache.set_cache(dataset_hash[missing], raw_prediction)
+            cached_predictions.loc[missing] = raw_prediction.tolist()
```
mattbit, line 321:
```py
if missing.any():
    df = self.prepare_dataframe(…)
    cached_predictions.loc[missing] = self.predict_df(…)
    self.cache.set(…)
```

python-client/giskard/models/base/cache.py
```diff
@@ -0,0 +1,48 @@
...
+            yield x
+
+
+class ModelCache:
+    prediction_cache: Dict[str, Any]
+
+    vectorized_get_cache_or_na = None
+
+    def __init__(self, cache: Optional[pd.DataFrame] = None):
+        cache = cache if cache is not None else pd.DataFrame(data={}, index=[])
```
mattbit, line 23:
```py
cache = cache or pd.DataFrame()
```

python-client/giskard/models/base/cache.py
```diff
@@ -0,0 +1,48 @@
...
+
+        self.prediction_cache = cache.to_dict()
+
+        self.vectorized_get_cache_or_na = np.vectorize(self.get_cache_or_na, otypes=[object])
```
mattbit, line 27: I still believe it is faster to use a `pd.Series` as pointed out here: https://github.com/Giskard-AI/giskard/pull/1002#discussion_r1200479832
It will also make the code more compact and readable.

python-client/giskard/models/base/__init__.py
```diff
@@ -107,6 +117,8 @@ def __init__(
                     "Duplicates are found in 'classification_labels', please only provide unique values."
                 )
 
+        self.model_cache = ModelCache()
```
mattbit, line 120: Make private? `self._cache`?

python-client/tests/fixtures/german_credit_scoring.py
```diff
@@ -122,7 +122,7 @@ def german_credit_raw_model(german_credit_data):
     clf = Pipeline(steps=[("preprocessor", preprocessor), ("classifier", LogisticRegression(max_iter=100))])
 
     Y = german_credit_data.df["default"]
-    X = german_credit_data.df.drop(columns="default")
+    X = german_credit_data.df[german_credit_data.columns].drop(columns="default")
```
mattbit, line 125: not necessary

python-client/tests/testing/test_side_effects.py
```diff
@@ -23,6 +23,9 @@ def test_dataset_index_is_preserved(german_credit_data, german_credit_model):
 
     # Try with an index that is not a sequence of integers
     dataset.df.set_index(pd.to_datetime(dataset.df.index), inplace=True)
+    # Reset hash since it is based on index
+    dataset._dataset_hash = None
+    dataset._dataset_hash_initialized = False
```
mattbit, line 28: Right, that’s not your fault, the test was badly written (sorry!). Better to create a new dataset before the second test, e.g. `dataset = Dataset(dataset.df.set_index(…), column_types=dataset.column_types, …)`

python-client/tests/models/test_model_cache.py
```diff
@@ -0,0 +1,91 @@
...
+        return np.array(df.values)
+
+    wrapped_model = Model(
+        prediction_function,
+        model_type="regression"
+    )
+    wrapped_model.disable_cache = True
+
+    wrapped_dataset = Dataset(
+        df=pd.DataFrame([0, 1])
```
mattbit, line 48: Would also try with a dataset having multiple columns, repeated rows, etc.

python-client/tests/models/test_model_cache.py
```diff
@@ -0,0 +1,91 @@
...
+        return df
+
+    transformed_dataset = wrapped_dataset.transform(replace_row_one, row_level=False)
+    second_prediction = wrapped_model.predict(transformed_dataset)
+
+    assert called_indexes == list(wrapped_dataset.df.index) + [
+        1], 'The  prediction should have been called once for row 0 and twice of row 1'
+    assert list(prediction.raw) != list(second_prediction.raw)
+    assert list(prediction.raw_prediction) != list(second_prediction.raw_prediction)
+    assert list(prediction.prediction) != list(second_prediction.prediction)
```
mattbit, line 91: Should also add a test of the `ModelCache` object

python-client/tests/models/test_model_cache.py
mattbit, line 1: Should also test the dataset hashes generation somewhere

mattbit (changes requested):

python-client/giskard/datasets/base/__init__.py
```diff
@@ -204,13 +223,29 @@ def add_transformation_function(self, transformation_function: TransformationFun
         self.data_processor.add_step(transformation_function)
         return self
 
+    def dataset_hash(self):
```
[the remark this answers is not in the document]
kevinmessiaen: Well some transformation partially modify the dataset: (eg. add typo with a small rate). So its small optimisation but add a bit of complexity.

But anyway I need to compute the difference using `pd.ne` to reset the cache. So it probably slow the performances in most cases.

mattbit (approved):

mattbit:
@andreybavt if for you the backend part is ok we can merge this.

andreybavt (changes requested):

andreybavt (approved):
LGTM

andreybavt:
@kevinmessiaen , let me know if you don't have time to resolve the conflicts

PR #1009, opened 2023-05-11
Task/classification labels mapping

rabah-khalek:
Allowing the user to pass a "`id2label`" map to `classification_labels`. For instance:
```py
classification_labels = {0: "Default", 1: "Not default"}
```
which will display `"Default"` and `"Not default"` in the UI for the credit scoring dataset defined like this:
```py
    _classification_labels = {"Default": 0, "Not default": 1}
    gsk_dataset.df[gsk_dataset.target] = gsk_dataset.df[gsk_dataset.target].apply(lambda x: _classification_labels[x])
```


rabah-khalek:
This still needs to be properly propagated to the backend

andreybavt:
@rabah-khalek what's the use-case for this feature? When a user messes up an order of classification_labels?

rabah-khalek:
It’ll help this a bit, but no it’s suppose to be a solution for the label being not “human readable” issue. This dict will do the mapping internally without the need for the user to change his df. Its Done in python, but still needed one of you in order to propagate it through the other parts of the code. Do you think it’s a lot of work?On 24 May 2023, at 17:54, Andrey Avtomonov ***@***.***> wrote:﻿
@rabah-khalek what's the use-case for this feature? When a user messes up an order of classification_labels?

andreybavt:
@rabah-khalek 
Actually, I don’t think I understand the use case for this PR. While previously we were creating `predicted_labels` by using a list and taking elements from it by their positions, now we use a dictionary and take elements by key which is still a position of the column in the prediction matrix. 

There was indeed a problem when people used `classification_labels` different to what they actually stored in the target column or used an incorrect order (which this PR doesn't solve). You'd still have to change `classification_labels`, but instead of changing the order of elements in the list you have to change dictionary keys.

There was a feature request (from Max) that was about supporting a mapping between a "technical" prediction and a human-readable explanation. For example in case you have a target columns containing classes "A", "B" and "C" you need your `model.predict(...).predicted_labels` to still return a list of values from {A, B, C} in order to calculate metrics, but to show this in the UI you'd use human-readable representation of each code using a user-provided mapping:
`{"A": "Iris Setosa", 
"B": "Iris virginica",
"C": "Iris versicolor"}`

rabah-khalek:
This branch doesn't have the goal to solve the "order" of classification_labels issue.

It supposed to solve what Max is talking about. But I also added on the python side, the action that, if you provide a dict, I suppose that you want to do a mapping between your model output (via the classification_labels dict) and your target variable.

If you think we should not allow the user to do a mapping between his model and target variable (the issue we faced in this [notebook](https://github.com/princyiakov/customer_conversion/blob/main/notebooks/task_2_FeatureSelection_Modelling_Pipeline.ipynb)) -- which is what HF does with `id2label`, then it becomes purely a UI feature, and they keys of the dict must match the target variable of the df.

mattbit:
@rabah-khalek I had just a quick look but I see an issue with the internal usage of `raw_prediction.argmax(axis=1)` to determine the label index which does not seem to be addressed in the code. We expect labels to be in order 0, 1, 2, etc. but with dict support this would break.

andreybavt:
> but with dict support this would break
IIRC dicts guarantee order or element in recent Python, so it shouldn't be an issue in principle. 

However this PR only adds a support of `Dict[int, str]` as a new classification_labels type. So the case when your model returns a not-so-human-readable prediction, but not necessarily a number from 0 to N you can't map it to a human-readable label that you'd show in the UI. This was the case of Max, he was using encoded targets, so you could roughly understand what the target means but it was far from being readable.

And yes, I think this is more of a UI feature

rabah-khalek:
yes, the argmax shouldn't be a problem, but that means that the dict need to also be ordered correctly.

The PR as it stands, not only supports `dict` type for `classification_labels` + allows a mapping between the model output and the `dataset.target`.
Take the credit scoring example, and say the `dataset.target` had `[0, 1]` instead of `['Default', 'Not default']`, passing this dict `classification_labels={0: 'Default', 1: 'Not default'}` will make `PredictionResults` cast to the dict values, resulting in the same output we get usually from the default credit scoring example, this can be summarised by this unit-test: https://github.com/Giskard-AI/giskard/blob/021ad16d749a23595939493a6385ee32000dbc3a/python-client/tests/test_classification_labels.py#L41-L57 

So in principle, this will propagate the `classification_labels` mapping to the UI for predictions, but not the `dataset.target` which is what i meant here: https://github.com/Giskard-AI/giskard/pull/1009#issuecomment-1545862439.

My end goal, was to make the dict `classification_labels` cast both `model predictions` and `dataset.target` from the `key` to the `values` provided in the dict.

With or without this casting, there's some work to do on the UI, if you think it's not that crucial before the launch, I'll put this PR on hold. We can discuss it further after the launch.

PR #1013, opened 2023-05-16
Task/gsk 1035 add lang detection

mathieuroques:
I used detect_lang and I support english and french


mathieuroques:
Not the right branch

PR #1015, opened 2023-05-16
implemented partial inference of column_types

rabah-khalek:
This is now possible:
```py
    valid_df = pd.DataFrame({"categorical_column": ['turtle', 'crocodile', 'turtle'],
                         "text_column": ['named Giskard', 'a nile crocodile', 'etc'],
                         "numeric_column": [15.5, 25.9, 2.4]})
    valid_df_column_types = {'categorical_column': 'category', 'text_column': 'text', 'numeric_column': 'numeric'}

    column_types = {'categorical_column': 'text'}
    expected_df_column_types = {'categorical_column': 'text', 'text_column': 'text', 'numeric_column': 'numeric'}

    # Option 1: only one column in column_types is provided
    my_dataset = Dataset(valid_df, column_types=column_types)
    assert my_dataset.column_types == expected_df_column_types

    # Option 2: one column in column_types is provided, one in cat_columns
    my_dataset = Dataset(valid_df, column_types=column_types, cat_columns=['categorical_column'])
    assert my_dataset.column_types == valid_df_column_types
```


python-client/giskard/datasets/base/__init__.py
```diff
@@ -315,39 +312,39 @@ def _infer_column_types(self, cat_columns: List[str]):
         Returns:
             dict: A dictionary that maps column names to their inferred types, one of 'text', 'numeric', or 'category'.
         """
-        column_types = {}
-        nuniques = self.df.nunique()
-        df_columns = list(self.df.columns)
+        _column_types = column_types.copy()
+        df_columns = set(self.df.drop([self.target], axis=1).columns) if self.target else set(self.df.columns)
```
mattbit, line 316: Hey this will copy the whole dataframe! I think you can `df_columns = df.columns.drop(self.target)`). Just make sure that `self.target` is in fact in `df.columns` (I don’t know if this is validated before or after).

python-client/giskard/datasets/base/__init__.py
```diff
@@ -315,39 +312,39 @@ def _infer_column_types(self, cat_columns: List[str]):
...
                 raise ValueError(
                     "The provided 'cat_columns' are not all part of your dataset 'columns'. "
                     "Please make sure that `cat_columns` refers to existing columns in your dataset."
                 )
-
             for cat_col in cat_columns:
                 if cat_col != self.target:
-                    column_types[cat_col] = SupportedColumnTypes.CATEGORY.value
-            df_columns = set(df_columns) - set(cat_columns)
+                    _column_types[cat_col] = SupportedColumnTypes.CATEGORY.value
```
mattbit, line 327: Maybe more concise
```py
_column_types.update({col: SupportedColumnTypes.CATEGORY for col in cat_columns})
```

mattbit: Should we check if `cat_columns` are compatible with the `column_types` declared?

python-client/giskard/datasets/base/__init__.py
```diff
@@ -315,39 +312,39 @@ def _infer_column_types(self, cat_columns: List[str]):
         Returns:
             dict: A dictionary that maps column names to their inferred types, one of 'text', 'numeric', or 'category'.
         """
-        column_types = {}
-        nuniques = self.df.nunique()
-        df_columns = list(self.df.columns)
+        _column_types = column_types.copy()
```
mattbit, line 315: Maybe we can just shadow at the beginning and keep a single `column_types`:
```py
column_types = column_types.copy()
```
Just to avoid possible future mistakes if one forgets to always use `_column_types` in the following.

mattbit (changes requested):
Great! Little changes for performance/robustness and some optional suggestion.

mattbit:
Just noticed that some tests fail with
```
ValueError: The provided keys ['target'] in 'column_types' are not part of your dataset 'columns'. Please make sure that the column names in `column_types` refers to existing columns in your dataset.
```
Maybe we should get rid of this. When it’s about the target, this is also confusing because **it is** part of the dataframe that the user put in, so from their point of view the message may be confusing.

It’s good to raise a warning, but in my opinion if we can continue execution we should do it. In a case like this, it means that the user specified extra column types, in principle we have all what we need to create the dataset, and can just throw a warning saying
```
WARNING: The column types for `feature_a`, `feature_b` were not taken into account since these columns were not found in the dataframe.
```

rabah-khalek:
This error was mainly raised in the above because I removed by mistake this line:
```py
column_types.pop(self.target, None)  # no need for target type
```
Otherwise, I think it's a useful error, In case the user-provided columns in `column_types` assuming they have them in their df, but turns out that they don't... that could be seen as a debugging message to investigate the df.

mattbit:
> 



> Otherwise, I think it's a useful error, In case the user-provided columns in `column_types` assuming they have them in their df, but turns out that they don't... that could be seen as a debugging message to investigate the df.

I agree but I think a warning would be better for this case, since it is not a fatal problem.

python-client/giskard/datasets/base/__init__.py
```diff
@@ -328,38 +323,41 @@ def _infer_column_types(self, cat_columns: List[str]):
         Returns:
             dict: A dictionary that maps column names to their inferred types, one of 'text', 'numeric', or 'category'.
         """
-        column_types = {}
-        nuniques = self.df.nunique()
-        df_columns = list(self.df.columns)
+        if not column_types:
+            column_types = {}
```
mattbit, line 327: Either this is useless, or the type hint in the function signature is not correct.

mattbit (changes requested):

mattbit (approved):

PR #1022, opened 2023-05-24
task/Numpy encoder for classification_labels

rabah-khalek:
Solves the `Object of type int64 is not JSON serializable` issue mainly encountered in `sklearn` and `catboost` models where `.classes_` is a numpy array.

### DIFF
--- a/python-client/giskard/models/utils.py
+++ b/python-client/giskard/models/utils.py
@@ -1,5 +1,7 @@
 import itertools
 from collections.abc import Iterator
+from typing import List
+import numpy as np
 
 
 def map_to_tuples(data: Iterator):
@@ -10,3 +12,12 @@ def map_to_tuples(data: Iterator):
         return data
 
     return map(lambda x: (x,), data)
+
+
+def numpy_encoder(l: List):
+    e = l[0]
+    if isinstance(e, np.integer):
+        return [int(i) for i in l]
+    if isinstance(e, np.floating):
+        return [float(i) for i in l]
+    return l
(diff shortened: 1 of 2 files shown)


python-client/giskard/models/utils.py
```diff
@@ -10,3 +12,12 @@ def map_to_tuples(data: Iterator):
         return data
 
     return map(lambda x: (x,), data)
+
+
+def numpy_encoder(some_list: List):
+    if some_list is not None:
+        if isinstance(some_list[0], np.integer):
+            return [int(i) for i in some_list]
```
andreybavt, line 20: there's a [.item()](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.item.html) method that can do it for numerous numpy types that we can reuse

andreybavt: also since we're iterating over the whole array anyway it's safer to do the check on each element:

```
def numpy_encoder(some_list: List):
    if some_list is None:
        return None
    else:
        return [i.item() if isinstance(i, np.generic) else i for i in some_list]
```

python-client/giskard/models/utils.py
```diff
@@ -10,3 +12,12 @@ def map_to_tuples(data: Iterator):
         return data
 
     return map(lambda x: (x,), data)
+
+
+def numpy_encoder(some_list: List):
```
andreybavt, line 17: I'd rename it to something less cryptic like `np_types_to_native` or something similar to give a hint of what it does

andreybavt (changes requested):

python-client/giskard/models/utils.py
rabah-khalek: I didn't know about this one, thanks!

andreybavt (approved):

PR #1027, opened 2023-05-25
GSK-935 Support regression models in scan

mattbit:
Depends on #1025


python-client/giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -58,10 +67,37 @@ def _calculate_meta(self, model, dataset):
...
+            elif loss == "mse":
+                loss_values = [
+                    metrics.mean_squared_error([y_true], [y_pred]) for y_true, y_pred in zip(true_target, pred.raw)
+                ]
+            elif loss == "mae":
+                loss_values = [
+                    metrics.mean_absolute_error([y_true], [y_pred]) for y_true, y_pred in zip(true_target, pred.raw)
+                ]
+            else:
+                raise ValueError(f"Invalid loss function `{loss}`.")
```
kevinmessiaen, line 100: This code shouldn't be reachable since we validate the loss method before

kevinmessiaen (approved):

PR #1040, opened 2023-05-30
task/wrapping_bound_prediction_functions

rabah-khalek:
Fixing: https://linear.app/giskard/issue/GSK-1131/model-wrapper-fails-to-wrap-function-object

Now it's possible to do:
```py
from giskard import Model
wrapped_model = Model(
  model=clf.predict_proba,
  model_type="classification",
  classification_labels=clf.classes_, # Their order MUST be identical to the prediction_function's output order
  feature_names=list(feature_types.keys()),  # Default: all columns of your dataset
  # name="my_iris_classification_model", # Optional
  # classification_threshold=0.5, # Default: 0.5
)
```

I tested it's upload too.


python-client/giskard/models/__init__.py
```diff
@@ -26,6 +26,13 @@ def get_class(_lib, _class):
 
 
 def infer_giskard_cls(model: Any):
+    # bound-methods case (like model.predict_proba)
+    if hasattr(model, '__self__'):
+        # wrap bound-method with a function
+        def prediction_function(df): return model(df)
+
+        model = prediction_function
+
```
mattbit, line 35: I don’t get it, why don’t we keep the bound method? `inspect.ismethod` can help.

mattbit (changes requested):

mattbit (approved):
I added a test, if it’s okay for you @rabah-khalek you can merge.

PR #1042, opened 2023-05-30
Task/inverted classification labels validation

rabah-khalek:
fixes: https://linear.app/giskard/issue/GSK-1113/warning-if-the-classification-label-are-inverted-check-if-the-accuracy

Now the user gets the following warning during the scan:
```py
from giskard.demo import titanic  # for demo purposes only 🛳️
original_model, df = titanic()  # Replace with your dataframe creation

from giskard import Model, Dataset
cl = original_model.classes_
giskard_model = Model(original_model, model_type="classification", name="Titanic model", classification_labels=[cl[1], cl[0]])
giskard_dataset = Dataset(df, target="Survived", name="Titanic dataset")

from giskard.core.model_validation import validate_model
validate_model(giskard_model, giskard_dataset)
```
```
UserWarning: The accuracy of your model is very low (0.21). Make sure you have not inverted the order of the 'classification_labels' when you created the Giskard Model.
```


python-client/giskard/core/model_validation.py
```diff
@@ -250,3 +253,13 @@ def validate_classification_prediction(classification_labels: Union[np.ndarray,
         )
     if prediction.shape[1] != len(classification_labels):
         raise ValueError("Prediction output label shape and classification_labels shape do not match")
+
+
+def validate_order_classifcation_labels(model, dataset):
+    from giskard.testing.tests.performance import test_accuracy
+    result = test_accuracy(model=model, dataset=dataset).execute()
+    if result.metric <= 0.4:
```
andreybavt, line 261: for binary classification we can even do a more clever check by trying to flip the order ourselves and see if it improves the metric

mattbit: @rabah-khalek @andreybavt I think there just an easy, definitive check that we can do: balanced accuracy. For a random predictor, it should always be $1 / N_\text{classes}$ (right @rabah-khalek?).
Thus if we find balanced accuracy < $1 / N_\text{classes}$ it means that the model is worse than a random predictor, i.e. is actively making bad predictions which means that labels must be inversed.

mattbit (changes requested):
All good but I think we should just look for `balanced_accuracy <= 1 / len(classification_labels)`, no need to reverse labels and check again.
Could be done in one libe with `sklearn.metrics.balanced_accuracy_score(y_true, y_pred)` where `y_true = dataset.df[dataset.target]` and `y_pred = model.predict(dataset).prediction`.

(Correct me if I’m wrong concerning `balanced_accuracy`)

rabah-khalek:
Ok I used the balanced accuracy as @mattbit suggested, this is now the warning:
```
UserWarning: The balanced accuracy of your model is very low (0.24). Make sure you have not inverted the order of the 'classification_labels' when you created the Giskard Model.
```
Should we give more explanation? I'd rather keep it short like this.

mattbit (approved):
Good for me!

PR #1058, opened 2023-05-31
feature/gsk-1145-Added-warning-for-non-unique-indexes

mathieuroques:
- Scan fails with the next error: Fixed


python-client/giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -48,6 +48,12 @@ def run(self, model: BaseModel, dataset: Dataset):
             logger.debug(f"PerformanceBiasDetector: Limiting dataset size to {max_data_size} samples.")
             dataset = dataset.slice(lambda df: df.sample(max_data_size, random_state=42), row_level=False)
 
+        if not dataset.df.index.is_unique:
+            logger.warning("PerformanceBiasDetector: Resetting dataframe indexes because values are not unique")
```
mattbit, line 52: Two things:
- Not a logger warning, but a real warning! `python_utils.warning`
- We should do this at the very start of the scan, in `scanner.Scanner`

mattbit (changes requested):

mattbit (changes requested):
@mathieuroques can you add a test?

mattbit (approved):

PR #1061, opened 2023-06-01
[GSK-1173] Create a new test suite inside Catalog/Tests

henchaves:
## Description

This PR aims to add a new feature within the AddTestToSuite modal.
The idea is to enable the user to create a new test suite from the Test suite lists if there is no test suite created before or even if the user just wants to separate the current test from the others.

## Related Issue

[GSK-1173 (available on Linear)](https://linear.app/giskard/issue/GSK-1173/new-test-suite-inside-catalog)

## Type of Change

- [ ] 🔐 Security fix


frontend/src/views/main/project/TestSuites.vue
```diff
@@ -48,9 +48,17 @@ onActivated(async () => await testSuitesStore.loadTestSuites(props.projectId))
 
 async function createTestSuite() {
     const project = await api.getProject(props.projectId)
+    const date = new Date();
+    const formattedDate = date.getFullYear() + '-' +
```
kevinmessiaen, line 52: we should use `formatDate` inside of  `filters`. Plus there is better way to format by using format template `YYYY-MM-DD HH:mm:ss`

frontend/src/views/main/project/modals/AddTestToSuite.vue
```diff
@@ -10,8 +10,9 @@
           <v-card-text>
             <v-row>
               <v-col cols=12>
-                <ValidationProvider name="Tes suite" mode="eager" rules="required" v-slot="{ errors }">
-                  <v-select outlined label="Test suite" v-model="selectedSuite" :items="testSuites" :item-text="'name'" :item-value="'id'" dense hide-details></v-select>
+                <ValidationProvider name="Test suite" mode="eager" rules="required" v-slot="{ errors }">
+                  <v-select outlined label="Test suite" v-model="selectedSuite" :items="[...testSuites, { id: 0, name: '+ New test suite' }]" :item-text="'name'" :item-value="'id'" dense hide-details @change="checkIfCreateNew">
```
kevinmessiaen, line 14: It's cleaner to use slot to happen item: https://v2.vuetifyjs.com/en/components/selects/#append-and-prepend-item

frontend/src/views/main/project/modals/AddTestToSuite.vue
```diff
@@ -99,11 +101,40 @@ async function submit(close) {
...
+async function checkIfCreateNew() {
+  if (selectedSuite.value === 0) {
+    await createTestSuite();
+  }
+}
+
+async function createTestSuite() {
+  const project = await api.getProject(projectId)
+  const date = new Date();
+  const formattedDate = date.getFullYear() + '-' +
```
kevinmessiaen, line 118: Here also we should use format method

kevinmessiaen (changes requested):

frontend/src/views/main/project/TestSuites.vue
henchaves: Actually, I'm gonna remove this date from the name, because the user will be able to change the suite name from Testing tab.

kevinmessiaen (approved):

PR #1065, opened 2023-06-01
task/fix nan values error

rabah-khalek:
fixes: https://linear.app/giskard/issue/GSK-1130/problem-with-cast-column-to-dtypes


python-client/giskard/models/model_explanation.py
```diff
@@ -102,7 +102,7 @@ def get_list_words_weights(exp):
 
 def background_example(df: pd.DataFrame, input_types: Dict[str, str]) -> pd.DataFrame:
     example = df.mode(dropna=False).head(1)  # si plusieurs modes, on prend le premier
-    example.fillna("", inplace=True)
+    # example.fillna("", inplace=True)
```
rabah-khalek, line 105: @andreybavt this was causing issues on the front-end side, is there a reason why it's there? it doesn't seem to break any unit-test upon removal

kevinmessiaen (approved):

PR #1108, opened 2023-06-02
task/conserve url punctuation

rabah-khalek:
fixes: https://linear.app/giskard/issue/GSK-1083/http-or-hashtags-should-not-be-stripped-by-punctuation-removal

The solution is based on: https://stackoverflow.com/questions/44015316/how-to-get-rid-of-punctuation-while-maintaining-url

but we can also try: https://pypi.org/project/urlextract/

But Gruber's pattern seems to do the job:
```py
from giskard.ml_worker.testing.functions.transformation import strip_punctuation
from giskard import Dataset
import pandas as pd

df = pd.DataFrame({"text": "This.., is my site..,. http://www.example.com/, and .,.,. this site.,.,. http://stackoverflow.com rules!.."}, index=[0])

ds = Dataset(df)
stripped_ds = ds.transform(strip_punctuation, column_name="text")
print(stripped_ds.df['text'].iloc[0])
```
```
Output: This is my site http://www.example.com/ and  this site http://stackoverflow.com rules
```

### DIFF
--- a/python-client/giskard/ml_worker/testing/functions/transformation.py
+++ b/python-client/giskard/ml_worker/testing/functions/transformation.py
@@ -37,24 +37,7 @@ nearbykeys = {
 }
 
 gruber = re.compile(
-    r"""(?i)\b((?:https?:(?:/{1,3}|[a-z0-9%])|[a-z0-9.\-]+[.](?:com|net|org|edu|gov|mil|aero|asia|biz|cat|coop|info|int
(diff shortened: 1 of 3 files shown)


mattbit (changes requested):
Sorry @rabah-khalek, this referred to the scan transformations. Don’t ask me why we have double transformations functions, but they were unbeknowingly worked on in parallel.

The one causing problems is in `giskard.scanner.robustness.text_transformations.TextPunctuationRemovalTransformation`.

rabah-khalek:
propagated it to the scanner, opened a card to remove duplicates after the launch https://linear.app/giskard/issue/GSK-1191/duplicates-of-transformation-functions-in-scanner-and-registry

mattbit (approved):

PR #1113, opened 2023-06-02
Do not scan columns which are not in model features

mattbit:


mattbit:
Closing this since we have included it in #1064

PR #1114, opened 2023-06-02
fixed auc metric

rabah-khalek:
Fixing AUC score according to the doc in the binary case: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html


AbSsEnT (approved):
Nice job!

AbSsEnT:
Approved

mattbit:
@rabah-khalek can you check that this is ok also in `scanner.performance.metrics.AUC`?

andreybavt (approved):

rabah-khalek:
Thanks for pointing it out @mattbit, I fixed it in the scan, let's see if it breaks any tests.

PR #1132, opened 2023-06-06
Preparation for relase

mattbit:
- Update readme for  PyPI
- Update docs with `giskard==2.0.0b2` → needs to be updated with the correct version


mattbit:
Ready for me, demo here: https://test.pypi.org/project/giskard/2.0.0b2.dev5/

PR #1134, opened 2023-06-06
Added missing model type to allow upload of langchain models

kevinmessiaen:
## Description

Added missing model type to allow upload of langchain models


andreybavt:
@kevinmessiaen , don't forget to re-run j2ts ;)

andreybavt (approved):

PR #1135, opened 2023-06-06
[GSK-1241] Fix type casting in drift tests

mattbit:


mattbit:
Actually this does not fix the issue because there is a `validate_classification_label` decorator

PR #1137, opened 2023-06-06
fixing class saving

rabah-khalek:
fixes (in the UI):
```
Traceback (most recent call last):
  File "/Users/rak/Documents/giskard_repos/giskard_1/python-client/giskard/ml_worker/utils/request_interceptor.py", line 50, in wrapper
    res = await loop.run_in_executor(pool, behavior, request, context)
  File "/usr/local/Caskroom/miniconda/base/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/rak/Documents/giskard_repos/giskard_1/python-client/giskard/ml_worker/server/ml_worker_service.py", line 377, in runModel
    model = BaseModel.download(self.client, request.model.project_key, request.model.id)
  File "/Users/rak/Documents/giskard_repos/giskard_1/python-client/giskard/models/base/__init__.py", line 427, in download
    clazz = cls.determine_model_class(meta, local_dir)
  File "/Users/rak/Documents/giskard_repos/giskard_1/python-client/giskard/models/base/__init__.py", line 196, in determine_model_class
    return getattr(importlib.import_module(meta.loader_module), meta.loader_class)
AttributeError: module '__main__' has no attribute 'GiskardModelCustomWrapper'
```

### DIFF
--- a/python-client/giskard/models/automodel/__init__.py
+++ b/python-client/giskard/models/automodel/__init__.py
@@ -125,6 +125,7 @@ class Model(CloudpickleBasedModel, ABC):
                 possibly_overriden_cls = cls
                 possibly_overriden_cls.save_model = giskard_cls.save_model
                 possibly_overriden_cls.load_wrapped_model = giskard_cls.load_wrapped_model
+                possibly_overriden_cls.should_save_model_class = True
             elif giskard_cls:
                 input_type = "'prediction_function'" if giskard_cls == PredictionFunctionModel else "'model'"
                 logger.info(


rabah-khalek:
in a nutshell, the issue was that `should_save_model_class` was declared True by default as a class attribute, but if the class is overriden, we first infer the predefined the giskard model class (SKLearnModel, etc.) and then replace its methods by the one of the overriden class... except that the inferred giskard model class has `should_save_model_class=False` by default... so in case of overriding, it needs to be set to True so we can save the modified version of the class and load it back.

rabah-khalek:
I have tested it with 2 notebooks so far (titanic and tripadvisor), it works. 

Although in tripadvisor, I'm facing another issue, that we notified @kevinmessiaen about:
Screenshot 2023-06-06 at 22 55 51

andreybavt (approved):

PR #1138, opened 2023-06-06
Fix unsupported syntax

mattbit:


andreybavt:
just pushed a similar fix

PR #1143, opened 2023-06-06
Add Google Colab hint to all pages

Googleton:
Screenshot 2023-06-07 at 01 03 15


alexcombessie (changes requested):
@Googleton @jmsquare I tried it on a new fresh colab, it fails with this error:
CleanShot 2023-06-07 at 09 52 11


I guess the command is not meant to be executed in the Colab? Could we improve the wording to explain it? What is the right solution?

alexcombessie (changes requested):
@Googleton this syntax will not work if the user open the notebook in their colab.
`[documentation](../../guides/installation_app/index.html))`

You need to put the real HTTPS link: https://docs.giskard.ai/en/latest/guides/installation_app/index.html

jmsquare (approved):
The message is clear, ready to go!

jmsquare (changes requested):
@Hugo can you change the link of th Colab

jmsquare (approved):

PR #1151, opened 2023-06-07
Fixed text suite executions responsiveness

kevinmessiaen:
## Description

Fixed text suite executions responsiveness


andreybavt:
works fine

PR #1155, opened 2023-06-07
Feature/gsk 1233 when a test is buggy the whole test suite does not return

kevinmessiaen:
## Description

Handle test error when executing test suite 

## Type of Change

- [ ] 🔐 Security fix


python-client/giskard/core/suite.py
```diff
@@ -224,11 +225,19 @@ def run(self, verbose: bool = True, **suite_run_args):
...
-            if verbose:
-                print("""Executed '{0}' with arguments {1}: {2}""".format(test_partial.test_name, test_params, result))
+            try:
+                test_params = self.create_test_params(test_partial, suite_run_args)
+                result = test_partial.giskard_test.get_builder()(**test_params).execute()
+                res.append((test_partial.test_name, result))
+                if verbose:
+                    print("""Executed '{0}' with arguments {1}: {2}""".format(test_partial.test_name, test_params,
+                                                                              result))
+            except Exception:
```
andreybavt, line 235: can you add # noqa NOSONAR to make IDE and Sonar happy?

andreybavt (changes requested):

python-client/giskard/core/suite.py
```diff
@@ -224,11 +225,25 @@ def run(self, verbose: bool = True, **suite_run_args):
...
-                print("""Executed '{0}' with arguments {1}: {2}""".format(test_partial.test_name, test_params, result))
+            try:
+                test_params = self.create_test_params(test_partial, suite_run_args)
+                result = test_partial.giskard_test.get_builder()(**test_params).execute()
+                res.append((test_partial.test_name, result))
+                if verbose:
+                    print(
+                        """Executed '{0}' with arguments {1}: {2}""".format(test_partial.test_name, test_params, result)
+                    )
+            except BaseException as e:  # noqa SONAR
```
andreybavt, line 236:
```suggestion
            except BaseException as e:  # noqa NOSONAR
```

python-client/giskard/core/suite.py
```diff
@@ -224,11 +225,25 @@ def run(self, verbose: bool = True, **suite_run_args):
...
+                error = traceback.format_exc()
+                logging.exception("An error happened during test execution")
```
andreybavt, line 238:
```suggestion
                logging.exception("An error happened during test execution", e)
```

andreybavt (changes requested):

python-client/giskard/core/suite.py
kevinmessiaen: Well actually when logging exception in python we don't need to pass the exception and it will log it automatically, I got an error if I pass as in the code snippet.

https://docs.python.org/3/library/logging.html#logging.Logger.exception

PR #1164, opened 2023-06-09
[GSK-1235] Fix segmentation fault problems when running pytest

mattbit:
Should solve the problems with fixtures `tripadvisor` and `fraud_detection` when running pytest locally on certain environments. @AbSsEnT could you check that this also solves the problem on your machine?


andreybavt (changes requested):

AbSsEnT:
@mattbit will check it in 20 minutes

AbSsEnT:
I am still getting serrault for the `fraud_detection` fixture.

andreybavt:
@AbSsEnT , can you test it again to see if you still get a segfault?

AbSsEnT:
> @AbSsEnT , can you test it again to see if you still get a segfault?

@andreybavt still getting segfault for the fraud_detection fixture.

mattbit:
> @mattbit , do we still need to make changes to the lightgbm version to solve Mykyta's issue?

It would be useful because it removes a few unnecessary steps (download of stopwords which are not used), avoids downloading and initializing DistilBERT at import time, and frees memory since model is now loaded in the test scope.

andreybavt (approved):

PR #1169, opened 2023-06-13
[GSK-1275] Importance of metrics calculated on partial data slice

### LINKED ISSUE #1159: [GSK-1275] Importance of metrics calculated on partial data slice
User on reddit pointed out that

> And does the # samples refer to the # samples in the data slice? It should be the denominator of the metric for the slice. For example, if that second row's recall of 0.111 is 1 predicted positive / 9 true positive, it's debatable whether to flag that.

This is right. We may have 1000 samples in our data slice, but to calculate for example the recall we only use the positive samples, which may be just a few samples out of the total, making the detection a false positive.

<sub>From [SyncLinear.com](https://synclinear.com) | [GSK-1275](https://linear.app/giskard/issue/GSK-1275/importance-of-metrics-calculated-on-partial-data-slice)</sub>

mattbit:
Fixes #1159.


python-client/giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -153,23 +214,53 @@ def _detect_for_metric(
...
+    sliced_dataset = dataset.slice(slice_fn)
+    slice_metric = metric(model, sliced_dataset)
+
+    if not with_pvalue:
+        return sliced_dataset, slice_metric, None
+
+    # Perform statistical tests
+    complementary_dataset = dataset.slice(lambda df: df[~df.index.isin(sliced_dataset.df.index)], row_level=False)
+    comp_metric = metric(model, complementary_dataset)
+    pvalue = None
```
andreybavt, line 243: you don't necessarily have to initialize it here since it's down in every branch below

python-client/giskard/scanner/performance/metrics.py
```diff
@@ -67,57 +83,78 @@ class F1Score(SklearnClassificationScoreMixin, ClassificationPerformanceMetric):
     greater_is_better = True
     _sklearn_metric = "f1_score"
 
+    def _calculate_affected_samples(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel) -> int:
+        if model.is_binary_classification:
+            # F1 score will not be affected by true negatives
+            neg = model.meta.classification_labels[0]
```
andreybavt, line 89: Why do we only do it for binary classification and not for all cases?

andreybavt (commented):
Also LGTM with some questions

PR #1172, opened 2023-06-13
Await for backend to be ready before showing link

kevinmessiaen:
## Description

Await for backend to be ready before showing link

## Type of Change

- [ ] 🔐 Security fix


henchaves (approved):
I didn't manage to properly test it. It's necessary to build a new Docker image

python-client/giskard/commands/cli_server.py
```diff
@@ -96,6 +96,30 @@ def get_container(version=None, quit_if_not_exists=True) -> Optional[Container]:
...
 
 
+def _backend_ready(endpoint) -> bool:
+    try:
+        response = requests.get(endpoint)
+
+        if response.status_code != 200:
+            return False
+
+        return "UP" == response.json()['components']['readinessState']['status']
```
andreybavt, line 106: I'm testing on dev setup, but my json structure isn't the same, I don't have `components` `readinessState`

image

python-client/giskard/commands/cli_server.py
```diff
@@ -96,6 +96,30 @@ def get_container(version=None, quit_if_not_exists=True) -> Optional[Container]:
...
+    except OSError:
```
andreybavt, line 107: `requests` can throw other errors (`ConnectTimeout` for example), I think we can simply catch BaseException and return False if it happens. (make sure to handle `KeyboardInterrupt` separately and reraise `click.Abort()`)

In this case you can even replace 
```
        if response.status_code != 200:
            return False
```

with 

```
response.raise_for_status()
```

since it'll be caught by the same BaseException handler

python-client/giskard/commands/cli_server.py
```diff
@@ -96,6 +96,30 @@ def get_container(version=None, quit_if_not_exists=True) -> Optional[Container]:
...
+    except KeyError:
+        return False
+
+
+def _await_backend_ready(port: int):
+    endpoint = f"http://localhost:{port}/management/health"
+    backoff_time = 2
+    up = False
+
+    while not up:
```
andreybavt, line 118: I think we should have a waiting limit. Let's say we wait for 3 minutes and if `_backend_ready` still returns False we break the loop and print an explanation like:

```
Giskard backend takes unusually long time to start, please check the logs with `giskard server logs backend`
```

Also I think it's worth adding a feedback while waiting, for example printing dots every `backoff_time` seconds, like this it's clear that the process isn't frozen, but is actually checking the status
```
click.echo(".", nl=False)
```
should do the trick, because print(end='') won't work in the loop without explicit stdout flushing

andreybavt (changes requested):

python-client/giskard/commands/cli_server.py
kevinmessiaen: I got a different format, will use status instead to be sure.
 
image

python-client/giskard/commands/cli_server.py
```diff
@@ -91,6 +92,33 @@ def get_container(version=None, quit_if_not_exists=True) -> Optional[Container]:
...
+
+
+def _wait_backend_ready(port: int) -> bool:
+    endpoint = f"http://localhost:{port}/management/health"
+    backoff_time = 2
+    max_duration_second = 3 * 60
+    started_time = time.time()
+    up = False
+
+    while not up or time.time() - started_time > max_duration_second:
```
andreybavt, line 113: it should actually be 

```suggestion
    while not up and time.time() - started_time <= max_duration_second:
```

python-client/giskard/commands/cli_server.py
```diff
@@ -91,6 +92,33 @@ def get_container(version=None, quit_if_not_exists=True) -> Optional[Container]:
...
 
 
+def _is_backend_ready(endpoint) -> bool:
+    try:
+        response = requests.get(endpoint)
+        response.raise_for_status()
+        return "UP" == response.json()["status"]
+    except KeyboardInterrupt:
+        raise click.Abort()
+    except BaseException:
```
andreybavt, line 102: you can add 
```suggestion
    except BaseException:  # noqa NOSONAR
```
to let Sonar know that it's OK

andreybavt (changes requested):
Last batch of comments and then it looks good to me

python-client/giskard/commands/cli_server.py
```diff
@@ -115,7 +143,16 @@ def _start(attached=False, version=None):
...
-    logger.info(f"Giskard Server {version} started. You can access it at http://localhost:{port}")
+
+    up = _wait_backend_ready(port)
+
+    if up:
+        logger.info(f"Giskard Server {version} started. You can access it at http://localhost:{port}")
+    else:
+        logger.warning(
+            "Giskard backend takes unusually long time to start, please check the logs with `giskard server "
+            "logs backend"
```
andreybavt, line 154: looks like a missing closing **`** after "backend"

andreybavt: or perhaps to make it even more readable you can break the long line before the `giskard server logs backend` command, for example after the first comma

andreybavt (changes requested):

PR #1175, opened 2023-06-14
GSK-1272- Filter slicing functions clauses when column name are not available

kevinmessiaen:
## Description

Filter slicing functions clauses when column name are not available
In slicing function catalog, filter dataset for clause based functions

## Type of Change

- [ ] 🔐 Security fix


andreybavt:
@kevinmessiaen, we still don't separate uploaded slicing functions per project. This causes a problem for example with slicing function upload:

The snippet below fails on last row execution because `SlicingFunctionController` isn't ready to take `project/{projectKey}` inside urls

This is because today we don't store a link between a slicing function and a project, but we should. Then, when accessing a list of slicing functions in the UI we should filter based on whether a given slicing function belongs to a project or not

(same probably applies to transformation functions)


```
import pandas as pd

import giskard
from giskard import slicing_function

# Replace this with your own data & model creation.
df = giskard.demo.titanic_df()
data_preprocessing_function, clf = giskard.demo.titanic_pipeline()

# Wrap your Pandas DataFrame
giskard_dataset = giskard.Dataset(
    df=df,
    target="Survived",
    name="Titanic dataset",
    cat_columns=["Pclass", "Sex", "SibSp", "Parch", "Embarked"],
)


# Wrap your model
def prediction_function(df):
    preprocessed_df = data_preprocessing_function(df)
    return clf.predict_proba(preprocessed_df)


giskard_model = giskard.Model(
    model=prediction_function,
    model_type="classification",
    name="Titanic model",
    classification_labels=clf.classes_,
    feature_names=[
        "PassengerId",
        "Pclass",
        "Name",
        "Sex",
        "Age",
        "SibSp",
        "Parch",
        "Fare",
        "Embarked",
    ],
)

# Create a Giskard client
token = "XXX"  # API Access Token
client = giskard.GiskardClient(
    url="http://localhost:9000", token=token  # URL of your Giskard instance
)


@slicing_function(row_level=False)
def my_func1(df: pd.DataFrame):
    df["Age"] = df["Age"] > 1
    return df


my_func1.upload(client, "slicing")

```

andreybavt:
@kevinmessiaen , I fixed the project-level slicing function upload, there's just the part of filtering of the UI left. Inside the project you should only see global artefacts and the ones uploaded to this project.

kevinmessiaen:
> @kevinmessiaen , I fixed the project-level slicing function upload, there's just the part of filtering of the UI left. Inside the project you should only see global artefacts and the ones uploaded to this project.

@andreybavt it's now handled in the UI as well as some improvement in the backend

andreybavt (changes requested):
@kevinmessiaen , generally the PR looks good to me. Could you just replace the modified `getCatalog` and `createSlicingFunction` in `api.ts` by the autogenerated `api-v2.ts` methods and remove both of them from api.ts. Like this we don't add extra work for the migration in the future

PR #1182, opened 2023-06-19
[GSK-1300] Reorganize breadcrumbs

henchaves:
## Description
This PR aims to change how breadcrumbs are displayed.
The idea is to move the title of the current tab to below the breadcrumbs, making it more visible to the user.

## Related Issue

[GSK-1300 (available on Linear)](https://linear.app/giskard/issue/GSK-1300/reorganize-breadcrumbs)

## Type of Change

- [ ] 🔐 Security fix


kevinmessiaen (changes requested):
I'm not a big fan that we put some part of the breadcrumb in the project home and then continue it inside of each component with a hack to but it in the right location.

For me better solution would be to have:
- A Breadcrumb component that is called directly in each component
- A dynamic component in ProjectHome based on some route props giving the correct route name

frontend/src/views/main/project/TestSuites.vue
```diff
@@ -501,4 +501,10 @@ onActivated(async () => {
     height: 50px;
     margin: 0 auto;
 }
+
+::v-deep .v-toolbar__content {
```
kevinmessiaen, line 505: Do we still need this style?

frontend/src/views/main/project/FeedbackList.vue
```diff
@@ -254,4 +254,10 @@ div.v-input.flex-1 {
   background-color: rgba(49, 110, 62, 0.2);
   cursor: default;
 }
+
+::v-deep .v-toolbar__content {
```
kevinmessiaen, line 258: Same here do we need to keep this style?

frontend/src/views/main/project/Debugger.vue
```diff
@@ -256,4 +258,10 @@ onActivated(async () => {
 .expansion-panel {
   margin-top: 10px !important;
 }
+
+::v-deep .v-toolbar__content {
```
kevinmessiaen, line 262: Do we need to keep this style?

kevinmessiaen (approved):

PR #1185, opened 2023-06-20
Fixed metamorphic tests for LLM

kevinmessiaen:
## Description

Fixed metamorphic tests for LLM

## Type of Change

- [ ] 🔐 Security fix


python-client/giskard/testing/tests/metamorphic.py
```diff
@@ -14,6 +16,7 @@
 from giskard.ml_worker.utils.logging import timer
 from giskard.models.base import BaseModel
 from giskard.models.utils import fix_seed
+from giskard.scanner.llm.utils import LLMImportError
```
mattbit, line 19: Maybe we should move `LLMImportError` outside of scanner now that we are integrating it more with the rest of the codebase. But not super important.

mattbit (approved):
Good for me!

mattbit:
@kevinmessiaen would be nice to add a simple test if you have time, using `FakeListLLM` from `langchain`.

python-client/giskard/testing/tests/metamorphic.py
```diff
@@ -61,6 +64,22 @@ def _compare_prediction(results_df, prediction_task, direction, output_sensitivi
         if prediction_task == SupportedModelTypes.CLASSIFICATION:
             passed_idx = results_df.loc[results_df["prediction"] == results_df["perturbed_prediction"]].index.values
 
+        elif prediction_task == SupportedModelTypes.TEXT_GENERATION:
+            try:
+                import evaluate
+            except ImportError as err:
+                raise LLMImportError() from err
+
+            scorer = evaluate.load("bertscore")
```
andreybavt, line 73: It's better to add `evaluate.load("bertscore")` to the try-except block since it relies on another extra dependency that we add. In case a person already has evaluate (previously installed for example) but not bert-score he'd also get the same `LLMImportError` with  explanations

andreybavt (changes requested):

PR #1187, opened 2023-06-21
Fixed postgres 13 installation

kevinmessiaen:
## Description

Fixed postgresql 13 installation in docker image to avoid issue when the version is being bumped automatically

## Type of Change

- [ ] 🔐 Security fix

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ RUN apt-get update && \
     apt-get -y install wget gnupg nginx openjdk-17-jre-headless supervisor git gettext-base
 
 RUN wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -44,8 +44,12 @@ ENV SPRING_PROFILES_ACTIVE=prod
 ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y \
-    postgresql nginx openjdk-17-jre-headless supervisor git gettext-base
+    apt-get -y install wget gnupg nginx openjdk-17-jre-headless supervisor git gettext-base
```
andreybavt, line 47: we can add a repo before doing a first `apt-get update`  and then just do the update once, can't we?
You'll be also able to simply add `postgresql-13` to the existing `apt-get install` list

andreybavt (changes requested):

Dockerfile
kevinmessiaen: I'll still need to install `wget` and `gnupg` before in order to add the postgres repo

PR #1189, opened 2023-06-21
[GSK-1321] Integration with MLflow

rabah-khalek:
<details> <summary> <b> 1. mlflow Plug-in via <code>evaluate</code> </b> </summary> 

## Description
Integration of `mlflow` via the `model_evaluator` [plugin](https://mlflow.org/docs/latest/plugins.html). 

Installation requirements:
```
pip install mlflow # mlflow is needed for .evaluate() not mlflow-skinny shipped with giskard
pip install "git+https://github.com/Giskard-AI/giskard.git@gsk-1321/mlflow-integration#subdirectory=python-client" -q
```

Code example:
```py
import mlflow

from giskard import demo
model1, df = demo.titanic(max_iter=5)
model2, df = demo.titanic(max_iter=100)

with mlflow.start_run(run_name="model1") as run1:
    model1_uri = mlflow.sklearn.log_model(model1, "sklearn_model1", pyfunc_predict_fn="predict_proba").model_uri
    mlflow.evaluate(model=model1_uri, model_type="classifier", data=df, targets="Survived", evaluators="giskard", evaluator_config={"classification_labels": ["no", "yes"]})

with mlflow.start_run(run_name="model2") as run2:
    model2_uri = mlflow.sklearn.log_model(model2, "sklearn_model2", pyfunc_predict_fn="predict_proba").model_uri
    mlflow.evaluate(model=model2_uri, model_type="classifier", data=df, targets="Survived", evaluators="giskard", evaluator_config={"classification_labels": ["no", "yes"]})
```

Running `mlflow ui` in the terminal, one gets: 
- the html results of scan embedded as **Artifacts**.
- the test suite result of scan logged as **Metrics**.
- model and dataset artifacts

## Two notebooks to test this feature:
Run them locally in order to run the `mlflow ui`
- titanic demo (sklearn): https://colab.research.google.com/drive/1j2w7MtTThXej8fEagdYUxwbeSnMrg9Cc?usp=sharing
- text classification (TensorFlow): https://colab.research.google.com/drive/1MbWVLYZeryKtv_VrTJ_zlYdNoB_BFlaf?usp=sharing

</details>
<details> <summary> <b> 2. Giskard API via <code>to_mlflow</code> </b> </summary>

## Description

logging artifacts and metrics from giskard to mlflow

Code example:
```py
scan_results = giskard.scan(giskard_model, giskard_dataset)
test_suite = results.generate_test_suite("My first test suite")
test_suite_results = test_suite.run()

import mlflow

# Option 1 (via the fluent API)
with mlflow.start_run() as run:
    giskard_model.to_mlflow()
    giskard_dataset.to_mlflow()
    scan_results.to_mlflow()
    test_suite_results.to_mlflow()

# Option 2 (via MlflowClient)
from mlflow import MlflowClient

client = MlflowClient()
experiment_id = "0"
run = client.create_run(experiment_id)

giskard_model.to_mlflow(client, run.info.run_id) 
giskard_dataset.to_mlflow(client, run.info.run_id) 
scan_results.to_mlflow(client, run.info.run_id) 
test_suite_results.to_mlflow(client, run.info.run_id) 
```

## notebook to test this feature:
Run them locally in order to run the `mlflow ui`:
- titanic demo (sklearn): https://colab.research.google.com/drive/1cnYTOmFQnQXRGTlYif3MMYe7aOkCwC5d?usp=sharing

</details>

<details> <summary> <b> Open questions </b> </summary> 

- what if the model logged into MLflow is not giskard-compliant? 
- how does MLflow handle data preprocessing?
  -  MLflow allows the custom creation of models if pre- or post-processing is needed, see: https://mlflow.org/docs/latest/python_api/mlflow.pyfunc.html?highlight=preprocessing
</details>

### Tech
- [x] better html rendering -- fixed in https://github.com/Giskard-AI/giskard/pull/1209
- [x] better metrics logging
- [x] <s> use native mlflow saving when possible -- implemented inhttps://github.com/Giskard-AI/giskard/pull/1189/commits/e83667df6674c229bb922c797f338b2a5d2b4bd3 </s>
  - [x] remove `save_model` from `MLflowBasedModel` and introduce a `dataclass` class to create flags and disable the saving/loading validation during the scan. (PyFunc models are not meant to be saved, and the model going through `mlflow.evaluate` is a PyFunc model) 
- [x] implement the above notebooks as functional-tests
- [x] `tempfile.NamedTemporaryFile` doesn't seem to delete files after context
- [x] add text as `model_type` 
- [x] ability to compare two models (via `scan_summary.json`)
- [x] log scan results as json file to be rendered in artifact view of mlflow and allow comparison
- [x] implement `to_mlflow()` on (see [doc](https://mlflow.org/docs/latest/python_api/mlflow.html)):
  - [x] `ScanResult`
  - [x] `TestSuiteResult`
  - [x] `giskard.Model`
  - [x] `giskard.Dataset`
- [x] artifact view similar to [doc](https://www.databricks.com/blog/announcing-mlflow-24-llmops-tools-robust-model-evaluation)
- [x] Allow people to add any `Model.__init__()` argument to `evaluator_config` not only `classification_labels`
- [x] Better error handling + their unit-tests 
- [x] run `pdm lock -G:all`
- [x] Implement `EvaluationResult`
- [x] implement telemetry tracking
- [x] Implement an output message pushing for `mlflow ui`

### Writing

<details> <summary> <b> LLM models comparison </summary>

```py
import pandas as pd
from langchain import PromptTemplate, LLMChain
from langchain.llms import OpenAI
import mlflow
import openai
import os

df = pd.read_csv('https://raw.githubusercontent.com/sunnysai12345/News_Summary/master/news_summary_more.csv')
df_filtered = pd.DataFrame(df["text"].sample(10, random_state=11))

prompt = PromptTemplate(template="Create a reader comment according to the following article summary: '{text}''",
                        input_variables=["text"])

openai.api_key = os.getenv("OPENAI_API_KEY")

llm1 = OpenAI(openai_api_key=openai.api_key,
             request_timeout=20,
             max_retries=100,
             temperature=0,
             model_name="text-ada-001", ) # Possibility to select another model

chain1 = LLMChain(prompt=prompt, llm=llm1)

with mlflow.start_run(run_name="text-ada-001") as run2:
    model_uri = mlflow.langchain.log_model(chain1, "langchain").model_uri
    mlflow.evaluate(model=model_uri, model_type="text", data=df_filtered, evaluators="giskard")

llm2 = OpenAI(openai_api_key=openai.api_key,
              request_timeout=20,
              max_retries=100,
              temperature=0,
              model_name="text-embedding-ada-002", ) # Possibility to select another model

chain2 = LLMChain(prompt=prompt, llm=llm2)

with mlflow.start_run(run_name="text-embedding-ada-002") as run1:
    model_uri = mlflow.langchain.log_model(chain2, "langchain").model_uri
    mlflow.evaluate(model=model_uri, model_type="text", data=df_filtered, evaluators="giskard")
```

Screenshot 2023-07-10 at 15 01 19

</details> 

- [ ] doc
- [ ] Write an article

### Optional
- [ ] Populate MLflow's evaluation [examples](https://github.com/mlflow/mlflow/tree/master/examples/evaluation) with the giskard evaluator.
- [ ] Capture `giskard.scanner.logger ERROR    Detector LLMToxicityDetector failed with error: 'PyFuncModel' object has no attribute 'rewrite_prompt'` and output instead something like `giskard.scanner.logger ERROR    Detector LLMToxicityDetector is not supported by giskard plug-in through mlflow.evaluate`


rabah-khalek:
@andreybavt I am currently wrapping the `PyFunc` model I have access to in `mlflow.evaluate` with a custom giskard `CloudPickleBasedModel`. I only needed to customise `model_predict`, which was the advantage, since `PyFunc` standardise `predict()`.

The drawback of not having access to the underlying model object, is that we can't run all scan detectors. This mainly happened only once, for the toxicity detector where we needed the underling langchain model object to regenerate prompts, see: https://github.com/Giskard-AI/giskard/blob/116e1fd4f652cedcc0497a54eb7a78ecba089c89/python-client/giskard/models/langchain/__init__.py#L51-L71.

We can put a pin in this, but I think we might be forced to re-think the unwrapping of the model-object from the artifact logged by MLflow for our own purposes (scan).

python-client/giskard/datasets/base/__init__.py
```diff
@@ -615,6 +615,15 @@ def copy(self):
 
         return dataset
 
+    def to_mlflow(self, client=None, run_id=None):
+        import mlflow
+        with tempfile.NamedTemporaryFile(prefix="dataset-", suffix=".csv") as f:
+            self.df.to_csv(f.name)
```
andreybavt, line 621: why not use `giskard.client.io_utils.save_df` which handles None/empty string differentiation ?

python-client/giskard/models/base/__init__.py
```diff
@@ -765,6 +768,18 @@ def load_model(cls, local_dir):
...
+                  artifact_path="prediction-function-from-giskard",
+                  **kwargs):
+        import mlflow
+
+        class MLflowModel(mlflow.pyfunc.PythonModel):
+            def predict(self, df):
+                return self.predict(df)
+
+        mlflow_model = MLflowModel()
+        return mlflow.pyfunc.log_model(artifact_path=artifact_path, python_model=mlflow_model)
```
andreybavt, line 781: should we also log pre and post processing functions here? With `log_artefact`? Otherwise GIskard model will only be partially logged

python-client/giskard/models/sklearn/__init__.py
```diff
@@ -95,3 +95,14 @@ def model_predict(self, df):
             return self.model.predict(df)
         else:
             return self.model.predict_proba(df)
+
+    def to_mlflow(self,
+                  artifact_path="sklearn-model-from-giskard",
+                  **kwargs):
+        if self.is_classification:
```
andreybavt, line 102: let's reuse the code in `save_model` instead of duplicating it

python-client/giskard/scanner/result.py
```diff
@@ -123,3 +126,22 @@ def _track_suite(self, suite, name):
             "scan:generate_test_suite",
             {"suite_name": anonymize(name), "tests_cnt": len(suite.tests), **tests_cnt},
         )
+
+    def to_mlflow(self, client=None, run_id=None, summary=True, model_artifact_path=""):
```
andreybavt, line 130: I'd rename `client` to `mlflow_client` and add types to the arguments to avoid confution with GiskardClient

python-client/giskard/integrations/mlflow/giskard_evaluator.py
```diff
@@ -0,0 +1,101 @@
...
+                             "repository for further assistance: https://github.com/Giskard-AI/giskard.") from e
+
+        try:
+            test_suite = scan_results.generate_test_suite("scan test suite")
+            test_suite_results = test_suite.run()
+
+            # log metrics resulting from scan
+            test_suite_results.to_mlflow(client=self.client, run_id=self.run_id)
+        except Exception:  # noqa
+            pass
```
andreybavt, line 101: we should at least send a telemetry report in this case. Also why not log this error?

andreybavt (changes requested):

python-client/giskard/models/base/__init__.py
rabah-khalek: MLflow allows the custom creation of models if pre- or post-processing is needed, see: https://mlflow.org/docs/latest/python_api/mlflow.pyfunc.html?highlight=preprocessing. This means that if there's pre/post processing, it'll be already in the `predict()` of their pyfunc model, thus no need to log any additional functions from our side.

python-client/giskard/integrations/mlflow/giskard_evaluator.py
rabah-khalek: Even if the test suite wasn't generated, the user can still the results of the scan and the other artifacts. I will issue a warning instead. And yes I added telemetry.

rabah-khalek: the telemetry is minimal, since information is already collected in the scan. But I picked up the dataset and model ids under the mlflow-integration so we can identify when people use giskard through the mlflow plugin. Is that enough?

How can we make it simple from mixpanel to separate people that go through scan from mlflow-plugin and not explicitely? if there's more things to log to make this happen, let me know

rabah-khalek: in fact due to https://linear.app/giskard/issue/GSK-1328/symmetrise-dataset-and-model-upload-process-and-id-creation, i had to save the model and inner model names instead of ids, since I don't validate the saving of the model during scan, the model id is not created. We should fix this.

python-client/giskard/models/huggingface/__init__.py
```diff
@@ -144,3 +144,7 @@ def _get_predictions(self, data):
             return [[p[label] for label in self.meta.classification_labels] for p in _predictions]
 
         return self.model(**data)
+
+    def to_mlflow(self, artifact_path="transformers-model-from-giskard", **kwargs):
+        import mlflow
```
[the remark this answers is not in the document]
andreybavt: hmm, `HuggingFaceModel` inherits from `WrapperModel` which has a global `mlflow` import so I don't see how mlflow won't be imported by the time you reach line 149 here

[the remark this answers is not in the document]
rabah-khalek: You are describing a cpp-like include system where declarations are just placed in order in a single translation unit. That doesn't apply to python. Importing the same module everywhere is necessary. If you use inherit from class A to define a new class B you need to import again the libraries needed in B even if module A already imports them. Anyways, once a module is loaded by python interpreter remaining imports are not very expensive.

andreybavt:
> @andreybavt I am currently wrapping the `PyFunc` model I have access to in `mlflow.evaluate` with a custom giskard `CloudPickleBasedModel`. I only needed to customise `model_predict`, which was the advantage, since `PyFunc` standardise `predict()`.
> 
> The drawback of not having access to the underlying model object, is that we can't run all scan detectors. This mainly happened only once, for the toxicity detector where we needed the underling langchain model object to regenerate prompts, see:
> 
> https://github.com/Giskard-AI/giskard/blob/116e1fd4f652cedcc0497a54eb7a78ecba089c89/python-client/giskard/models/langchain/__init__.py#L51-L71
> 
> We can put a pin in this, but I think we might be forced to re-think the unwrapping of the model-object from the artifact logged by MLflow for our own purposes (scan).

I think we can look at `model_type` argument (or the value of `model_type` in model_config) to either wrap it with just a `PyFuncModel` or a `LangchainModel` and then be able to call `rewrite_prompt` in case it's a `text_generation` model

rabah-khalek:
> > @andreybavt I am currently wrapping the `PyFunc` model I have access to in `mlflow.evaluate` with a custom giskard `CloudPickleBasedModel`. I only needed to customise `model_predict`, which was the advantage, since `PyFunc` standardise `predict()`.
> > The drawback of not having access to the underlying model object, is that we can't run all scan detectors. This mainly happened only once, for the toxicity detector where we needed the underling langchain model object to regenerate prompts, see:
> > https://github.com/Giskard-AI/giskard/blob/116e1fd4f652cedcc0497a54eb7a78ecba089c89/python-client/giskard/models/langchain/__init__.py#L51-L71
> > 
> > We can put a pin in this, but I think we might be forced to re-think the unwrapping of the model-object from the artifact logged by MLflow for our own purposes (scan).
> 
> I think we can look at `model_type` argument (or the value of `model_type` in model_config) to either wrap it with just a `PyFuncModel` or a `LangchainModel` and then be able to call `rewrite_prompt` in case it's a `text_generation` model

As it's not a blocker, I think it's a good idea to merge this branch and to take care of this point in a new one. This will unlock some marketing actions. WDYT?

PR #1196, opened 2023-06-22
GSK-1335 & GSK-1264 & GSK-1255 - Test re-execution result not clear

kevinmessiaen:
## Description

- [x] Fixed issue where it's not clear whenever the test are rerun in the catalog
- [x] Added loading spinner
- [x] Added possibility to run test on whole dataset 

## Type of Change

- [ ] 🔐 Security fix


rabah-khalek (approved):
I haven't checked the code, but I have tested it, and it does exactly what it says on the tin.

frontend/src/api.ts
```diff
@@ -457,18 +457,18 @@ export const api = {
...
   },
   async getCatalog(projectId: number) {
     return apiV2.get<unknown, CatalogDTO>(`/catalog`, {
       params: {
-        projectId,
-      },
+        projectId
+      }
     });
   },
```
henchaves, line 473: Why don't we use `projectId` as a path param?

kevinmessiaen: It's true that it also can be a path variable in this case. I think it was set as a filter to show the data associated to the project but now that all catalog items will be linked to projects it make more sense to use a path variable

henchaves (approved):

PR #1198, opened 2023-06-23
Push Feature - Merged

Googleton:


Googleton:
For some reason my formatter still changed a ton of white spaces. Any idea @andreybavt ?

python-client/giskard/ml_worker/server/ml_worker_service.py
```diff
@@ -492,6 +493,98 @@ def filter_wrapper(row):
...
+
+        if overconf is not None:
+            overconf_grpc = overconf.to_grpc()
+        else:
+            overconf_grpc = None
+
+        if borderl is not None:
+            borderl_grpc = borderl.to_grpc()
+        else:
+            borderl_grpc = None
```
kevinmessiaen, line 544: I have feeling that this code is a bit repetitive. To me it would be better if we store those push in a list/dict to better handle them in a loop:

```
pushes = {
   PushKind.Perturbation: create_perturbation_push(model, dataset, request.rowidx),
   ...
}

pushes_grpc = { push_kind: push.to_grpc() if push is not None else None for push_kind, push for pushes.items()
```

python-client/giskard/ml_worker/server/ml_worker_service.py
```diff
@@ -492,6 +493,98 @@ def filter_wrapper(row):
...
+
+        if request.cta_kind > 0 and request.push_kind > 0:
+            if request.push_kind == PushKind.Perturbation:
```
kevinmessiaen, line 547: Plus this would become: `push = pushes[request.push_kind]`

python-client/giskard/push/contribution.py
```diff
@@ -2,15 +2,20 @@
...
+def create_contribution_push(model, ds, df):
     if model.meta.model_type == SupportedModelTypes.CLASSIFICATION and _existing_shap_values(ds):
-        shap_res = _contribution_push(model, ds, idrow, df)
-        slice_df = ds.slice(lambda df: df.loc[[idrow]], row_level=False)
+        shap_res = _contribution_push(model, ds, df)
+        slice_df = Dataset(df=df, target=ds.target, column_types=ds.column_types.copy(), validation=False)
+        # slice_df = ds.copy()
+        # slice_df.df = df.reset_index(drop=True)
+        # slice_df.meta.number_of_rows = 1
+        # slice_df = ds.slice(lambda row: row.eq(df.iloc[0]).all(axis=None)) 3 solutions I think the first is the best
```
kevinmessiaen, line 18: Yes I also think that the first looks cleaner

python-client/giskard/push/contribution.py
```diff
@@ -39,8 +44,8 @@ def create_contribution_push(model, ds, idrow, df):
                     return res
 
     if model.meta.model_type == SupportedModelTypes.REGRESSION and _existing_shap_values(ds):
-        shap_res = _contribution_push(model, ds, idrow)
-        slice_df = ds.slice(lambda df: df.loc[[idrow]], row_level=False)
+        shap_res = _contribution_push(model, ds, df)
+        slice_df = ds.slice(lambda row: row.equals(df.iloc[0]))
```
kevinmessiaen, line 48: I'm not sure why we need to do a row by row comparison like that, it seems suboptimal.

backend/src/main/java/ai/giskard/service/PushService.java
```diff
@@ -0,0 +1,36 @@
...
+import ai.giskard.ml.MLWorkerClient;
+import ai.giskard.service.ml.MLWorkerService;
+import ai.giskard.worker.*;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+import java.util.Map;
+
+@Service
+//@Transactional
```
andreybavt, line 12: It's rightfully not transactional since it's communicating with the ML Worker inside, you can remove the comment

frontend/src/api.ts
```diff
@@ -162,335 +163,350 @@ axios.interceptors.request.use(jwtRequestInterceptor);
...
+        });
+    },
+    async datasetProcessing(projectId: number, datasetUuid: string, functions: Array<ParameterizedCallableDTO>, sample: boolean = true) {
+        return apiV2.post<unknown, DatasetProcessingResultDTO>(
+            `/project/${projectId}/datasets/${encodeURIComponent(datasetUuid)}/process?sample=${sample}`,
+            functions
+        );
+    },
+    async getPushes(modelId: string, datasetId: string, idx: number, features: any) {
+        return apiV2.post<unknown, unknown>(`/pushes/${modelId}/${datasetId}/${idx}`, {
```
andreybavt, line 498: Why do you use Post request to get elements?

Why not specifying the type of the return object?

frontend/src/stores/debugging-sessions.ts
```diff
@@ -1,51 +1,72 @@
...
-        await this.loadDebuggingSessions(this.projectId);
-      }
+    state: (): State => ({
+        projectId: null,
+        debuggingSessions: [],
+        currentDebuggingSessionId: null,
+        selectedSlicingFunction: {
+            uuid: undefined,
+            params: [],
+            type: 'SLICING'
```
andreybavt, line 20: 'SLICING' should probably be an enum

frontend/src/stores/push.ts
```diff
@@ -0,0 +1,60 @@
...
+        current: undefined,
+        identifier: undefined
+    }),
+    getters: {},
+    actions: {
+        async fetchPushSuggestions(modelId: string, datasetId: string, rowNb: number, inputData: any, modelFeatures: string[]) {
+            this.identifier = {modelId, datasetId, rowNb, inputData, modelFeatures};
+            this.current = undefined;
+
+            // if (!this.pushes[modelId]) {
```
andreybavt, line 38: cleanup

frontend/src/components/PushPopover.vue
```diff
@@ -0,0 +1,170 @@
+<template>
+  <v-menu offset-x bottom :close-on-content-click="false" v-model="opened" v-if="show">
+    <template v-slot:activator="{ on: onMenu }">
+      <v-tooltip right>
+        <template v-slot:activator='{ on: onTooltip }'>
+          <v-btn small icon outlined color="warning" class="ml-1"
+                 v-on="{ ...onMenu, ...onTooltip }">
+            <v-icon size="18" color="warning">mdi-alert-outline</v-icon>
+            <!--<v-badge top overlap color="transparent">
```
andreybavt, line 9: cleanup

andreybavt (changes requested):
Check Sonar's issues and please fix them too

backend/src/main/java/ai/giskard/web/dto/ApplyPushDTO.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Getter
+@Setter
+@UIModel
+@NoArgsConstructor
+public class ApplyPushDTO {
+    private UUID modelId;
+    private UUID datasetId;
+    private int rowIdx;
+    private PushKind pushKind;
```
kevinmessiaen, line 21: Since PushKind and CallToActionKind is a MLWorker generated type we have `any` type in ts due to `missing translation`.  @andreybavt do you think there is a quick fix or we can keep like this.

frontend/src/components/PushPopover.vue
```diff
@@ -0,0 +1,242 @@
...
+    case "contribution":
+      return value.value?.contribution;
+    case "perturbation":
+      return value.value?.perturbation;
+    case "overconfidence":
+      return value.value?.overconfidence;
+    case "borderline":
+      return value.value?.borderline;
+    default:
+      return false;
```
kevinmessiaen, line 118: Actually this can be simplified by
```suggestion
  return value.value?.hasOwnProperty(props.type) ?  value.value[props.type] : false;
```

frontend/src/components/PushPopover.vue
```diff
@@ -0,0 +1,242 @@
...
+  let uuid = await pushStore.applyPush(push.value!.kind, kind);
+
+
+  switch (kind) {
+    case "CreateSlice":
+    case "SaveExample":
+      mainStore.addSimpleNotification("Successfully saved");
+      break;
+    case "CreateSliceOpenDebugger":
+      await catalogStore.loadCatalog(projectStore.currentProjectId ?? 0);
```
kevinmessiaen, line 150: I think it'll be clearer to break down each cases into method and just calling them.

Furthermore I also think `kind` should be an enum instead of a string

frontend/src/components/PushPopover.vue
kevinmessiaen: Actually I think it'll make more sense to answer `undefined` instead of `false` when no value found.

frontend/src/components/PushPopover.vue
```diff
@@ -0,0 +1,242 @@
...
+      return "mdi-approximately-equal";
+    default:
+      return "mdi-bug";
+  }
+});
+
+async function applyCta(kind: string) {
+  loading.value = kind;
+  // modelId: string, datasetId: string, rowNb: number, pushKind: string, ctaKind: string
+  // @ts-ignore
```
kevinmessiaen, line 140: Shouldn't be necessary anymore if `push` computed method reply undefined instead of false

frontend/src/views/main/project/Inspector.vue
```diff
@@ -211,15 +267,15 @@ function isFeatureEditable(featureName: string) {
...
-async function onValuePerturbation(featureMeta) {
-  mixpanel.track("Feature perturbation", {
-    columnType: featureMeta.type,
-    featureName: anonymize(featureMeta.name),
-    modelId: props.model.id,
-    datasetId: props.dataset.id
-  })
-  emit('update:inputData', props.inputData)
-}
+// async functiononValuePerturbation(featureMeta) {
```
kevinmessiaen, line 270: We can delete this method

frontend/src/views/main/project/InspectorWrapper.vue
```diff
@@ -434,6 +455,26 @@ async function handleSwitchSample() {
     }
   });
 }
+
+const modelFeatures = computed(() => {
```
kevinmessiaen, line 459: I think this can be removed totally since it just return a copy of `inputMetaData`

PR #1201, opened 2023-06-23
[GSK-1189] Fix scan widget issues on narrow window size

mattbit:


mattbit:
Converting to draft since there are doubts about the solution

mattbit:
I close this in favour of #1207.

PR #1207, opened 2023-06-26
[GSK-1346, GSK-1347, GSK-1189] Update scan widget

mattbit:
- Added links to documentation for various issues
- Fixed layout problems with the tab headers when the window is narrow
- Updated code snippet and CTA
- Update spurious correlation description
- EXTRA: Fixed formatting in the docs


mattbit:
Example here: https://colab.research.google.com/drive/1PXm1ERGA9Ce9B_hIWBtORo1jotUo1z9d?usp=sharing

rabah-khalek (approved):

rabah-khalek:
I will merge it, but we need to fix the sonarcloud checks

mattbit:
> I will merge it, but we need to fix the sonarcloud checks

I checked the sonarcloud, those are false positives we can safely ignore.

PR #1209, opened 2023-06-27
Make scan widget optionally embeddable

mattbit:
- made the scan result `to_html` produce a complete HTML page
- optional `embed=True` produces the iframe to embed in notebook
- fixed problem in the j2 templates that prevented the correct layout to be used


rabah-khalek (approved):
works both in notebook and static html. Thanks!

PR #1210, opened 2023-06-27
[GSK-1373] Configure nightly Giskard build that runs all tests

Inokinoki:
## Description

Add a Gradle task and a workflow to run full Python tests when nobody works(hopefully).

The `schedule` trigger is set to be run at 1:00 a.m. everyday, UTC(https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule). The corresponding time is 2:00 or 3:00 a.m. CEST.

## Type of Change

- [ ] 🔐 Security fix

## Checklist


python-client/build.gradle.kts
```diff
@@ -80,6 +80,12 @@ tasks {
         command = "run test-fast"
     }
 
+    create<PythonTask>("full-python-test") {
```
andreybavt, line 83: I'd rename it to `test-all` to make it "python" independent in case we do the same separation for backend tests and in this case `gradle test-all` would trigger all tests in both: java and python

.github/workflows/nightly-python-test.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Get number of available cores for Python test workers
+        id: nb_workers
+        run: python -c 'import os; print(f"nb_workers={max(1,os.cpu_count()-1)}")' >> $GITHUB_OUTPUT
+
+      - name: Run Python commands
+        run: |
+          python -m pip install virtualenv -U --upgrade pip
+          echo "Python ($(which python)) packages:\n $(python -m pip list)"
+
+      - name: Run full Python tests
```
andreybavt, line 56: I'd actually do a full build (including all tests execution), not just python tests. Like this we'll know each morning that we've tested everything we could test

andreybavt (changes requested):

Inokinoki:
Currently can run without issue locally on an ARM64 Linux worker, through act.

The scheduled task only take effect on the default branch.

PR #1234, opened 2023-07-06
[GSK-1343] Better support for CUDA and models + docs improvements

mattbit:
- Added support for batching in `WrapperModel`
- Fixed a bug in `CatboostModel` that prevented `catboost` models from being loaded correctly
- Reorganized the `models` package and revised all docstrings
- Added tests for `CatboostModel`, `WrapperModel`, `BaseModel`, `CloudpickleSerializableModel`
- Re-enabled skipped tests that now can work! (`test_catboost_changed_column_order` and `test_tabular_titanic_binary_classification`)
- Added all model wrappers to the docs, under “API Reference”

### TODO
- [x] write test for batching
- [x] manual test with pytorch on CUDA
- [x] manual test with tensorflow on CUDA


mattbit:
@andreybavt few questions about things I encountered while revisiting the models code:

1. We have not tests for the `BaseModel::download` method. Do you think there is an easy way to add them, in a similar fashion to what was done for the `upload` (with a mocked client)?
2. We define a `load` method in `BaseModel` which, from what I understand, is never used. This method uses `cloudpickle` but I think this was moved to the specific cloudpickle model class (which concrete implementations like `PredictionFunctionModel` inherit from). Is this just a left over from previous releases? In that case I would remove it.

andreybavt:
@mattbit 

1. Sure, covering `download` with tests is a good idea
2. In case you want to create a custom model you'd inherit a `CustomModel` class in this case you'd use a load from BaseModel

python-client/docs/reference/models/integrations/base_classes.rst
```diff
@@ -0,0 +1,5 @@
+Base model classes
+==================
+
+.. automodule:: giskard.models.base
+    :members: BaseModel, WrapperModel, MLFlowSerializableModel, CloudpickleSerializableModel
```
rabah-khalek, line 5: Could we rewrite this in a way that shows these 4 classes as a drop-down list when we click on "Base model classes" on the left toc menu?

python-client/giskard/models/base/wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @configured_validate_arguments
+    def predict_df(self, df: pd.DataFrame):
+        if self.batch_size and self.batch_size > 0:
+            dfs = np.array_split(df, np.arange(self.batch_size, len(df), self.batch_size))
+        else:
+            dfs = [df]
+
+        outputs = []
+        for batch in map(self._preprocess, dfs):
```
rabah-khalek, line 116: Ok i like this general solution

python-client/giskard/models/base/wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+            A list of classification labels. Default is None.
+        batch_size : Optional[int], optional
+            The batch size to use for inference. Default is ``None``, which
+            means inference will be done on the full dataframe.
+        """
+        super().__init__(model_type, name, feature_names, classification_threshold, classification_labels, **kwargs)
+        self.model = model
+        self.data_preprocessing_function = data_preprocessing_function
+        self.model_postprocessing_function = model_postprocessing_function
+        self.batch_size = batch_size
```
rabah-khalek, line 73: If you think `batch_size` is general enough, I'd suggest adding it to `ModelMeta` in `BaseModel`: https://github.com/Giskard-AI/giskard/blob/28e2ca2e42c3354f7508717d7da23d90c6068091/python-client/giskard/models/base/model.py#L126-L134

This is currently not saved anywhere, anytime you add new attributes to a daughter class, you'd need to save it so that, upon loading (which you also need to override), you feed it to the constructor, see what we did in this class for `data_preprocessing_function` and `model_postprocessing_function`: https://github.com/Giskard-AI/giskard/blob/28e2ca2e42c3354f7508717d7da23d90c6068091/python-client/giskard/models/base/wrapper.py#L217-L226

or also what we did here: https://github.com/Giskard-AI/giskard/blob/28e2ca2e42c3354f7508717d7da23d90c6068091/python-client/giskard/models/pytorch.py#L200-L229

rabah-khalek (changes requested):
- the `batch_size` to be saved and reloaded properly in `WrapperModel`
- good catch the catboost missing `load_model`, I remember vividly to have fixed it, but maybe in a branch that hasn't been ever merged.
- good job on the tests
- good job on the doc
- Overall I like the new refactoring of models, the previous modular structure we followed only makes sense if we didn't have classes, only collection of methods (such as mlflow)

rabah-khalek:
A not-so-urgent suggestion: to avoid devs from forgetting the saving of daughter model class' meta such as `batch_size` in your case, would be wise to extend `model_validation` in https://github.com/Giskard-AI/giskard/blob/c860a43a7ba1b7a5c6e11ddeafe7047d9f82409a/python-client/giskard/core/model_validation.py#L126 to compare the attributes values (as much as possible) before and after the saving, with an associated test

python-client/giskard/models/base/wrapper.py
mattbit: I’m not sure I want to save the `batch_size`, because it depends on hardware (and potentially dataset). So when you run it on the ml worker situation might be different.

Actually the same would hold for the torch device…

mattbit:
> A not-so-urgent suggestion: to avoid devs from forgetting the saving of daughter model class' meta such as `batch_size` in your case, would be wise to extend `model_validation` in
> 
> https://github.com/Giskard-AI/giskard/blob/c860a43a7ba1b7a5c6e11ddeafe7047d9f82409a/python-client/giskard/core/model_validation.py#L126
> to compare the attributes values (as much as possible) before and after the saving, with an associated test

The problem with that would be deciding which attributes should be checked. I’m not sure that all model attributes would need to be persisted and reloaded.

python-client/docs/reference/models/integrations/base_classes.rst
mattbit: You mean the sidebar?

python-client/giskard/models/base/wrapper.py
mattbit: @andreybavt what is your take on this? Should we save this kind of attributes like `device` or `batch_size` which depend on the hardware performing the inference?

My opinion is that they are bound to the local environment. If we envision a use case where models can be downloaded and executed by workers which are not necessarily on the same hardware, this could lead to failures.

Example:
I trained and uploaded my pytorch model on my GPU machine, I set `device="cuda"` and large `batch_size`.
Now I run my worker on a smaller machine which has less GPU memory. When I try to run `model.predict`, I get an out of memory error.
Similar case: I start my worker on my laptop which has no GPU, as soon as it tries to load the model from the giskard server it fails because `device="cuda"` is not a valid device.

Whether this kind of situation would occur depends on the expected usage of the worker.

andreybavt: I agree that saving and loading `device` and `batch_size` can be risky when running a worker on a different machine. Our load method already allows passing `kwargs` so these parameters can be specified at the load time.
I think we should either not save them at all or save and load (like we already do for pytorch) but prefer the values from `kwargs` if they're provided.

rabah-khalek: I agree that `batch_size` (post-inference) and `device` are not model properties, but the moment a user wraps their model with a specific `batch_size` and `device`, they (I think) expect the default outcome of this, is that these are the parameters that will be user throughout giskard -- either in the scan (pre-saving) or on the server (post-loading).

So I prefer 

> save and load (like we already do for pytorch) but prefer the values from kwargs if they're provided 

I find it a bit strange that a user wraps his model with `device="cuda"` but suddenly when they upload it to the server and run it, it would run on the default `device="cpu"` (the download/load on the backend happens implicitly, so a user can't specify the `device` when he runs the model on the server after upload).

python-client/docs/reference/models/integrations/base_classes.rst
rabah-khalek: yes, the left sidebar - not really urgent this one. I admittedly tried to do it myself, wasn't super intuitive without some files re-organization.

rabah-khalek (approved):
Thanks for the update @mattbit, I just did some minor refactoring in a70b89aaf7cef9d34a1d0b11f1e9a9215ad0d9bf. To me, it's good to go.

python-client/giskard/models/base/wrapper.py
```diff
@@ -0,0 +1,291 @@
...
+                },
+                f,
+                default_flow_style=False,
+            )
+
+    @classmethod
+    def load(cls, local_dir, **kwargs):
+        kwargs["data_preprocessing_function"] = cls.load_data_preprocessing_function(local_dir)
+        kwargs["model_postprocessing_function"] = cls.load_model_postprocessing_function(local_dir)
+        kwargs.update(cls.load_wrapper_meta(local_dir))
```
andreybavt, line 241: It's going to do the opposite of what we discussed:

`save and load (like we already do for pytorch) but prefer the values from kwargs if they're provided`

meaning that you won't be able to override the value of batch_size from the kwargs

andreybavt (changes requested):

python-client/giskard/models/base/wrapper.py
mattbit: That’s true, I will add a test

andreybavt (approved):

PR #1235, opened 2023-07-06
Split train and test set in fraud detection fixtures

mattbit:
Should fix the `test-slow` build.


Inokinoki:
Thanks, I manually triggered the full test on this branch, seems fixed: https://github.com/Giskard-AI/giskard/actions/runs/5480193112

PR #1236, opened 2023-07-06
[GSK-1365] Make websocket reactive to worker status changes

henchaves:
## Description

This PR aims to migrate the polling system of  the ML Worker websocket to only send the current worker status message to the channel if one of the following conditions happens:

 - worker connected
 - worker disconnected
 - new client subscriber to the channel
 
## Related Issue

[GSK-1365 (available on Linear)](https://linear.app/giskard/issue/GSK-1365/make-websocket-reactive-to-worker-status)

## Type of Change

- [ ] 🔐 Security fix

### DIFF
--- a/backend/src/main/java/ai/giskard/web/socket/WorkerStatusSocketService.java
+++ b/backend/src/main/java/ai/giskard/web/socket/WorkerStatusSocketService.java
@@ -1,17 +1,14 @@
 package ai.giskard.web.socket;
 
+import ai.giskard.event.UpdateWorkerStatusEvent;
 import ai.giskard.service.ml.MLWorkerService;
 import lombok.RequiredArgsConstructor;
 import org.springframework.context.event.EventListener;
 import org.springframework.messaging.simp.SimpMessagingTemplate;
 import org.springframework.stereotype.Service;
 import org.springframework.web.socket.messaging.SessionSubscribeEvent;
-
-import javax.annotation.PostConstruct;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Timer;
-import java.util.TimerTask;
 
 @Service
 @RequiredArgsConstructor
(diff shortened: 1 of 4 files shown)


backend/src/main/java/ai/giskard/web/socket/WorkerStatusSocketService.java
```diff
@@ -18,19 +17,22 @@ public class WorkerStatusSocketService {
...
+    }
+
+    @EventListener
+    public void handleUpdateWorkerStatusEvent(UpdateWorkerStatusEvent event) {
+        sendCurrentStatus();
+    }
+
+    public void sendCurrentStatus() {
+        Map<String, Boolean> data = new HashMap<>();
+        data.put("connected", mlWorkerService.isExternalWorkerConnected());
```
kevinmessiaen, line 32: I think that we should take more advantage of the `UpdateWorkerStatusEvent`.

For me the event should carry a variable `boolean connected` and  this variable should be set as a global variable in `WorkerStatusSocketService` by default at `false`. This way we won't need to call `isExternalWorkerConnected` that will then call the MLWorker since we already have the status.

Actually `isExternalWorkerConnected` would ideally check for the latest status event

kevinmessiaen (changes requested):

backend/src/main/java/ai/giskard/web/socket/WorkerStatusSocketService.java
henchaves: I'm not sure about changing `isExternalWorkerConnected`. I think it's important to keep this method forcing to check if there is a worker connected at the moment it's called instead of replicating the status of the last emitted `UpdateWorkerStatusEvent`.

backend/src/main/java/ai/giskard/web/socket/WorkerStatusSocketService.java
```diff
@@ -18,19 +17,35 @@ public class WorkerStatusSocketService {
...
+        data.put("connected", true);
+        sendData(data);
+    }
+
+    @EventListener(condition = "#event.isConnected == false")
+    public void handleWorkerDisconnectedEvent(UpdateWorkerStatusEvent event) {
+        Map<String, Boolean> data = new HashMap<>();
+        data.put("connected", false);
+        sendData(data);
+    }
```
kevinmessiaen, line 37: We can simply get the status from the event and one listener to forward it (we will need to add a getter in the event)

```suggestion
    @EventListener()
    public void handleWorkerStatusChangeEvent(UpdateWorkerStatusEvent event) {
        Map<String, Boolean> data = new HashMap<>();
        data.put("connected", event.isConnected());
        sendData(data);
    }
```

kevinmessiaen (approved):

PR #1242, opened 2023-07-12
[GSK-1408] Fix console errors on frontend

henchaves:
## Description
This PR aims to solve a lot of "console error" logs. The idea is to make everything working really well to migrate the frontend from Vue 2 to Vue 3 smoothly.
It modifies many files, so I apologise for the huge amount of code to review without being related to each other.

## Related Issue

[GSK-1408 (available on Linear)](https://linear.app/giskard/issue/GSK-1408/fix-console-errors-on-frontend)

## Type of Change

- [ ] 🔐 Security fix


frontend/src/components/AddDebuggingSessionModal.vue
```diff
@@ -19,22 +51,19 @@ interface Props {
...
 const isMLWorkerConnected = computed(() => {
   return state.workerStatus.connected;
 });
 
 const missingValues = computed(() => {
-  if (selectedDataset.value === null || selectedModel.value === null) {
+  if (selectedDatasetId.value === null || selectedModelId.value === null) {
     return true;
   }
   return false;
```
kevinmessiaen, line 69: Can be simplified to: `return selectedDatasetId.value === null || selectedModelId.value === null`

frontend/src/views/main/project/TestSuite.vue
```diff
@@ -102,6 +113,11 @@ const { suite, inputs, executions, hasTest, hasInput, statusFilter, searchFilter
 
 const openWorkerInstructions = ref(false);
 
+const suiteName = computed(() => {
+  if (!suite.value) return `Unnamed test suite`;
+  return suite.value.name ? suite.value.name : `Unnamed test suite`;
```
kevinmessiaen, line 118: In js you can use nullish coalescing instead of ternary in this case: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing 

It won't work if name is an empty string but logical or (`||`) would work

```suggestion
  return suite.value?.name ?? 'Unnamed test suite';
```

frontend/src/views/main/project/Debugger.vue
```diff
@@ -144,6 +144,7 @@ async function showPastSessions() {
     params: {
       projectId: props.projectId.toString()
     }
+  }).catch(() => {
```
kevinmessiaen, line 147: I'm not a big fan of catching error and silencing them. Apparently a solution to update query params on the same route is to push the params only, but we will need to check if the route is the same or not each time...

```
await router.push({
    params: {
      projectId: props.projectId.toString()
    }
  })
```

kevinmessiaen (approved):

PR #1243, opened 2023-07-13
Make it possible to completely delete users not just disable

kevinmessiaen:
## Description

This PR add the possibility to delete a disabled user from the administration panel. The delete icon is only showed for disabled users.
image

The following business rules are applied:
- If the deleted is the owner of projects, those projects and associated entities will be deleted from database and local storage (models, datasets, test suites and python pickled methods).
- The feedbacks generated by the user are kept and the owner will be shown ad "deleted user"   

## Related Issue

N/A

## Type of Change

- [ ] 🔐 Security fix

## Checklist


henchaves:
I get an error when I try to delete a user.

image
image

henchaves:
I can't see the "deleted user" after deleting a user. For example, I deleted the user "henrique" but I can still see its name on feedback.
image

henchaves:
I think we should display different icons for disable and delete actions. It was the first time I created / deleted a user, and it was a bit confusing. I didn't understand that after clicking on the bin icon this would me the user disabled instead of deleting it.

kevinmessiaen:
> I think we should display different icons for disable and delete actions. It was the first time I created / deleted a user, and it was a bit confusing. I didn't understand that after clicking on the bin icon this would make the user disabled instead of deleting it.

You're right, now that we have the possibility to delete users we should update the disable icon. Let me update it

henchaves (approved):
The "Restore User" doesn't work at the moment, but don't worry about that on this PR. There is a issue already created to deal with that https://linear.app/giskard/issue/GSK-1396/restore-user-doesnt-work

PR #1253, opened 2023-07-22
Fix transformation issue

mathieuroques:
Now numerical values are supported


mathieuroques:
I struggle to add drop-down selector to column_name and default value to mad_num
image

mathieuroques:
Not sure we need the save transformation CTA anymore as we can add it directly in the default config

mathieuroques:
@rabah-khalek It should be ok now

python-client/giskard/ml_worker/testing/functions/transformation.py
```diff
@@ -139,3 +140,17 @@ def change_writing_style(
...
     x.at[index, column_name] = chain_rewrite.run({"text": x.at[index, column_name], "style": style})
     return x
+
+
+@transformation_function(name="MAD Increment", tags=["num"], row_level=False)
+def mad_transformation(
+    data: pd.DataFrame, column_name: str, factor: float = 3, value_added: float = None
+) -> pd.DataFrame:
+    """
+    Add 3 times the MAD value to the column
```
rabah-khalek, line 150: This should rather be:
```py
"""
Add 3 times the value_added to the column, or if unavailable, add 3 times the MAD value.
"""
```

rabah-khalek (approved):
I just left you a small comment on the docstring of `mad_transformation`

python-client/giskard/push/utils.py
```diff
@@ -31,3 +31,10 @@ def bins_count(model, dataframe):  # done at the beggining
         flag = {value: count < ratio for value, count in counts.items()}
         value_counts[column] = {"value_counts": counts, "nunique": nunique, "ratio": ratio, "flag": flag}
     return value_counts
+
+
+def compute_mad(x):
```
rabah-khalek, line 36: I would also suggest to use the scipy stats method, it's fast and more transparent. The only down side could be the version compatibility, but as long as it works with the scipy that we ship, should be fine. 
Screenshot 2023-07-24 at 16 35 10

rabah-khalek (changes requested):

PR #1257, opened 2023-07-25
Feature/gsk 1405 make sure borderline threshold is consistent

mathieuroques:
Make sure that borderline threshold is the same on for the push and debug feature

- [x] Borderline: 

> - [x] Push
> - [x] Debug

----Test on:
> - [x] Binary classification
> - [x] Multi-class classification


python-client/giskard/push/prediction.py
```diff
@@ -33,26 +34,22 @@ def create_overconfidence_push(model, ds, df):
...
-    diff = max_val - second_max_val
-    # diff = abs(max_val - second_max_val)/second_max_val
-    # diff = abs(max_val - second_max_val)
-    return diff, max_val, second_max_val
+    row_slice = Dataset(df=df, target=ds.target, column_types=ds.column_types.copy(), validation=False)
+    prediction_results = model.predict(row_slice)
+    values = row_slice.df
+    target_value = values[ds.target].values.item()
+
+    if model.is_classification:
```
rabah-khalek, line 42: what about the regression case?

python-client/giskard/push/prediction.py
```diff
@@ -33,26 +34,22 @@ def create_overconfidence_push(model, ds, df):
...
+        target_value_proba = prediction_results.all_predictions[target_value].values.item()
+        if len(model.meta.classification_labels) > 2 or model.meta.classification_threshold is None:
```
rabah-khalek, line 44: `not model.is_binary_classification()` instead of `len(model.meta.classification_labels) > 2`.
`model.meta.classification_threshold is None` What justified this condition? by default it is `0.5` (`classification_threshold: Optional[float] = 0.5` in `BaseModel`)

python-client/giskard/push/prediction.py
```diff
@@ -33,26 +34,22 @@ def create_overconfidence_push(model, ds, df):
...
+            sorted_predictions = np.sort(prediction_results.raw[0])
+            abs_diff = sorted_predictions[-1] - sorted_predictions[-2]
```
rabah-khalek, line 46: This should be `abs_diff = sorted_predictions[-2] - sorted_predictions[-1]` since they're sorted otherwise it's negative.

what if `len(model.meta.classification_labels) == 3`, why not also check if `abs_diff = sorted_predictions[-3] - sorted_predictions[-2]` <= 0.1?

what if `len(model.meta.classification_labels) == N`, why not also check if `abs_diff = sorted_predictions[-N] - sorted_predictions[-N-1]` <= 0.1 for all Ns?

rabah-khalek (changes requested):

python-client/giskard/push/prediction.py
mathieuroques: I believe that 
```python
abs_diff = sorted_predictions[-2] - sorted_predictions[-1]
```  
is positive as the last value is higher than the penultimate one

mathieuroques: About the other N -1 cases, the goal is to identify the examples that can switch from one prediction to another easily, I'm not sure that we want to use the N-1 other features

rabah-khalek: > I believe that
> 
> ```python
> abs_diff = sorted_predictions[-2] - sorted_predictions[-1]
> ```
> 
> is positive as the last value is higher than the penultimate one

Ah you're right, the sorting is descending. Sorry.



> About the other N -1 cases, the goal is to identify the examples that can switch from one prediction to another easily, I'm not sure that we want to use the N-1 other features

Okay you're right.

rabah-khalek (approved):

PR #1261, opened 2023-07-26
[GSK-1477] Ensuring backward compatibility of models

rabah-khalek:
The saving and loading of `batch_size` should be backward compatible.


mattbit (approved):
Good catch @rabah-khalek! Thanks!

PR #1262, opened 2023-07-26
[GSK-1087] Avoid multiple requests for license/settings

henchaves:
## Description

This PR aims to fix a bug which request many times to the backend to /license and /settings endpoint.

## Related Issue

[GSK-1087 (available on Linear)](https://linear.app/giskard/issue/GSK-1087/a-lot-of-subsequent-settingslicense-requests)

## Type of Change

- [ ] 🔐 Security fix


frontend/src/stores/user.ts
```diff
@@ -1,167 +1,165 @@
...
+    async checkLoggedIn() {
+      const mainStore = useMainStore();
+      if (mainStore.authAvailable) {
+        if (!this.isLoggedIn) {
+          let token = this.token;
+          if (!token) {
+            const localToken = getLocalToken();
+            if (localToken) {
+              this.token = localToken;
+              token = localToken;
```
kevinmessiaen, line 72: I don't think that we really need a `token` variable since it's always equals to `this.token` 

It seems (need to verify) that we can simplify everything by just

```
this.token = this.token || getLocalToken()
```

frontend/src/stores/user.ts
```diff
@@ -1,167 +1,165 @@
...
-                throw new Error(error.response.data.detail);
-            }
+          }
+        }
+      } else if (!this.isLoggedIn) {
+        const response = await api.getUserAndAppSettings();
+        this.isLoggedIn = true;
+        this.userProfile = response.user;
+        let appConfig = response.app;
+        mainStore.setAppSettings(appConfig);
```
kevinmessiaen, line 90: This lines are duplicated with lines 76-80. I think the order of if/else it can be improved to make it more concise

```typescript
if (!this.isLoggedIn) {
  if (mainStore.authAvailable) {
    this.token = this.token || getLocalToken();
    if (!token) {
      this.removeLogin();
      return;
    }
  }
  const response = await api.getUserAndAppSettings();
  this.isLoggedIn = true;
  this.userProfile = response.user;
  mainStore.setAppSettings(response.app);
}
```

(ideally we should break down to sub function to avoid having too much layers of if)

kevinmessiaen (approved):
Looks good to me! Some small code refactoring possible but it's not directly related to the PR

frontend/src/stores/user.ts
henchaves: this doesn't work because `getLocalToken()` may output null

frontend/src/stores/user.ts
henchaves: I refactored this `checkLoggedIn` method.
Could you recheck it?

frontend/src/stores/user.ts
kevinmessiaen: In that case something like this should work: `this.token = this.token || getLocalToken() || ''`

frontend/src/stores/user.ts
kevinmessiaen: I saw that you forgot about setting `this.token`

kevinmessiaen (approved):

frontend/src/stores/user.ts
henchaves: Thanks!!

PR #1271, opened 2023-07-28
Feature/gsk 1493 rename borderline as underconfidence

mathieuroques:
Replacing borderline name by underconfidence


andreybavt:
@rabah-khalek @Googleton I see that the card is marked as done on linear but not yet merged.

rabah-khalek:
I moved the card back to-do, I can take care of the refactoring, but will do it in a new PR.

PR #1276, opened 2023-08-01
Fixed issue where `QueryBasedSliceFunction` wasn't working with bytes

kevinmessiaen:
## Description

Fixed issue where `QueryBasedSliceFunction` wasn't working with bytes

## Type of Change

- [ ] 🔐 Security fix

### DIFF
--- a/python-client/giskard/slicing/slice.py
+++ b/python-client/giskard/slicing/slice.py
@@ -19,6 +19,10 @@ def escape(value) -> str:
     return str(value) if type(value) is not str else "%s" % value
 
 
+def _decode(series: pd.Series) -> pd.Series:
+    return series.str.decode("utf-8").fillna(series)
+
+
 class Clause(ABC):
     @abstractmethod
     def mask(self, df: pd.DataFrame) -> pd.Series:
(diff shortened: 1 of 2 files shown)


python-client/giskard/slicing/slice.py
```diff
@@ -19,6 +19,10 @@ def escape(value) -> str:
     return str(value) if type(value) is not str else "%s" % value
 
 
+def _decode(series: pd.Series) -> pd.Series:
+    return series.str.decode("utf-8").fillna(series)
```
rabah-khalek, line 23: what do you need `.fillna(series)` for? what does the decoder do for `NaN` values?

rabah-khalek: Otherwise, I have just tested it, it works.

kevinmessiaen: The `decode` return `NaN` for any value that are not `bytes`. So we decode `bytes` and put the original value for other row that was non `bytes`.

This case it will also wok for mix of `bytes` and `str`:

```python
series = pd.Series()

decoded = series.str.decode("utf-8") # return [NaN, 'this is bytes']

final_value = decoded.fillna(series) # return ['this is str', 'this is bytes']
```

rabah-khalek (approved):

PR #1302, opened 2023-08-07
[GSK-1316, GSK-1415] Spurious Correlation tests

rabah-khalek:
Titanic scan:
Screenshot 2023-08-07 at 14 06 14

Todo: write docstrings

### DIFF
--- a/python-client/giskard/testing/tests/statistic.py
+++ b/python-client/giskard/testing/tests/statistic.py
@@ -297,96 +297,10 @@ def _cramer_v(x, y):
     return stats.contingency.association(ct, method="cramer")
 
 
-@test(name="Cramer's V", tags=["statistic", "classification"])
-def test_cramer_v(
-    model: BaseModel, dataset: Dataset, slicing_function: SlicingFunction, threshold: float = 0.5, debug: bool = False
-) -> TestResult:
-    """
-    TBF
-    :param model:
-    :param dataset:
-    :param slicing_function:
-    :param threshold:
-    :param debug:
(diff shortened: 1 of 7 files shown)


python-client/giskard/testing/tests/statistic.py
```diff
@@ -296,9 +286,155 @@ def test_disparate_impact(
...
+    output_ds = None
+    if not passed and debug:
+        output_ds = sliced_dataset.copy()  # copy all properties
+        test_name = inspect.stack()[0][3]
+        output_ds.name = debug_prefix + test_name
+    # ---
+
+    messages = [TestMessage(type=TestMessageLevel.INFO, text=f"metric = {metric}, threshold = {threshold}")]
+
+    return TestResult(metric=metric, passed=passed, messages=messages, output_df=output_ds)
```
mattbit, line 440: The only difference with the previous test is that we call `_teil_u` instead of `_mutual_information`. This could be refactored in a single `test_nominal_association` with a `method` attribute for example (e.g. theil or mutual_info) to avoid code repetition.

mattbit (changes requested):
- Decide if it’s better to have a single `test_nominal_association` with different methods (could be easier to understand for the user) or specific tests for each metric like `test_theil_u` (this would be for power users I would say) or both.
- Small refactoring of code to use a common `_test_association` with the different methods (internal).

mattbit (approved):
Minor comment: the docstrings do not respect the max line length limit, could you fix that?

PR #1304, opened 2023-08-07
[GSK-1506] Update code snippet for pickled testfunction and its default values

henchaves:
## Description

This PR aims to fix the code snippet displayed for a custom testing function.
Basically, we must provide to user the code to instantiate `GiskardClient` and download the testing function.
Also, we should provide the code to download a slicing function/transformation function if they are linked to the given testing function.

## Related Issue
[GSK-1506 (available on Linear)](https://linear.app/giskard/issue/GSK-1506/code-snippet-for-pickled-function-and-default-values-does-not-work)

## Type of Change

- [ ] 🔐 Security fix


frontend/src/views/main/project/TestsCatalog.vue
```diff
@@ -197,12 +199,44 @@ const selectedTestUsage = computed(() => {
       .map('type')
       .uniq()
       .value()
-      .filter(i => i !== 'str'),
-    selected.value.name
+      .filter(i => i !== 'str')
```
kevinmessiaen, line 202: I think that we should ignore any primitives (`int`, `float`, '`str`, `bool`, ...).

kevinmessiaen: Furthermore we also only need to import when the default value is set

kevinmessiaen (changes requested):

frontend/src/views/main/project/TestsCatalog.vue
henchaves: > Furthermore we also only need to import when the default value is set

The line 196 creates an array of only required params

kevinmessiaen (approved):

PR #1307, opened 2023-08-09
GSK-1533

AbSsEnT:
Added new functionality of SHAP Information logging in the scope of WandB integration.

### DIFF
--- a/python-client/giskard/models/model_explanation.py
+++ b/python-client/giskard/models/model_explanation.py
@@ -40,25 +40,25 @@ def _prepare_for_explanation(input_df: pd.DataFrame, model: BaseModel, dataset: 
     return prepared_df
 
 
-def background_example(df: pd.DataFrame, input_types: Dict[str, str]) -> pd.DataFrame:
-    """Create back-ground example for the SHAP explainer as a mode/median of features of the data to explain."""
+def _get_background_example(df: pd.DataFrame, feature_types: Dict[str, str]) -> pd.DataFrame:
+    """Create background example for the SHAP explainer as a mode/median of features."""
     median = df.median()
-    example = df.mode(dropna=False).head(1)
-    num_columns = [key for key in list(df.columns) if input_types.get(key) == "numeric"]
+    background_sample = df.mode(dropna=False).head(1)
 
-    # Replace a numerical features' mode on their median.
-    for column in num_columns:
-        example[column] = median[column]
+    # Use median of the numerical features.
+    numerical_features = [feature for feature in list(df.columns) if feature_types.get(feature) == "numeric"]
+    for feature in numerical_features:
+        background_sample[feature] = median[feature]
 
-    example = example.astype(df.dtypes)
-    return example
+    background_sample = background_sample.astype(df.dtypes)
+    return background_sample
 
 
 def _calculate_dataset_shap_values(model: BaseModel, dataset: Dataset) -> np.ndarray:
     """Perform SHAP values calculation for samples of a given dataset."""
     # Prepare background sample to be used in the KernelSHAP.
     background_df = model.prepare_dataframe(dataset.df, dataset.column_dtypes, dataset.target)
-    background_sample = background_example(background_df, dataset.column_types)
+    background_sample = _get_background_example(background_df, dataset.column_types)
 
     # Prepare input data for an explanation.
     data_to_explain = _prepare_for_explanation(dataset.df, model=model, dataset=dataset)
@@ -93,7 +93,7 @@ def _calculate_sample_shap_values(model: BaseModel, dataset: Dataset, input_data
         arr_df = pd.DataFrame(array, columns=list(df.columns))
         return model.predict_df(_prepare_for_explanation(arr_df, model=model, dataset=dataset))
 
-    example = background_example(df, dataset.column_types)
+    example = _get_background_example(df, dataset.column_types)
     kernel = KernelExplainer(predict_array, example)
     shap_values = kernel.shap_values(data_to_explain, silent=True)
     return shap_values
--- a/python-client/giskard/models/model_explanation.py
+++ b/python-client/giskard/models/model_explanation.py
@@ -17,18 +17,10 @@ from shap import KernelExplainer, Explanation, Explainer  # noqa
 logger = logging.getLogger(__name__)
 
 
-def _get_cls_prediction_explanation(model: BaseModel, dataset: Dataset, shap_values: list) -> list:
-    """Get SHAP of predictions with the highest probability."""
-    # Get raw indices of predictions with the highest probability.
+def _get_highest_prob_shap(model: BaseModel, dataset: Dataset, shap_values: list) -> list:
+    """Get SHAP explanations of classes with the highest predicted probability."""
     predictions = model.predict(dataset).raw_prediction
-
-    # Select SHAP values, which explain model's predictions with the highest probability.
-    filtered_shap_values = list()
-    for sample_idx, sample_prediction_idx in enumerate(predictions):
-        prediction_explanation = shap_values[sample_prediction_idx][sample_idx]
-        filtered_shap_values.append(prediction_explanation)
-
-    return filtered_shap_values
+    return [shap_values[predicted_class][sample_idx] for sample_idx, predicted_class in enumerate(predictions)]
 
 
 def prepare_df(df: pd.DataFrame, model: BaseModel, dataset: Dataset) -> pd.DataFrame:
@@ -87,7 +79,7 @@ def explain_with_shap(model: BaseModel, dataset: Dataset) -> ShapResult:
 
     # For classification, take SHAP of prediction with the highest probability.
     if model.is_classification:
-        shap_values = _get_cls_prediction_explanation(model, dataset, shap_values)
+        shap_values = _get_highest_prob_shap(model, dataset, shap_values)
 
     # Put shap and feature names to the Explanation object for a convenience.
     feature_names = model.meta.feature_names or list(dataset.df.columns.drop(dataset.target, errors="ignore"))
--- a/python-client/giskard/models/model_explanation.py
+++ b/python-client/giskard/models/model_explanation.py
@@ -51,6 +51,20 @@ def prepare_df(df: pd.DataFrame, model: BaseModel, dataset: Dataset) -> pd.DataF
     return prepared_df
 
 
+def background_example(df: pd.DataFrame, input_types: Dict[str, str]) -> pd.DataFrame:
+    """Create back-ground example for the SHAP explainer as a mode/median of features of the data to explain."""
+    example = df.mode(dropna=False).head(1)
+    median = df.median()
+    num_columns = [key for key in list(df.columns) if input_types.get(key) == "numeric"]
+
+    # Replace a numerical features' mode on their median.
+    for column in num_columns:
+        example[column] = median[column]
+
+    example = example.astype(df.dtypes)
+    return example
+
+
 def explain_with_shap(model: BaseModel, dataset: Dataset) -> ShapResult:
     """Perform SHAP values calculation for samples of a given dataset."""
     feature_names = model.meta.feature_names or list(dataset.df.columns.drop(dataset.target, errors="ignore"))
@@ -126,16 +140,6 @@ def explain_text(model: BaseModel, input_df: pd.DataFrame, text_column: str, tex
     except Exception as e:
         logger.exception(f"Failed to explain text: {text_document}", e)
         raise Exception("Failed to create text explanation") from e
-
-
-def background_example(df: pd.DataFrame, input_types: Dict[str, str]) -> pd.DataFrame:
-    example = df.mode(dropna=False).head(1)  # si plusieurs modes, on prend le premier
-    # example.fillna("", inplace=True)
-    median = df.median()
-    num_columns = [key for key in list(df.columns) if input_types.get(key) == "numeric"]
-    for column in num_columns:
-        example[column] = median[column]
-    return example.astype(df.dtypes)
 
 
 def summary_shap_classification(
--- a/python-client/giskard/models/model_explanation.py
+++ b/python-client/giskard/models/model_explanation.py
@@ -16,7 +16,7 @@ warnings.filterwarnings("ignore", message=".*The 'nopython' keyword.*")
 logger = logging.getLogger(__name__)
 
 
-def _get_highest_prob_shap(model: BaseModel, dataset: Dataset, shap_values: list) -> list:
+def _get_highest_prob_shap(shap_values: list, model: BaseModel, dataset: Dataset) -> list:
     """Get SHAP explanations of classes with the highest predicted probability."""
     predictions = model.predict(dataset).raw_prediction
     return [shap_values[predicted_class][sample_idx] for sample_idx, predicted_class in enumerate(predictions)]
@@ -42,7 +42,7 @@ def _prepare_for_explanation(input_df: pd.DataFrame, model: BaseModel, dataset: 
 
 def _get_background_example(df: pd.DataFrame, feature_types: Dict[str, str]) -> pd.DataFrame:
     """Create background example for the SHAP explainer as a mode/median of features."""
-    median = df.median()
+    median = df.median(numeric_only=True)
     background_sample = df.mode(dropna=False).head(1)
 
     # Use median of the numerical features.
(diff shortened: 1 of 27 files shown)


python-client/giskard/models/model_explanation.py
```diff
@@ -1,20 +1,180 @@
...
+    # Prepare background sample to be used in the KernelSHAP.
+    background_df = model.prepare_dataframe(dataset.df, dataset.column_dtypes, dataset.target)
+    background_sample = background_example(background_df, dataset.column_types)
+
+    # Prepare input data for explanation.
+    input_df = prepare_df(input_data)
+
+    # Obtain SHAP explanations.
+    explainer = shap.KernelExplainer(
+        model=model.predict_df, data=background_sample, feature_names=input_df.columns, keep_index=True
```
rabah-khalek, line 50: I would rather pass `model=model(predict).raw` just for the case where if the predictions was cached, we don't need to calculate them again.

python-client/giskard/models/model_explanation.py
```diff
@@ -1,20 +1,180 @@
...
+    df = pd.DataFrame(data={_FEATURE_COLUMN: feature_names, _SHAP_COLUMN: shap_general_means})
+    table = wandb.Table(dataframe=df)
+    plot = wandb.plot.bar(
+        table, label=_FEATURE_COLUMN, value=_SHAP_COLUMN, title="General Mean(Abs(SHAP)) across all features"
+    )
+
+    return plot
+
+
+def shap_to_wandb(model: BaseModel, dataset: Dataset, **kwargs) -> None:
```
rabah-khalek, line 136: I think this should be split into two functions:
Similar to the way we do:
```py
scan_results = giskard.scan(model, dataset)
scan_results.to_wandb()
```
we should be able to do:
```py
explanation_results = giskard.explain_with_shap(model, dataset)
explanation_results.to_wandb()
```
so you can just name `explain_full` into `explain_with_shap` that should return a `dataclass` containing the `shap_values` (after removing `explainer`). This `dataclass` should support a `to_wandb()` method which is the rest of your `shap_to_wandb()` function.

Something like:
```py
@dataclass
class ShapResults:
  shap_values: np.array 

  def to_wandb(self):
    ...
```

rabah-khalek (changes requested):

python-client/giskard/models/model_explanation.py
```diff
@@ -1,20 +1,180 @@
...
+    """Log SHAP explanation graphs to the WandB run."""
+    from giskard.integrations.wandb.wandb_utils import wandb_run
+
+    with wandb_run(**kwargs) as run:
+        _FEATURE_NAMES = model.meta.feature_names or list(dataset.df.columns.drop(dataset.target, errors="ignore"))
+        _FEATURE_TYPES = {key: dataset.column_types[key] for key in _FEATURE_NAMES}
```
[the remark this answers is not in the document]
AbSsEnT: Solved

python-client/giskard/models/model_explanation.py
```diff
@@ -1,20 +1,180 @@
...
+    # Prepare input data for explanation.
+    input_df = prepare_df(input_data)
+
+    # Obtain SHAP explanations.
+    explainer = shap.KernelExplainer(
+        model=model.predict_df, data=background_sample, feature_names=input_df.columns, keep_index=True
+    )
+    shap_values = explainer.shap_values(input_df, silent=True)
+
+    return shap_values, explainer
```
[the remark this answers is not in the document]
AbSsEnT: Solved

python-client/giskard/models/model_explanation.py
AbSsEnT: I do not understand. 'model' param accepts a reference to a function, but 'model(predict).raw' is a result of a function execution, isn't it?

python-client/giskard/models/model_explanation.py
AbSsEnT: Gonna do this tomorrow.

python-client/giskard/models/model_explanation.py
rabah-khalek: Typo: model.predict(dataset).raw

rabah-khalek: Sorry, in fact, what you wrote initially is correct!

python-client/giskard/models/model_explanation.py
AbSsEnT: Done

AbSsEnT (commented):
Done

python-client/giskard/models/model_explanation.py
```diff
@@ -6,46 +6,114 @@
...
+        shap_values = _get_cls_prediction_explanation(model, dataset, shap_values)
+
+    # Put shap and feature names to the Explanation object for a convenience.
+    feature_names = model.meta.feature_names or list(dataset.df.columns.drop(dataset.target, errors="ignore"))
+    feature_types = {key: dataset.column_types[key] for key in feature_names}
+    shap_explanations = Explanation(values=shap_values, data=dataset.df[feature_names], feature_names=feature_names)
+    return ShapResult(shap_explanations, feature_types, feature_names)
+
+
+def explain_one(model: BaseModel, dataset: Dataset, input_data: Dict) -> np.ndarray:
```
rabah-khalek, line 99: let's refactor this into: `_calculate_sample_shap_values`

python-client/giskard/models/model_explanation.py
```diff
@@ -6,46 +6,114 @@
...
+
+
+def explain_with_shap(model: BaseModel, dataset: Dataset) -> ShapResult:
+    """Get SHAP explanation result."""
+    # Obtain SHAP explanations.
+    shap_values = explain_full(model, dataset)
+
+    # For classification, take SHAP of prediction with the highest probability.
+    if model.is_classification:
+        shap_values = _get_cls_prediction_explanation(model, dataset, shap_values)
```
rabah-khalek, line 90: let's keep this a general function:
- I'd introduce a boolean `only_highest_proba` to `explain_with_shap` and `ShapResult` (to allow the user to choose) as follows:
- I'd introduce `model_type` in `Shapresult`
```py
def explain_with_shap(model: BaseModel, dataset: Dataset, only_highest_proba=True):
  # ...
  if only_highest_proba and model.is_classification:
    shap_values = _get_highest_proba_shap_values(model, dataset, shap_values)

  return ShapResult(shap_explanations, feature_types, feature_names, model.meta.model_type, only_highest_proba)
```
so that:
```python
@dataclass
class ShapResult:
    explanations: Explanation = None
    feature_types: dict = None
    feature_names: list = None
    model_type: ModelType = None
    only_highest_proba: bool = True
```
and then I would check `if self.only_highest_proba and self.model_type == SupportedModelTypes.CLASSIFICATION ` inside `to_wandb` to filter out the proba, and `elif not self.only_highest_proba and self.model_type == SupportedModelTypes.CLASSIFICATION ` I would raise an error saying that we only support SHAP logging with `only_highest_proba=True` for `classification` models.

python-client/giskard/models/shap_result.py
```diff
@@ -0,0 +1,111 @@
...
+    return plot
+
+
+@dataclass
+class ShapResult:
+    explanations: Explanation = None
+    feature_types: dict = None
+    feature_names: list = None
+
+    def to_wandb(self, **kwargs) -> None:
```
rabah-khalek, line 88: see comment above

python-client/tests/integrations/test_wandb.py
```diff
@@ -2,25 +2,36 @@
...
+
     dataset = request.getfixturevalue(dataset_name)
     model = request.getfixturevalue(model_name)
-    _to_wandb(model, dataset)
+    _compare_explain_functions(model, dataset)
+
+    if dataset_name in exception_fixtures:
+        with pytest.raises(NotImplementedError) as e:
+            _to_wandb(model, dataset)
+        assert e.match(r"We do not support the SHAP logging of text*")
```
rabah-khalek, line 34: there should be an else here. 
```py
else:
  _to_wandb(model, dataset)
```
the main point was to check if `to_wandb` works when there's no text features.

python-client/tests/integrations/test_wandb.py
```diff
@@ -34,9 +45,16 @@ def test_fast(dataset_name, model_name, request):
...
```
rabah-khalek, line 57: same as above

python-client/tests/integrations/test_wandb.py
```diff
@@ -51,4 +69,25 @@ def _to_wandb(model, dataset):
...
     test_suite_results.to_wandb()
 
+    # Verify that the logging of the SHAP explanation charts works.
+    explanation_results = explain_with_shap(model, dataset)
+    explanation_results.to_wandb()
+
     assert re.match("^[0-9a-z]{8}$", str(wandb.run.id))
+
+
+def _compare_explain_functions(model, dataset):
```
rabah-khalek, line 79: can you move this to `test_model_explanation.py`?

rabah-khalek (changes requested):

python-client/giskard/models/model_explanation.py
```diff
@@ -6,46 +6,114 @@
...
 
-    df = model.prepare_dataframe(dataset.df, column_dtypes=dataset.column_dtypes, target=dataset.target)
-    feature_names = list(df.columns)
+    # Select SHAP values, which explain model's predictions with the highest probability.
+    filtered_shap_values = list()
+    for sample_idx, sample_prediction_idx in enumerate(predictions):
+        prediction_explanation = shap_values[sample_prediction_idx][sample_idx]
+        filtered_shap_values.append(prediction_explanation)
+
+    return filtered_shap_values
```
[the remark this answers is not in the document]
AbSsEnT: Done

python-client/giskard/models/model_explanation.py
```diff
@@ -6,46 +6,114 @@
...
 from giskard.datasets.base import Dataset
-from giskard.ml_worker.utils.logging import timer
 from giskard.models.base import BaseModel
+from giskard.models.shap_result import ShapResult
+from giskard.ml_worker.utils.logging import timer
 
 warnings.filterwarnings("ignore", message=".*The 'nopython' keyword.*")
-import shap  # noqa
+from shap.maskers import Text  # noqa
+from shap import KernelExplainer, Explanation, Explainer  # noqa
```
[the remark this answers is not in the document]
AbSsEnT: Done

python-client/giskard/models/model_explanation.py
```diff
@@ -6,46 +6,114 @@
...
+
+
+def prepare_df(df: pd.DataFrame, model: BaseModel, dataset: Dataset) -> pd.DataFrame:
+    """Prepare dataframe for an inference step."""
+    df = model.prepare_dataframe(df, column_dtypes=dataset.column_dtypes, target=dataset.target)
+
+    if dataset.target in df.columns:
+        prepared_dataset = Dataset(df, column_types=dataset.column_types, target=dataset.target)
+    else:
+        prepared_dataset = Dataset(df, column_types=dataset.column_types)
```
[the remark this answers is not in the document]
AbSsEnT: Done

python-client/giskard/models/model_explanation.py
```diff
@@ -6,46 +6,114 @@
...
+    # Replace a numerical features' mode on their median.
+    for column in num_columns:
+        example[column] = median[column]
+
+    example = example.astype(df.dtypes)
+    return example
 
-    input_df = prepare_df(pd.DataFrame([input_data]))
+
+def explain_full(model: BaseModel, dataset: Dataset) -> np.ndarray:
```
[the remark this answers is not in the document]
AbSsEnT: Done

python-client/giskard/models/model_explanation.py
AbSsEnT: Done

python-client/giskard/models/model_explanation.py
AbSsEnT: I do not understand, why we need to filter probe once again inside ShapResult().to_wandb(), if we already did this before, in the 'explain_with_shap'

rabah-khalek: yeah I wrote this at the beginning and forgot to update it, of course, the filtering is needed once in `explain_with_shap`

AbSsEnT: thanks!

AbSsEnT: Done

python-client/giskard/models/shap_result.py
AbSsEnT: Done

python-client/tests/integrations/test_wandb.py
AbSsEnT: Done

python-client/tests/integrations/test_wandb.py
AbSsEnT: Done

python-client/tests/integrations/test_wandb.py
AbSsEnT: Done

python-client/giskard/models/shap_result.py
```diff
@@ -0,0 +1,124 @@
...
+
+@dataclass
+class ShapResult:
+    explanations: Explanation = None
+    feature_types: dict = None
+    feature_names: list = None
+    model_type: ModelType = None
+    only_highest_prob: bool = True
+
+    def _validate_config(self):
```
rabah-khalek, line 92: Since it's particular to `wandb`, i'd call it `_validate_wandb_config`

python-client/giskard/models/model_explanation.py
```diff
@@ -4,48 +4,103 @@
...
-            prepared_ds.df, column_dtypes=prepared_ds.column_dtypes, target=prepared_ds.target
-        )
-        columns_in_original_order = (
-            model.meta.feature_names
-            if model.meta.feature_names
-            else [c for c in dataset.df.columns if c in prepared_df.columns]
-        )
-        # Make sure column order is the same as in df
-        return prepared_df[columns_in_original_order]
+def _get_highest_prob_shap(shap_values: list, model: BaseModel, dataset: Dataset) -> list:
```
rabah-khalek: could you please name this `_get_highest_proba_shap` instead?

rabah-khalek (changes requested):

python-client/giskard/models/shap_result.py
AbSsEnT: Done

python-client/giskard/models/model_explanation.py
```diff
@@ -4,48 +4,103 @@
...
+    # Prepare input data for an explanation.
+    data_to_explain = _prepare_for_explanation(dataset.df, model=model, dataset=dataset)
+
+    # Obtain SHAP explanations.
+    explainer = KernelExplainer(model.predict_df, background_sample, data_to_explain.columns, keep_index=True)
+    shap_values = explainer.shap_values(data_to_explain, silent=True)
+    return shap_values
+
+
+def explain_with_shap(model: BaseModel, dataset: Dataset, only_highest_prob: bool = True) -> ShapResult:
```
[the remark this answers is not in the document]
AbSsEnT: Done

python-client/giskard/models/shap_result.py
```diff
@@ -0,0 +1,124 @@
+from enum import Enum
+from typing import Any, Iterable
+from dataclasses import dataclass
+
+import wandb
```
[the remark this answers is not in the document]
AbSsEnT: The point is that I am going to use "shap_results.py" module just in case, when I need to import ShapResults class, which requires wandb. So I do not need anything else from this module, which do not use wandb. i.e. I am gonna use this module only if wandb is installed. Otherwise, I need to put wandb import to all private functions and ShapResults.to_wandb() which looks like an overhead. WDYT?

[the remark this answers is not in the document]
rabah-khalek: since we decided to make `explain_with_shap` a global function that returns `ShapResult`, so that users can use it without necessarily wanting to log it with wandb (maybe later we want to log it to mlflow or other mlops).

several imports of the same libraries is not an issue in python btw, see: https://stackoverflow.com/questions/37067414/python-import-multiple-times. 

If you don't like the idea of import `wandb` in every plotting function, you can refactor them into `wandb_utils.py`, I think that's a cleaner solution.

python-client/giskard/models/model_explanation.py
AbSsEnT: Done

python-client/giskard/models/shap_result.py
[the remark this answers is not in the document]
AbSsEnT: But if I put plotting functions into `wandb_utils.py`, I will also get an error, but defined by us. Do you expect such behaviour and we are ok with it?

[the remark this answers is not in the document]
AbSsEnT: `try:
    import wandb  # noqa
    from wandb.wandb_run import Run
except ImportError as e:
    raise ImportError("The 'wandb' python package is not installed. To get it, run 'pip install wandb'.") from e`

rabah-khalek (approved):

PR #1321, opened 2023-08-21
[GSK-1589] Analytics infinite loop

mattbit:


Hartorn:
Using report error filter out the exception only when it's Giskard one.
If connection to mix panel fails, it will not raise a giskard exception so it will no be caught again

python-client/giskard/utils/analytics_collector.py
```diff
@@ -90,17 +82,16 @@ def get_dataset_properties(dataset):
     }
 
 
-def _report_error(e):
+def _report_error(e, error_type = "python error"):
```
mattbit, line 85: code style

mattbit (commented):
@Hartorn ok for me, just ensure you have the pre-commit hook installed to run black & ruff.

(Can’t approve because I contributed to the PR)

Hartorn (approved):

PR #1333, opened 2023-08-24
[GSK-1552] Fix test suite not being listed on Testing tab

henchaves:
## Description

This PR aims to fix the error generated when a Test Suite without a name is uploaded. To solve that, basically we add a default name to Suite python class if there is no name before uploading it.

## Related Issue

[GSK-1552 (available on Linear)](https://linear.app/giskard/issue/GSK-1552/test-suite-isnt-listed-if-it-doesnt-contain-a-name)

## Type of Change

- [ ] 🔐 Security fix


kevinmessiaen (changes requested):
I think that we should add some check in the backend to throw an error when trying to save a suite with a blank name

backend/src/main/java/ai/giskard/service/TestSuiteService.java
```diff
@@ -49,6 +49,19 @@ public class TestSuiteService {
     private final MLWorkerWSCommService mlWorkerWSCommService;
     private final TestFunctionRepository testFunctionRepository;
 
+    public Long saveTestSuite(String projectKey, TestSuiteDTO dto) {
+        if (dto.getName() == null || dto.getName().isEmpty()) {
+            throw new IllegalArgumentException("Test suite name cannot be empty");
```
kevinmessiaen, line 54: Nit pick, you can use `Strings.isEmpty()` util that is `null` safe. 

IMO we should check for blank to avoid name like `  `

```suggestion
        if (Strings.isBlank()) {
            throw new IllegalArgumentException("Test suite name cannot be blank");
```

kevinmessiaen (approved):
Looks good!

PR #1335, opened 2023-08-25
[GSK-1617] ModelCache does not load cached results correctly on Windows

mattbit:
This is caused by bad formatting of newlines in csv read/write, see [here](https://docs.python.org/3/library/csv.html#id3).

Fixing also a series of other issues:

- Fix cache loading when cache is disabled
- Fix broken cache for text generation models (and added a test)
- Fix broken cache path on Windows
- Fix broken CSV writing on Windows
- Ensure regression models output is casted to float

### DIFF
--- a/python-client/giskard/models/cache/cache.py
+++ b/python-client/giskard/models/cache/cache.py
@@ -35,25 +35,31 @@ class ModelCache:
         self.cache_dir = cache_dir or Path(
             settings.home_dir / settings.cache_dir / "global" / "prediction_cache" / self.id
         )
+        self.vectorized_get_cache_or_na = np.vectorize(self.get_cache_or_na, otypes=[object])
+        self.model_type = model_type
+        self._warmed_up = False
 
+    def warm_up_from_disk(self):
         if id is not None:
             if (self.cache_dir / CACHE_CSV_FILENAME).exists():
                 with open(self.cache_dir / CACHE_CSV_FILENAME, "r") as pred_f:
                     reader = csv.reader(pred_f)
                     for row in reader:
-                        if model_type == SupportedModelTypes.TEXT_GENERATION:
+                        if self.model_type == SupportedModelTypes.TEXT_GENERATION:
                             self.prediction_cache[row[0]] = row[1:]
-                        elif model_type == SupportedModelTypes.REGRESSION:
+                        elif self.model_type == SupportedModelTypes.REGRESSION:
                             self.prediction_cache[row[0]] = float(row[1])
                         else:
                             self.prediction_cache[row[0]] = [float(i) for i in row[1:]]
-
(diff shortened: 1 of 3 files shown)


python-client/giskard/models/cache/cache.py
```diff
@@ -29,29 +30,45 @@ class ModelCache:
...
+                        # Classification models return list of probabilities
+                        self.prediction_cache[row[0]] = [float(i) for i in row[1:]]
+        except Exception as e:
+            warning(f"Failed to load cache from disk for model {self.id}: {e}")
 
     def get_cache_or_na(self, key: str):
         return self.prediction_cache.get(key, NaN)
 
     def read_from_cache(self, keys: pd.Series):
+        if self.id and not self._warmed_up:
```
andreybavt, line 68: when can we have a missing `self.id`? There's `self.id = id or str(uuid.uuid4())` in the constructor

python-client/giskard/models/cache/cache.py
```diff
@@ -29,29 +30,45 @@ class ModelCache:
...
-
+        self.cache_dir = cache_dir or Path(
+            settings.home_dir / settings.cache_dir / "global" / "prediction_cache" / self.id
+        )
+        self.cache_file = self.cache_dir / CACHE_CSV_FILENAME
         self.vectorized_get_cache_or_na = np.vectorize(self.get_cache_or_na, otypes=[object])
+        self.model_type = model_type
+        self._warmed_up = False
+
+    def warm_up_from_disk(self):
```
andreybavt, line 44: Why not warm up at creation time instead of `read_from_cache` ?

mattbit: Because I don’t want to warm up if the cache is disabled. But it’s still better to create the `ModelCache` object in the model, so that it can be used if the cache is enabled at a later time.

python-client/giskard/models/cache/cache.py
mattbit: You are right, I refactored the previous code but I think the generation of new UUID in the constructor should be removed.

python-client/giskard/models/cache/cache.py
andreybavt: I think in case cache is disabled the cleanest will be not to initialize `_cache` property at all and then warm it up in the ModelCache constructor depending on a cache type (in memory only vs FS backed), WDYT?

mattbit: But cache can be disabled temporarily. So if we don’t initialize the `_cache` property at all, we would need to check if it exists at every prediction, initializing it there and possibly warm it up at prediction time, in addition to checking that cache is enabled.

In any case I would avoid doing expensive and persistent operations upon construction of the instance.

andreybavt: ok let's keep it this way

andreybavt (approved):

PR #1338, opened 2023-08-28
[GSK-1587] Add MixPanel to backend and track WebSocket events with general info

Inokinoki:
## Type of Change

- [ ] 🔐 Security fix

## Checklist


backend/src/main/java/ai/giskard/service/AnalyticsCollectorService.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+
+@Service
+@RequiredArgsConstructor
+public class AnalyticsCollectorService {
+    private static final Logger log = LoggerFactory.getLogger(AnalyticsCollectorService.class);
+
+    private static final String DEV_MP_PROJECT_KEY = "4cca5fabca54f6df41ea500e33076c99";
+    private static final String PROD_MP_PROJECT_KEY = "2c3efacc6c26ffb991a782b476b8c620";
```
kevinmessiaen, line 25: It's better to have only one variable that is injected using `@Value` from the profile

kevinmessiaen (changes requested):

backend/src/main/java/ai/giskard/service/AnalyticsCollectorService.java
```diff
@@ -0,0 +1,95 @@
...
+            serverProps.put("Is HuggingFace", GeneralSettingsService.isRunningInHFSpaces);
+            serverProps.put("HuggingFace Space ID", GeneralSettingsService.hfSpaceId);
+            messageBuilder.set(settingsService.getSettings().getInstanceId(), serverProps);
+        } catch (NoSuchElementException e) {
+            // Do not track when we failed to initialize the server properties
+            return;
+        }
+
+        // Start a new thread
+        new Thread(() -> {
```
kevinmessiaen, line 73: We should use a `ThreadPoolExecutor` to improve performance (reusing threads instead of creating new ones each time)

https://www.baeldung.com/thread-pool-java-and-guava

backend/src/main/java/ai/giskard/service/AnalyticsCollectorService.java
```diff
@@ -0,0 +1,95 @@
...
+            // Merge with server info
+            props.put("Java version", javaVersion.toString());
+            props.put("Giskard version", buildVersion);
+
+            JSONObject message = messageBuilder.event(settingsService.getSettings().getInstanceId(), eventName, props);
```
kevinmessiaen, line 78: Since we have a single instance of message builder, we can have another thread that modify this message before we send it from time to time.

To avoid it we should probably create a new MessageBuilder for each event

kevinmessiaen (changes requested):
Looks good to me but we should fix the thread issues

backend/src/main/java/ai/giskard/service/AnalyticsCollectorService.java
Inokinoki: Good observation! Thx

kevinmessiaen (approved):

PR #1340, opened 2023-08-28
persistent api keys

andreybavt:
Save API keys in DB so that they don't expire

image
image


andreybavt:
suppressed by https://github.com/Giskard-AI/giskard/pull/1342

PR #1341, opened 2023-08-28
Fixed LOB column issue for H2 and Postgres

andreybavt:
- fixed LOB column issue for Postgres and H2
- added more verbose logging during the liquibase setup to know for sure which connection it's using

Here are the commands to run tests on H2 and Postgres:
- `../gradlew :backend:test :backend:integrationTest`
- `../gradlew :backend:test :backend:integrationTest -Ptestcontainers`


backend/src/main/java/ai/giskard/domain/BaseEntity.java
```diff
@@ -1,14 +1,15 @@
...
+import jakarta.persistence.Id;
+import jakarta.persistence.MappedSuperclass;
 import lombok.Getter;
 
-import jakarta.persistence.*;
 import java.io.Serializable;
 
 @Getter
 @MappedSuperclass
-@Inheritance(strategy = InheritanceType.TABLE_PER_CLASS)
```
Inokinoki, line 11: What is the base inheritance strategy without this line?

Inokinoki:
LGTM

Should our integration tests be executed under both profiles (w/o `testcontainers`)?

andreybavt:
> Should our integration tests be executed under both profiles (w/o testcontainers)?


@Inokinoki , yes, at least the nightly ones should be run under both profiles since testcontainers are slower than H2, I'll create a separate card for it

PR #1277, opened 2023-08-01
[GSK-1508] Add unit tests for WebSocket-based ML Worker communication (python)

Inokinoki:
## Type of Change

- [ ] 📚 Examples / docs / tutorials / dependencies update
- [ ] 🔧 Bug fix (non-breaking change which fixes an issue)
- [x] 🥂 Improvement (non-breaking change which improves an existing feature)
- [ ] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🔐 Security fix

## Checklist

- [ ] I've read the [`CODE_OF_CONDUCT.md`](https://github.com/Giskard-AI/ai-inspector/blob/master/CODE_OF_CONDUCT.md) document.
- [ ] I've read the [`CONTRIBUTING.md`](https://github.com/Giskard-AI/ai-inspector/blob/master/CONTRIBUTING.md) guide.
- [ ] I've updated the code style using `make codestyle`.
- [ ] I've written tests for all new methods and classes that I created.
- [ ] I've written the docstring in Google format for all the methods and classes that I used.

PR #1342, opened 2023-08-28
Feature/persistent api keys

andreybavt:
Save API keys in DB so that they don't expire

263748828-3927a381-b016-4c74-a3bc-8ed8f7cc7aab
263748930-1fcc5b39-09f3-47c0-a510-939c4353bbc8

> I also extracted `ml-worker-connect` from `/api/v2` into `/public-api` to start distinguishing API authenticated requests from regular internal api requests. In case we add some entry points specifically for public API access it should be under `/public-api` to enforce API Key authentication


python-client/giskard/ml_worker/ml_worker.py
```diff
@@ -107,13 +103,12 @@ def _connect_websocket_client(self, is_server=False):
...
             if not is_server:
-                # External ML Worker: use jwt token
-                # raise ConnectFailedException
+                # External ML Worker: use API key
                 self.ws_conn.connect(
                     with_connect_command=True,
                     wait=False,
                     headers={
-                        "jwt": self.client.session.auth.token,
+                        "api-key": self.client.session.auth.token,
```
Inokinoki, line 111: Will we change auth for internal worker later?

Inokinoki:
LGTM, the tests also passed.

I'll update #1277 accordingly, if any

python-client/giskard/ml_worker/ml_worker.py
andreybavt: yes, internal worker will use persistent API key too, good point

andreybavt: I think we should generate a "system" api key upon first Giskard boot or take it from env variable and use it by internal worker. The benefits compared to current jwt mechanism:
- system api key will remain the same between restarts
- depending on how we generate it it can be known by backend and ml worker at boot time (in case it's set in the env variable). This makes the auth mechanism easier as there's no need to wait for backend to boot to generate a jwt token

Inokinoki: No problem, it will be an internal change.
I created GSK-1663 and we can do it later, apart from this PR.

backend/src/main/java/ai/giskard/security/ee/GiskardAuthFilter.java
```diff
@@ -41,8 +46,17 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         if (!this.licenseService.getCurrentLicense().isActive()) {
             this.noLicenseAuthFilter.doFilter(request, response, chain);
         } else if (this.licenseService.hasFeature(FeatureFlag.AUTH) || httpServletRequest.getHeader(AUTHORIZATION_HEADER) != null) {
+            String authHeader = httpServletRequest.getHeader(HttpHeaders.AUTHORIZATION);
```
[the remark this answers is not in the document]
andreybavt: good catch, it's indeed the same!

PR #1352, opened 2023-08-30
Added "business" detector for the LLM scan

kevinmessiaen:
## Description

Added a new "business" detector that allow to scan generate functional issues for LLM tasks:

- Generate input dataset based on the function tasks
- Generate functional tests
- Updated debugger to handle LLM

## Type of Change

- [ ] 🔐 Security fix


python-client/giskard-2.0.0b14-py3-none-any.whl
mattbit, line 1: ???

python-client/giskard/scanner/llm/business_detector.py
```diff
@@ -0,0 +1,136 @@
...
+                    examples=df_with_pred_and_test_results[failed],
+                    tests=_generate_business_test,
+                )
+            )
+
+    return issues
+
+
+@detector("llm_business", tags=["business", "llm", "generative", "text_generation"])
+class LLMBusinessDetector:
```
mattbit, line 67: I think we should find a better name for this @jmsquare

python-client/giskard/llm/issues/__init__.py
mattbit, line 1: Can we move all of these inside the detector folder?
Also the prompts and everything. It’s quite tightly bound to the scan detector.

python-client/giskard/llm/utils/validate_test_case.py
mattbit, line 1: Same, should be moved the the scan

python-client/giskard/scanner/llm/business_detector.py
```diff
@@ -0,0 +1,136 @@
...
+
+                maybe_print(f"Generated potentially failing prompts: {potentially_failing_dataset.df}", verbose=verbose)
+
+                predictions = model.predict(potentially_failing_dataset).prediction
+
+                issues += validate_prediction(
+                    model, issue, test_cases, potentially_failing_dataset, predictions, self.threshold, verbose
+                )
+            except Exception as e:
+                maybe_print(f"Failed to evaluate {issue}: {e}", verbose=verbose)
```
mattbit, line 119: Should at least be a UserWarning

python-client/giskard/scanner/llm/business_detector.py
```diff
@@ -0,0 +1,136 @@
...
+        issues = []
+
+        for issue in LLM_ISSUE_CATEGORIES:
+            try:
+                test_cases = (
+                    issue.issue_generator(self.num_tests)
+                    .run_and_parse(model_name=model.meta.name, model_description=model.meta.description)
+                    .assertions[: self.num_tests]
+                )
+                maybe_print(f"Generated tests: {test_cases}", verbose=verbose)
```
mattbit, line 84: Note that for all detectors we use `logging.debug` or `info` instead of printing. We should do the same here for consistency.

python-client/giskard/scanner/llm/business_detector.py
```diff
@@ -0,0 +1,136 @@
...
+                    .run_and_parse_with_prompt(
+                        model_name=model.meta.name,
+                        model_description=model.meta.description,
+                        variables=model.meta.feature_names,
+                        generated_tests=test_cases,
+                    )
+                    .input[: self.num_samples]
+                )
+
+                potentially_failing_dataset = Dataset(
```
mattbit, line 97: Can we say that "potentially_failing_dataset" is just the "evaluation dataset"?

python-client/giskard/scanner/llm/toxicity_detector.py
mattbit, line 1: I think that toxicity detector could in principle be kept if we replace some input vars with prompt injection instead of rewriting the prompt.

python-client/giskard/scanner/report.py
```diff
@@ -98,6 +99,9 @@ def generate_test_suite(self, name=None):
         self._track_suite(suite, name)
         return suite
 
+    def generate_dataset(self):
```
mattbit, line 102: Does this belong here? I‘m not sure.

python-client/giskard/scanner/scanner.py
```diff
@@ -172,14 +171,17 @@ def _postprocess(self, issues: Sequence[Issue]) -> Sequence[Issue]:
...
-            )
-            from .llm.utils import load_default_dataset
-
-            dataset = load_default_dataset()
-            model = model.rewrite_prompt(
-                template="{question}", input_variables=["question"], feature_names=["question"]
+            logger.warning("No dataset provided. Inferring dataset")
+            from .llm.utils import infer_dataset
+
+            # TODO: remove this to use the datasets generated by BusinessDetector
```
mattbit, line 177: ??? Not clear to me

python-client/giskard/scanner/scanner.py
```diff
@@ -172,14 +171,17 @@ def _postprocess(self, issues: Sequence[Issue]) -> Sequence[Issue]:
...
+            dataset = infer_dataset(
```
mattbit, line 178: `infer_dataset` is generating a fake dataset, right? Why "infer"?

mattbit (changes requested):
Generally ok, code could be made more readable. Left a few minor comments, althought I have a major question: can we use OpenAI function calling? This can potentially make all of the evaluations much cheaper in terms of time and number of tokens. We could also just make a single call to the OpenAI API instead of relying on langchain chains.

python-client/giskard/scanner/scanner.py
kevinmessiaen: Actually I forgot to remove the `infer_dataset`. The idea is that each issue category generate a small dataset focused on the category inside the `BusinessDetector` and therefore it makes more sense to return the small datasets combined together instead of a more generic one that is not used in the scan anymore.

python-client/giskard/scanner/report.py
kevinmessiaen: We want to be able to extract the dataset generated by the scan. It's more for user that don't want to directly interact with the scan.

python-client/giskard/scanner/llm/business_detector.py
kevinmessiaen: Yes its the evaluation dataset your correct

kevinmessiaen:
Moved to https://github.com/Giskard-AI/giskard/pull/1450 after resolving conflict

PR #1355, opened 2023-08-30
[GSK-1623] Secure Giskard readonly demo space at Hugging Face Spaces

Inokinoki:
## Description

~~Preventing any requests with modifications (POST, PUT, DELETE) when `giskardai/giskard` detected.~~

Preventing Create, Update, Delete operations in DB after initialization.

## Type of Change

- [x] 🔐 Security fix

## Checklist


andreybavt:
since we're sending some POST requests like `predict` or `explain` we can't actually block on HTTP method level.

Instead I suggest we do it on a DB level (more precisely at JPA).

for example, if you define a class like
```java
import jakarta.persistence.PrePersist;
import jakarta.persistence.PreRemove;
import jakarta.persistence.PreUpdate;


public class ReadOnlyListener {
    private boolean isReadOnlyGiskardInstance() {
        // TODO: implement this, probably in another class
        return true;
    }

    @PrePersist
    @PreUpdate
    @PreRemove
    void onCUD(Object o) {
        if (isReadOnlyGiskardInstance()) {
            throw new IllegalStateException("This is a read-only Giskard instance. You cannot modify entities.");
        }
    }
}
```

Then you'll be able to use it at an `@Entity` level:
```java
@Entity(name = "api_keys")
@Getter
@NoArgsConstructor
@EntityListeners(ReadOnlyListener.class)
public class ApiKey extends AbstractAuditingEntity
```

In this case we could even allow some modifications but not others, like writing `TestSuiteExecution` (I think we could also have a scheduled method to delete old executions on demo instance so that the DB doesn't get polluted to much)

Inokinoki:
Also resolve GSK-1592

backend/src/main/java/ai/giskard/domain/ApiKey.java
```diff
@@ -1,15 +1,19 @@
...
+import lombok.Setter;
 import org.apache.commons.lang3.RandomStringUtils;
 
 import java.util.UUID;
 
 @Entity(name = "api_keys")
 @Getter
+@Setter
 @NoArgsConstructor
+@EntityListeners(GalleryDatabaseOperationListener.class)
```
kevinmessiaen, line 16: We should probably move the `@EntityListeners(GalleryDatabaseOperationListener.class)` annotation to the `AbstractAuditingEntity` and/or `BaseEntity` to avoid having to put it everywhere and forgot to add it for future entities

frontend/src/views/main/admin/settings/SettingsGeneral.vue
```diff
@@ -16,7 +16,14 @@
...
                   <tr>
                     <td>Instance</td>
-                    <td>{{ appSettings.generalSettings?.instanceId }}</td>
+                    <td>{{ appSettings.generalSettings?.instanceId }}
+                      <v-btn v-if="appSettings.isDemoHfSpace && !isUnlocked" icon @click="unlockModal = true">
+                        <v-icon>mdi-lock</v-icon>
+                      </v-btn>
+                      <v-btn v-if="appSettings.isDemoHfSpace && isUnlocked" icon @click="unlockModal = true">
+                        <v-icon>mdi-lock-open</v-icon>
+                      </v-btn>
```
kevinmessiaen, line 25:
```suggestion
                      <v-btn v-if="appSettings.isDemoHfSpace" icon @click="unlockModal = true">
                        <v-icon v-if="isUnlocked">mdi-lock-open</v-icon>
                        <v-icon v-else>mdi-lock</v-icon>
                      </v-btn>
```

backend/src/main/java/ai/giskard/web/rest/errors/GalleryDemoSpaceException.java
```diff
@@ -0,0 +1,9 @@
+package ai.giskard.web.rest.errors;
+
+import org.springframework.http.HttpStatus;
+
+public class GalleryDemoSpaceException extends GiskardException {
+    public GalleryDemoSpaceException(String description) {
+        super(HttpStatus.SERVICE_UNAVAILABLE, description);
```
kevinmessiaen, line 7: I would suggest to use NOT_ALLOWED (https://www.rfc-editor.org/rfc/rfc7231#section-6.5.5) that seems more suitable for us.

SERVICE_UNAVAILABLE is more when the server is overloaded or down.

backend/src/main/java/ai/giskard/web/rest/controllers/GalleryUnlockController.java
```diff
@@ -0,0 +1,36 @@
...
+    public GalleryUnlockDTO getUnlockStatus() {
+        GalleryUnlockDTO unlockDTO = new GalleryUnlockDTO();
+        unlockDTO.setUnlocked(InitService.isUnlocked());
+        unlockDTO.setToken(null);
+        return unlockDTO;
+    }
+
+    @PostMapping("")
+    public GalleryUnlockDTO setUnlockStatus(@RequestBody GalleryUnlockDTO unlockDTO) {
+        if (StringUtils.hasText(unlockDTO.getToken()) &&
```
kevinmessiaen, line 28: We should probably throw an exception is the token is not provided/wrong

backend/src/main/java/ai/giskard/service/InitService.java
```diff
@@ -143,19 +145,30 @@ public String getProjectByCreatorLogin(String login) {
...
+        initDefaultApiKey();
         List<String> profiles = Arrays.asList(env.getActiveProfiles());
         if (!profiles.contains("prod") && !profiles.contains("dev")) {
             initProjects();
         }
+        lockIndicator.setUnlocked(false);
+    }
+
+    private void initDefaultApiKey() {
+        apiKeyService.initDefaultApiKey();
```
kevinmessiaen, line 171: Why don't we just call the service line 162?

kevinmessiaen (changes requested):

backend/src/main/java/ai/giskard/web/rest/errors/GalleryDemoSpaceException.java
Inokinoki: Is it more generally used in the case where the HTTP method is not allowed?
I thought about adopting it thanks to our RESTful APIs, just worrying about that some APIs might also return NOT_ALLOWED when using an inappropriate HTTP method.

backend/src/main/java/ai/giskard/service/InitService.java
Inokinoki: Thought about some more operations for it but finally dropped.
Will improve it

backend/src/main/java/ai/giskard/web/rest/errors/GalleryDemoSpaceException.java
kevinmessiaen: Well the rule of thumbs is that 5xx means server error while 4xx means client error. Technically we also can have a 503 for other things.

In this case we can indeed have a 405 error when we try to `POST` a `GET` endpoint only. In our point it technically is the case as the `POST` is disabled when readonly if that makes sense. But subjectively it also could be a 403 error.

I think we can use the error message description to differenciate between different source (or response body)

kevinmessiaen (approved):

PR #1359, opened 2023-08-31
Feature/gsk 1649 use openapi typescript generation tools instead of j2ts

andreybavt:
Added the necessary infrastructure to generate a typescript client based on Giskard's OpenAPI specs

This new autogenerated API will eventually replace the old `api.ts`, the advantages are:
- no more need to manually call J2TS to generate typescript DTOS
- the whole client is auto-created, not just DTOs

This PR doesn't replace the old `api.ts`, but contains a few examples. We should remove old `api.ts` usage gradually.

### How to use:

There's a new gradle command to generate a web client:
`./gradlew generateOpenApiClient`

The generated code will be stored in `frontend/src/generated`

the new `generateOpenApiClient` task is already linked with the backend and frontend build, so simply running:

`./gradlew build` 

will also do the necessary to generate the client

### Example of use:
https://github.com/Giskard-AI/giskard/blob/c7652cb5eae7be13124e6af86ce9e82f61e41829/frontend/src/stores/main.ts#L95

> once we finish migrating `api.*` to `openapi.*` we can delete `api.ts` and rename `openapi` to `api` to keep it simple


Googleton (approved):
LGTM. I guess there isn't that much work on migrating, is it just replacing api to api-v2 calls ?

kevinmessiaen (approved):

PR #1361, opened 2023-08-31
GSK-1507

AbSsEnT:
Added 6 refurbished example notebooks.


AbSsEnT:
@rabah-khalek please, check if it is ok to skip the issue caused by the GitGuardian

rabah-khalek:
Seems like it's the old API key (not persisting one) so i guess it's fine. @andreybavt could you confirm?

@AbSsEnT could you please make sure that you remove it in a new commit please? Thanks

PR #1362, opened 2023-09-01
[GSK-1597] Push typo perturbation stochasticity

rabah-khalek:
I had to hash the typo transformation in order to be able to retrieve the same random perturbation when the user go back and forth in the debugger. 

The problem is described by two screenshots on linear.


python-client/giskard/push/perturbation.py
```diff
@@ -164,15 +168,28 @@ def _text(
...
+        t = text_transformation(column=feature, **kwargs)
 
         # Transform the slice
-        transformed = ds_slice_copy.transform(t)
+        if _is_typo_transformation:
+            _hash = xxh3_128_hexdigest(
+                f"{', '.join(map(lambda x: repr(x), ds_slice_copy.df.values))}".encode("utf-8")
+            )
+            if _hash not in hashed_typo_transformations.keys():
+                hashed_typo_transformations.update({_hash: ds_slice_copy.transform(t)})
```
andreybavt, line 189: this cache will never be emptied as long as the worker is running, can you make it a fixed-sized LRU cache?

andreybavt: why only caching typo transformations and not the rest?

andreybavt (changes requested):

python-client/giskard/push/perturbation.py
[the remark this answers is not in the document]
rabah-khalek: Agreed (https://stackoverflow.com/questions/15456158/python-dict-update-vs-subscript-to-add-a-single-key-value-pair).

rabah-khalek: > this cache will never be emptied as long as the worker is running, can you make it a fixed-sized LRU cache?

I think it's better to keep the cache as long as possible, it'll be confusing to use for instance `@lru_cache(maxsize=16)`, as user might loose a push notification about typo transformation at some point without knowing why. Is there a problem if the caches is never emptied?

> why only caching typo transformations and not the rest?

Among these, typo transformations are the only ones that are random
```py
    TextGenderTransformation,
    TextLowercase,
    TextPunctuationRemovalTransformation,
    TextTitleCase,
    TextTypoTransformation,
    TextUppercase,
```

python-client/giskard/push/perturbation.py
```diff
@@ -167,9 +168,16 @@ def _text(
...
     # Iterate over the possible text transformations
-    for text_transformation in text_transfo_list:
+    for text_transformation in text_transformation_list:
         # Create the transformation
-        t = text_transformation(column=feature)
+        _is_typo_transformation = issubclass(text_transformation, TextTypoTransformation)
+        kwargs = {}
+        if _is_typo_transformation:
+            hashed_seed = hash(f"{', '.join(map(lambda x: repr(x), ds_slice_copy.df.values))}".encode("utf-8"))
+            positive_hashed_seed = hashed_seed % ((sys.maxsize + 1) * 2)
```
andreybavt, line 177: @rabah-khalek could you add a commend why we're doing it? In a week we won't remember by heart

andreybavt (approved):

PR #1375, opened 2023-09-04
Vue2 -> Vue 3

andreybavt:
wip


andreybavt:
to be continued later...

PR #1377, opened 2023-09-04
[GSK-1634] From MAD to 10% change per sample

rabah-khalek:
Changing the perturbation of numerical features, from +/- 2*MAD, to 10%*value.

### DIFF
--- a/python-client/giskard/ml_worker/testing/functions/transformation.py
+++ b/python-client/giskard/ml_worker/testing/functions/transformation.py
@@ -158,3 +158,13 @@ def mad_transformation(
         value_added = compute_mad(data[column_name])
     data[column_name] = data[column_name].apply(lambda x: x + factor * value_added)
     return data
+
+
+@transformation_function(name="ADD value", tags=["num"], row_level=False)
+def add_value(data: pd.DataFrame, column_name: str, value_added: float = 0.0) -> pd.DataFrame:
+    """
+    Add the value_added to the column.
+    """
+    data = data.copy()
+    data[column_name] = data[column_name].apply(lambda x: x + value_added)
+    return data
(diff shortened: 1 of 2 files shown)


python-client/giskard/ml_worker/testing/functions/transformation.py
```diff
@@ -158,3 +158,13 @@ def mad_transformation(
         value_added = compute_mad(data[column_name])
     data[column_name] = data[column_name].apply(lambda x: x + factor * value_added)
     return data
+
+
+@transformation_function(name="ADD value", tags=["num"], row_level=False)
```
andreybavt, line 163: nitpick: does "ADD" need to be capitalcase?
```suggestion
@transformation_function(name="Add value", tags=["numerical"], row_level=False)
```

python-client/giskard/push/perturbation.py
```diff
@@ -195,23 +190,24 @@ def _numeric(
     transformation_functions_params,
     value_perturbed,
 ):
-    # Compute the MAD of the column
-    mad = compute_mad(ds.df[feature])  # Small issue: distribution might not be normal
+    # Compute 10% around the value to be perturbed
+    value_to_perturb = ds_slice.df[feature].iloc[0]
```
andreybavt, line 194: previously we were calculating MAD based on the whole series, now we're taking just the first element, isn't it less robust in case the first element is an outlier?

rabah-khalek: by first element, it's the sample we get in the debugger. Yes, @jmsquare suggested based on @AbSsEnT 's feedback, that we reduce the perturbation to a fixed rate (in this case 10% or less of the sample-value) instead of using MAD because it's less clear where it comes from, plus the perturbation was really sizeable (see the linear card for examples).

PR #1348, opened 2023-08-29
Fixed issues with upload

Googleton:
PR made on @kevinmessiaen's behalf to quickly merge it 🙏

PR #1381, opened 2023-09-05
feature/gsk-1692-catalog-not-showing-full-list-of-tests

andreybavt:


frontend/src/stores/catalog.ts
```diff
@@ -18,7 +18,7 @@ interface State {
 
 function latestVersions<E extends CallableDTO>(data?: Array<E>): Array<E> {
     return chain(data ?? [])
-      .groupBy(func => func.module ?? `${func.module}.${func.name}`)
+      .groupBy(func => `${func.module}.${func.name}` ?? func.module)
```
kevinmessiaen, line 21: Actually this is the same as `${func.module}.${func.name}` since the `??` would never be triggered.

kevinmessiaen (commented):
IMO we should probably have something that check the `displayName` if set to have something more robust. As the current solution will merge all lambda generated from same package (I know @Googleton  faced the issue in the push feature) and also if we rename a module (happened during a version migration)

It's similar to this PR: https://github.com/Giskard-AI/giskard/pull/1348/files

frontend/src/stores/catalog.ts
andreybavt: ah yes, sure 🤦 , let me clean it up

andreybavt: I ported a similar fix to https://github.com/Giskard-AI/giskard/pull/1348

https://github.com/Giskard-AI/giskard/pull/1348/commits/dec4a5ad8d9df11c1fe12f5a0936bf1ba51620c2

andreybavt:
superseded by https://github.com/Giskard-AI/giskard/pull/1348

PR #1382, opened 2023-09-06
Feature/gsk 1618 run python tests on windows builder

Hartorn:
## Type of Change

- [ ] 🔐 Security fix

## Checklist


.github/workflows/test-python-client.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Test
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: 1 #TODO(Bazire): put back dynamic number
+        run: ./gradlew :python-client:test-all --info --parallel
+
+      - name: Package
+        run: ./gradlew :python-client:package --info --parallel
+
+      - name: Lint
```
andreybavt, line 99: linting should happen first because it's a lot faster compared to tests execution. In case there are linting issues it'll fail fast

python-client/giskard/datasets/base/__init__.py
```diff
@@ -5,7 +5,7 @@
 import uuid
 from functools import cached_property
 from pathlib import Path
-from typing import Dict, Optional, List, Union, Hashable
+from typing import Dict, Hashable, List, Optional, Union
```
andreybavt, line 8: @Hartorn is it `black` that did the reformatiing?

andreybavt: if yes, could you:
- create a separate branch from `main`
- run black on the same files as you modified in this pr
- create a separate PR with only black modifications
- once this pr is merged merge `main` into `feature/gsk-1618-run-python-tests-on-windows-builder` 

This should only add the relevant changes in the diff of this PR, otherwise it's difficult to review :/

Hartorn: It must be black or isort.
Sorry about that, I will start from main with black & isort formatting, then, I will rebase.

Hartorn: I will also disable my toml linter.

.github/workflows/build-python-client.yml
```diff
@@ -0,0 +1,98 @@
...
+      matrix:
+        python-version: [ "3.8", "3.9", "3.10" ]
+        os: [ubuntu-latest]
+        experimental: [false]
+        # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+        include:
+          # - python-version: "3.11"
+          #   os: ubuntu-latest
+          #   experimental: true
+          # - python-version: "3.10"
```
andreybavt, line 26: why did you leave it commented out?

.github/workflows/build-python-client.yml
```diff
@@ -0,0 +1,98 @@
...
+        with:
+          path: python-client/.venv
+          key: ${{ matrix.os }}-${{ matrix.python-version }}-python-deps-${{ hashFiles('python-client/tests/fixtures/**/*py')}}
+          restore-keys: ${{ matrix.os }}-${{ matrix.python-version }}-python-deps
+
+      - name: Setup PDM
+        uses: pdm-project/setup-pdm@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: false
```
andreybavt, line 57: should we activate the caching of pdm installation here?

.github/workflows/build-python-client.yml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Cache Giskard test resources
+        uses: actions/cache@v3
+        with:
+          path: ~/.giskard
+          key: ${{ matrix.os }}-${{ matrix.python-version }}-python-test-resources-${{ hashFiles('python-client/tests/fixtures/**/*py')}}
+          restore-keys: ${{ matrix.os }}-${{ matrix.python-version }}-python-giskard-test-resources
+
+      - name: Install dependencies
+        working-directory: python-client 
+        run: pdm install -G :all
```
andreybavt, line 73: we don't probably need all groups here (`dev` for example) do we? `prod`, `test` and `doc` should be fine

.github/workflows/build-python-client.yml
```diff
@@ -0,0 +1,98 @@
...
+        run: pdm run lint
+
+      - name: Test code
+        working-directory: python-client
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: ${{ matrix.os == 'windows-2019' && 1 || 2 }}
+        run: pdm run test-fast
+  
+      - name: Build doc
+        working-directory: python-client
```
andreybavt, line 86: I think we can save some time by not building the docs for all python version and architectures since we only use 1 kind of setup on readthedocs

.github/workflows/build-python-client.yml
```diff
@@ -0,0 +1,98 @@
...
+
+      - name: Lint code
+        working-directory: python-client 
+        run: pdm run lint
+
+      - name: Test code
+        working-directory: python-client
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: ${{ matrix.os == 'windows-2019' && 1 || 2 }}
+        run: pdm run test-fast
```
andreybavt, line 83: don't forget to define `GSK_DISABLE_ANALYTICS` as `true` otherwise we'll have a ton of irrelevant telemetry

.github/workflows/sonar.yml
```diff
@@ -67,77 +62,16 @@ jobs:
...
-          path: python-client/.venv
-          key: ${{ runner.os }}-${{ matrix.python-version }}-client-${{ hashFiles('python-client/pdm.lock')}}
-          restore-keys: ${{ runner.os }}-client
-
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16
-
       - name: Set up JDK 17
         uses: actions/setup-java@v3
```
andreybavt, line 66: it might be faster to use `gradle-build-action` because it might cache gradle installation itself

python-client/giskard/datasets/base/__init__.py
```diff
@@ -659,16 +659,27 @@ def copy(self):
...
+            # Get name from file
+            artifact_name = Path(f.name).name
+            # Write the file on disk
+            f.write(save_df(self.df))
+        try:
             if mlflow_client is None and mlflow_run_id is None:
                 mlflow.log_artifact(local_path)
             elif mlflow_client and mlflow_run_id:
                 mlflow_client.log_artifact(mlflow_run_id, local_path=local_path)
+            # Note(Bazire): Should we have a raising else here ?
```
andreybavt, line 678: I'd say yes, to make sure there's no case where one of client or run_id are None
WDYT @rabah-khalek

python-client/giskard/datasets/base/__init__.py
```diff
@@ -659,16 +659,27 @@ def copy(self):
...
+        finally:
+            # Force deletion of the temps file
+            Path(f.name).unlink(missing_ok=True)
```
andreybavt, line 681: deletion from `NamedTemporaryFile`'s `__exit__` didn't do the job?

python-client/tests/fixtures/amazon_review__binary_classification.py
```diff
@@ -25,7 +24,7 @@
 FEATURE_COLUMN_NAME = "reviewText"
 
 # Data.
-DATA_URL = os.path.join("ftp://sys.giskard.ai", "pub", "unit_test_resources", "amazon_review_dataset", "reviews.json")
+DATA_URL = '/'.join(["ftp://sys.giskard.ai", "pub", "unit_test_resources", "amazon_review_dataset", "reviews.json"])
```
andreybavt, line 27: might as well just be a hardcoded URL, no benefit of using join here...

python-client/tests/fixtures/drug_classification__multiclass_classification.py
```diff
@@ -15,8 +14,8 @@
 
 
 # Data.
-DATA_URL = os.path.join(
-    "ftp://sys.giskard.ai", "pub", "unit_test_resources", "drug_classification_dataset", "drug200.csv"
+DATA_URL = '/'.join(
+    ["ftp://sys.giskard.ai", "pub", "unit_test_resources", "drug_classification_dataset", "drug200.csv"]
```
andreybavt, line 18: same as the other url: can be hardcoded

andreybavt (changes requested):

.github/workflows/build-python-client.yml
Hartorn: I was wondering if we should run it for mac os or not.
Same for windows-2022

for python3.11, since I changed the pyproject.toml, no need to run, I can remove it

.github/workflows/build-python-client.yml
andreybavt: ok, could you create a linear card to set the flag to `true` and reference this card it in a comment so that we don't forget

.github/workflows/build-python-client.yml
Hartorn: Oh, I thought it was for java backend, my bad.

.github/workflows/build-python-client.yml
andreybavt: got it, lets not overcomplicate it by adding all the runner types. I type of windows should be fine (I believe 2019 is a more common one these days, right?)
In any case let's clean up the commented out runners

.github/workflows/build-python-client.yml
Hartorn: Will do

.github/workflows/build-python-client.yml
andreybavt: it's taken into account by pydantic. The settings are defined in `settings.py` and every setting can be overriden from env variable with a GSK_ prefix (kind of like for Spring)

python-client/giskard/datasets/base/__init__.py
Hartorn: I had to put delete=False when opening tempfile.NamedTemporaryFile

When we are doing 
```
with tempfile.NamedTemporaryFile(
            prefix="dataset-", suffix=".csv", delete=False
        ) as f:
```
It boths create AND open the file (with mode w)

But as stated in the doc (https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)
>Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows). 

So for this to work for windows, we have to create a temp file, write it, close, read it and then only delete it.

.github/workflows/sonar.yml
Hartorn: I will try that, but after this one I think.
I still have to work on CODEOWNER, gradle cache and some other stuff, but this PR is already a bit too big I think

.github/workflows/build-python-client.yml
andreybavt: but it can be a good motivation to move the dependencies in the right groups and make sure that people don't make mistakes in the future (for example if you add something to dev and not to test the unit test will fail)

.github/workflows/sonar.yml
andreybavt: ok, could you create a separate card please?

python-client/giskard/datasets/base/__init__.py
andreybavt: ok, sounds good

.github/workflows/sonar.yml
Hartorn: Done ! https://linear.app/giskard/issue/GSK-1717/use-gradle-build-action-instead-of-calling-gradlew

.github/workflows/build-python-client.yml
Hartorn: I will split that into a separated PR then, if it's ok : https://linear.app/giskard/issue/GSK-1717/use-gradle-build-action-instead-of-calling-gradlew

.github/workflows/build-python-client.yml
Hartorn: Done : https://linear.app/giskard/issue/GSK-1712/activate-cache-on-pdm-setup-in-github-action

andreybavt (approved):

PR #1385, opened 2023-09-07
[GSK-1700] Issue when df has single column

rabah-khalek:
the example is attached on the linear card.


python-client/giskard/models/base/wrapper.py
```diff
@@ -99,6 +100,14 @@ def __init__(
     def _preprocess(self, data):
         if self.data_preprocessing_function:
             return self.data_preprocessing_function(data)
+
+        # In the case where a dataframe has one column, most sklearn models require a Series instead of a DataFrame.
+        # We check also the case of PredictionFunctionModel to be extra inclusive.
```
andreybavt, line 105: @rabah-khalek could you add a unit test for it please?

andreybavt (changes requested):

python-client/giskard/models/base/wrapper.py
rabah-khalek: done!

andreybavt (approved):

PR #1388, opened 2023-09-08
Fixing pdm lock running forever, and cleaning up some files

Hartorn:
## Description

Atm, when trying to run `pdm lock -G :all`, it will not work

It's mainly beacause of `safety` package, which depends on `packaging<22.0,>=21.0`
For exemple, `black` version 23.7 depends on packaging > 22
Also, trying to keep version 3.11 of python handled seems to be problematic

## Type of Change

- [ ] 🔐 Security fix

## Checklist


python-client/pyproject.toml
```diff
@@ -54,7 +55,7 @@ dev = [
     "pytest-xdist>=3.3.1",
     "ruff>=0.0.271",
     "mlflow>2",
-    "black[d]>=22.12.0",
```
andreybavt, line 57: `blackd` is used by intellij, it runs as a server and then intellij will call it every time you need to run reformatting.
`black[jupyter]` doesn't ship blackd command utility

python-client/pyproject.toml
```diff
@@ -71,8 +72,9 @@ test = [
     "tensorflow-macos>=2.10.0, <2.11; sys_platform == 'darwin' and platform_machine == 'arm64'",
     "tensorflow>=2.10.0, <2.11; sys_platform != 'darwin' or platform_machine != 'arm64'",
     "tensorflow-text>=2.10.0, <2.11; platform_machine == 'x86_64' or platform_machine == 'amd64'",
-    "mlflow>2",
-    "wandb"
+    "mlflow>2",                                                                                    # Should it be somewhere else ?
```
andreybavt, line 75: actually I don't know why we install a full mlflow in dev mode while mlflow-skinny being part of prod dependencies, maybe @rabah-khalek knows

python-client/pyproject.toml
```diff
@@ -71,8 +72,9 @@ test = [
...
+    "wandb",                                                                                       # Should it be somewhere else ?
```
andreybavt, line 76: no it's an optional dependency that we'll ask users to install if they try to use W&B integration.
We install it for tests because there are unit tests for it

python-client/pyproject.toml
rabah-khalek: `mlflow.evaluate()` (where we implemented our plug-in) is not shipped with `mlflow-skinny`. It is needed for functional tests of the plugin.

python-client/pyproject.toml
Hartorn: I will remove comment then

python-client/pyproject.toml
```diff
@@ -18,32 +24,27 @@ url = "https://download.pytorch.org/whl/cpu/torch_stable.html"
...
     "jupyter>=1.0.0",
     "jupyterlab>=3.4.2",
     "pre-commit>=2.19.0",
-    "pydocstyle>=6.1.1",
-    "pylint>=2.13.9",
-    "pyupgrade>=2.32.1",
-    "safety>=1.10.3",
     "mypy>=0.982",
     "deptry>=0.5.13",
-    "httpretty>=1.1.4",
```
andreybavt, line 46: was it `deptry` who sugested the removal?

python-client/pyproject.toml
```diff
@@ -72,7 +73,8 @@ test = [
     "tensorflow>=2.10.0, <2.11; sys_platform != 'darwin' or platform_machine != 'arm64'",
     "tensorflow-text>=2.10.0, <2.11; platform_machine == 'x86_64' or platform_machine == 'amd64'",
     "mlflow>2",
-    "wandb"
+    "wandb",
+    "tensorflow-io-gcs-filesystem==0.27.0",                                                        # Transitive dependency, needed to make it work for windows (downgrade from 0.33 to 0.27)
```
andreybavt, line 77: should we add `sys_platform` marker in this case? And let other OS users install more recent versions?

python-client/pyproject.toml
Hartorn: No, this one I checked it manually I think.

I used deptry, but I also had a look around on deps to see if some were unused.

python-client/pyproject.toml
andreybavt: otherwise we'll be stuck with python <3.11 forever

Hartorn: Well, this version is the recommended one according to https://github.com/tensorflow/io
I checked the matrix there, that's why I fixed it.

Honestly, we'll need to make changes to every tf package to handle 3.11 after.

andreybavt: could you also tell what exactly breaks on windows with more recent versions of `tensorflow-io-gcs-filesystem` ?

Hartorn: Basically, it cannot be installed. It's trying to find a version, but cannot find one matching the needed requirements.
I cannot find the build error anymore, I must have deleted them while cleaning yesterday.

I checked a bit into the lock file, and from I can see, there is not wheel registeres for 0.33 for win os.
image

I will try to change againt, I will see if it's still broken

Hartorn: OK, I reproduced it in here : https://github.com/Giskard-AI/giskard/pull/1389
https://github.com/Giskard-AI/giskard/actions/runs/6120611136/job/16612786445?pr=1389

And I found the root cause : 

https://github.com/tensorflow/io/issues/1789

Basically, since 0.32, there is no win wheel on pipy, I will see this afternoon for the details.
0.31 should work though

PR #1394, opened 2023-09-08
[GSK-1711] Added check of push output

rabah-khalek:
additional test for push

### DIFF
--- a/python-client/tests/test_push.py
+++ b/python-client/tests/test_push.py
@@ -1,9 +1,11 @@
+import sys
+
 import numpy as np
+import pandas as pd
 import pytest
 
 import giskard.push
 from giskard.ml_worker.testing.functions.transformation import mad_transformation
-
 from giskard.ml_worker.testing.registry.giskard_test import GiskardTest
 from giskard.ml_worker.testing.registry.slicing_function import slicing_function
 from giskard.push import Push
@@ -17,7 +19,6 @@ from giskard.push.utils import (
     slice_bounds_quartile,
 )
 from giskard.slicing.slice import QueryBasedSliceFunction
-import pandas as pd
 
 DATASETS = [
     pytest.param(("german_credit_model", "german_credit_data", 50), id="German Credit"),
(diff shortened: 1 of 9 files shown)


python-client/tests/test_push.py
```diff
@@ -126,18 +128,35 @@ def test_slicing_function_multi(enron_model, enron_data):
             assert isinstance(push.slicing_function, QueryBasedSliceFunction)
 
 
-def test_test_function_multi(enron_model, enron_data):
+def test_test_function_and_output(enron_model, enron_data):
+
+    # Define a list of push types and corresponding functions
+    push_types = [
```
andreybavt, line 134: Since you're modifying this test anyway, let's take a chance to improve it:
instead of storing `push_types` list you can turn it into a parameterized test. This will:
- improve parallelization as each test will be able to executed separately
- debuggability: it'll be possible to execute a test for a given set of inputs and know which combination lead to an error

python-client/tests/test_push.py
```diff
@@ -126,18 +128,35 @@ def test_slicing_function_multi(enron_model, enron_data):
...
+        (giskard.push.PerturbationPush, create_perturbation_push),
+        (giskard.push.OverconfidencePush, create_overconfidence_push),
+        (giskard.push.BorderlinePush, create_borderline_push),
+    ]
+
+    # Initialize dictionaries to store outputs
+    outputs = {push_type: [] for push_type, _ in push_types}
+
+    # Loop through the range and create push outputs
     for i in range(50):
```
andreybavt, line 145: I'm not a big fan of blindly taking 50 first rows of enron. In this case we're not really sure what we're testing on. If we know which input leads to which push generation we should use it explicitly.
For example you can take row #5 that is known to generate 2 pushes and assert that both of them are generated and contain expected results.

python-client/tests/test_push.py
```diff
@@ -126,18 +128,35 @@ def test_slicing_function_multi(enron_model, enron_data):
...
+        row = enron_data.df.iloc[[i]]
+        for push_type, create_push_function in push_types:
+            push_output = create_push_function(enron_model, enron_data, row)
+            if push_output is not None:
+                outputs[push_type].append(push_output)
+
+    # Check if push output has the correct type
+    # Check if the tests created by push are instance of GiskardTest
+    for push_type, push_outputs in outputs.items():
+        if push_outputs:
```
andreybavt, line 155: related to the comment above, if we know that a specific row should result in a specific set of pushes we shouldn't use `if push_outputs`, but rather `assert len(push_outputs)>0`

andreybavt (dismissed):

rabah-khalek (commented):
I'm happy with your implementations @Hartorn! it's mergeable to me.

Hartorn (approved):

PR #1396, opened 2023-09-11
Trying different workflows

Hartorn:


.github/workflows/build_backend.yml
```diff
@@ -0,0 +1,232 @@
...
+name: Full CI
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      run-integ:
+        description: 'If integ test should be run'
```
andreybavt, line 13: Was "integration test" label too long for a GH UI?

.github/workflows/build_backend.yml
```diff
@@ -0,0 +1,232 @@
...
+jobs:
+  pre-check:
+    name: Pre check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Extract branch name
```
andreybavt, line 38: you don't use it anymore, do you?

.github/workflows/build_backend.yml
Hartorn: No, changing it

.github/workflows/build_backend.yml
Hartorn: I was, but I can managed without it

andreybavt:
I think if tests pass and `ingeg` is renamed to `integration` in various places it should be fine to merge

Hartorn:
> I think if tests pass and `integ` is renamed to `integration` in various places it should be fine to merge

Was sure I did it, I must have not pushed it before.

Hartorn:
@andreybavt Should be all good now

PR #1399, opened 2023-09-12
Fix/improve unreliable test

Hartorn:


python-client/tests/test_logging.py
```diff
@@ -7,4 +7,4 @@ def test_timer():
     timer = Timer()
     time.sleep(0.1)
     duration = timer.stop()
-    assert duration.microseconds / 10**6 >= 0.1
+    assert abs(duration.microseconds - 10**5) / 10**5 < 0.1
```
andreybavt, line 10: let's use a pytest utility here:

```
import pytest

def test_timer():
    sleep_sec = 0.1
    timer = Timer()
    time.sleep(sleep_sec)
    duration = timer.stop()
    assert duration.microseconds == pytest.approx(sleep_sec * 10**6, 0.10)
```

andreybavt (changes requested):

andreybavt (approved):

rabah-khalek:
@andreybavt I don't think this PR should've been merged before updating the branch with main. 

The main reason @Hartorn introduced these changes is because of the PR of single column that is now reverted.

PR #1401, opened 2023-09-14
[GSK-1745] Support both Pydantic 1 & 2

Hartorn:
## Description
Major issues encountered : 
- base settings is in a different package for pydantic v2 
  - handling it with extra deps was not working, since pdm was not installing it
  - since it's used only in one class, just doing the behaviour by hand
- validate_arguments was renamed validate_call in pydantic v2
- validate_call works only on types, so was creating issues to pickle code on SlicingFunction, TransformationFunction, and test decorator
- validate_call has been moved inside decorator

Everything else is only minor changes to enforce types and respect pydantic validation

### DIFF
--- a/python-client/giskard/llm/talk/talk.py
+++ b/python-client/giskard/llm/talk/talk.py
@@ -44,13 +44,17 @@ Question: {input}
 Thought: I should look at the features that are required to see what I should gather to predict in the available tools
 {agent_scratchpad}"""
 
+class LenientBaseToolkit(BaseToolkit):
+    """Extended class to allow arbitrary_types_allowed for pydantic compatibility"""
+    class Config:
+        arbitrary_types_allowed = True
+
 
 class ModelSpec(BaseModel):
     """Base class for model spec."""
-
     model: GiskardBaseModel
-    dataset: Optional[Dataset]
-    scan_report: Optional[ScanReport]
+    dataset: Optional[Dataset] = None
+    scan_report: Optional[ScanReport] = None
 
     def _parse_json_inputs(self, json_inputs: str) -> Dict[str, Any]:
         features = json.loads(json_inputs)
@@ -242,7 +246,7 @@ class ModelQualityTool(BaseTool):
         return self._run(tool_input)
 
 
-class ModelToolkit(BaseToolkit):
+class ModelToolkit(LenientBaseToolkit):
     """Toolkit for interacting with an ML model."""
 
     spec: ModelSpec
--- a/python-client/giskard/ml_worker/testing/registry/slicing_function.py
+++ b/python-client/giskard/ml_worker/testing/registry/slicing_function.py
@@ -60,7 +60,7 @@ class SlicingFunction(RegistryArtifact[DatasetProcessFunctionMeta]):
         """
         self.is_initialized = False
         self.params = {}
-        self.func = func
+        self.func = configured_validate_arguments(func)
         self.row_level = row_level
         self.cell_level = cell_level
 
@@ -147,7 +147,7 @@ def slicing_function(_fn=None, row_level=True, name=None, tags: Optional[List[st
         if inspect.isclass(func) and issubclass(func, SlicingFunction):
             return func
 
-        return _wrap_slicing_function(func, row_level, cell_level)()
+        return _wrap_slicing_function(func, row_level, cell_level)#()
 
     if callable(_fn):
         return functools.wraps(_fn)(inner(_fn))
@@ -166,4 +166,4 @@ def _wrap_slicing_function(original: Callable, row_level: bool, cell_level: bool
     make_all_optional_or_suite_input(slicing_fn)
     set_return_type(slicing_fn, SlicingFunction)
 
-    return configured_validate_arguments(slicing_fn)
+    return slicing_fn()
--- a/python-client/giskard/ml_worker/testing/registry/slicing_function.py
+++ b/python-client/giskard/ml_worker/testing/registry/slicing_function.py
@@ -147,7 +147,7 @@ def slicing_function(_fn=None, row_level=True, name=None, tags: Optional[List[st
         if inspect.isclass(func) and issubclass(func, SlicingFunction):
             return func
 
-        return _wrap_slicing_function(func, row_level, cell_level)#()
+        return _wrap_slicing_function(func, row_level, cell_level)
 
     if callable(_fn):
         return functools.wraps(_fn)(inner(_fn))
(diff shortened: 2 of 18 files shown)


python-client/giskard/ml_worker/testing/registry/slicing_function.py
```diff
@@ -147,7 +147,7 @@ def inner(func: Union[SlicingFunctionType, Type[SlicingFunction]]) -> SlicingFun
         if inspect.isclass(func) and issubclass(func, SlicingFunction):
             return func
 
-        return _wrap_slicing_function(func, row_level, cell_level)()
+        return _wrap_slicing_function(func, row_level, cell_level)#()
```
Inokinoki, line 150: Should we clean up this?

python-client/tests/llm/test_talk.py
```diff
@@ -106,11 +105,12 @@ def test_model_ask_description(german_credit_model):
     sys.version_info.minor >= 9, reason="The Langchain agent use the new ast module implemented in Python 3.8"
 )
 def test_model_create_llm_agent_invalid_version(german_credit_model, german_credit_test_data):
-    with pytest.raises(ValidationError):
+    with pytest.raises(Exception) as exc_info:
```
Inokinoki, line 108: I do not quietly understand this. The test is skipped in >=Python3.9 (only be run in Python3.8).
Why we still assert `"This tool relies on Python 3.9 or higher" in str(exc_info)`

python-client/tests/scan/test_performance_bias_detector.py
```diff
@@ -32,7 +32,8 @@ def test_performance_bias_detector_trims_large_dataset(german_credit_model, germ
         detector.run(german_credit_model, large_dataset)
     except ValueError:
         pass
-
+    except TypeError:
+        pass
```
Inokinoki, line 36: Why are they passed here? Are there some nuisances between pydantic 1&2 on this?

python-client/tests/scan/test_performance_bias_detector.py
```diff
@@ -44,6 +45,8 @@ def test_performance_bias_detector_trims_large_dataset(german_credit_model, germ
         detector.run(german_credit_model, normal_dataset)
     except ValueError:
         pass
+    except TypeError:
+        pass
```
Inokinoki, line 49: idem

Inokinoki (changes requested):

python-client/tests/llm/test_talk.py
Hartorn: Since this code can only be run ton 3.9+, the test make sure the user in 3.8 will have a clear warning regarding the usage of this code.

python-client/giskard/llm/talk/talk.py
```diff
@@ -44,13 +44,17 @@
...
+        arbitrary_types_allowed = True
+
 
 class ModelSpec(BaseModel):
     """Base class for model spec."""
-
     model: GiskardBaseModel
-    dataset: Optional[Dataset]
-    scan_report: Optional[ScanReport]
+    dataset: Optional[Dataset] = None
```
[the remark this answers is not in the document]
Hartorn: I think so yeah. I guess pydantic enforce each field has a given of default value, so for Optional you need to give None.

At least, this was breaking with validation error when running it without it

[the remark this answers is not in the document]
Hartorn: If we merge this quickly, you could rebase the other feature also (or the opposite)

That way, CI will tell us :)

python-client/giskard/ml_worker/testing/registry/slicing_function.py
Hartorn, line 150:
```suggestion
        return _wrap_slicing_function(func, row_level, cell_level)
```

python-client/giskard/models/base/model.py
```diff
@@ -128,7 +128,7 @@ def __init__(
         self.meta = ModelMeta(
             name=name if name is not None else self.__class__.__name__,
             model_type=model_type,
-            feature_names=list(feature_names) if feature_names else None,
+            feature_names=list(feature_names) if feature_names is not None else None,
```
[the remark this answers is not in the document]
Hartorn: It's not working in some cases, like if feature_names is a panda Index or iterable
Pandas will raise an error, saying truth value of an array is ambiguous.

On the why it breaks, I think the object given here may have been 'coerced' by pydantic, but not anymore

python-client/giskard/settings.py
```diff
@@ -1,37 +1,54 @@
...
 
-class Settings(BaseSettings):
+class Settings(BaseModel):
     home: str = "~/giskard-home"
     ws_port: int = 9000
     ws_path: str = "/ml-worker"
     host: str = "localhost"
     max_workers: int = 10
-    loglevel = "INFO"
+    max_send_message_length_mb: int = 1024
```
Hartorn, line 34:
```suggestion
```

python-client/giskard/settings.py
```diff
@@ -1,37 +1,54 @@
...
+    max_receive_message_length_mb: int = 1024
```
Hartorn, line 35:
```suggestion
```

python-client/giskard/settings.py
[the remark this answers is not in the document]
Hartorn: 👍

python-client/tests/scan/test_performance_bias_detector.py
Hartorn: So as I understood the test, the detector will fail anyway, the important stuff is that slice is called, et sample also.
In this case, a new error type was raised, so i'm ignoring it also.

I guess I could have changed exception type to Exception overall

python-client/tests/scan/test_performance_bias_detector.py
Hartorn: Same

python-client/giskard/ml_worker/testing/registry/slicing_function.py
Hartorn: Done :)

python-client/giskard/models/base/model.py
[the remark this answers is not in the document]
Inokinoki: Ok, that's much clear. Thanks

python-client/giskard/llm/talk/talk.py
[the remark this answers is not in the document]
Inokinoki: Let's merge!

Inokinoki (approved):

PR #1404, opened 2023-09-15
[GSK-1751] text push always giving None as the most contributing word

rabah-khalek:


python-client/giskard/push/contribution.py
```diff
@@ -103,9 +103,13 @@ def _create_text_contribution_push(
         ContributionPush: An object representing the contribution push for text features.
     """
     # Explain the text feature
+    input_df = sliced_ds.df
+    if model.meta.feature_names:
```
andreybavt, line 107: why not calling `model.prepare_dataframe` that already this logic?

rabah-khalek: You're right, thanks!

andreybavt (approved):

PR #1405, opened 2023-09-15
[GSK-1700] Issue when df has single column (version 2)

rabah-khalek:
A better alternative to the solution https://github.com/Giskard-AI/giskard/pull/1385 (reverted from main in https://github.com/Giskard-AI/giskard/commit/aad7800846745f6a7c56134d16c84b320a5ff3a0).

The issue was that if the dataset contains only one column, this code breaks:
```py
def prediction_function(df: pd.DataFrame) -> np.ndarray:
    return model.predict_proba(df)
```
even if the feature is indicated in the feature_names of the model. This is due to the fact that an sklearn model expects 1D data to be a pd.Series instead of pd.DataFrame.

In this PR, the attribute `accepts_only_pd_series` was introduced to `wrapper.py`, which is `False` by default, and gets modified inside `validate_model` if needed to become `True` after testing the model evaluation both on a `pd.DataFrame` and `pd.Series` in case the dataset is made out of 1 column.

The shortcoming: what if the user bypass the `scan` (thus, `validate_model`)?


python-client/giskard/core/model_validation.py
```diff
@@ -86,7 +86,14 @@ def validate_model_execution(model: BaseModel, dataset: Dataset, deterministic:
     try:
         prediction = model.predict(validation_ds)
     except Exception as e:
-        raise ValueError(error_message) from e
+        if "index 1 is out of bounds for axis 0 with size 1" in str(e) and len(validation_ds.df.columns) == 1:
```
Hartorn: I think it's not a good idea to depends on the message.

It would be better to try to handle it using the exception type (IndexError I think here)

Hartorn:
```python
   prediction = None
   try:
        prediction = model.predict(validation_ds)
    except IndexError:
       model.accepts_only_pd_series = True
    except Exception as e:
       raise ValueError(error_message) from e
    if prediction is None:
        try:
           prediction = model.predict(validation_ds)
        except Exception as e:
           raise ValueError(error_message) from e
```

python-client/giskard/models/base/wrapper.py
```diff
@@ -99,6 +102,12 @@ def __init__(
     def _preprocess(self, data):
         if self.data_preprocessing_function:
             return self.data_preprocessing_function(data)
+
+        # In the case where a dataframe has one column, most sklearn models require a Series instead of a DataFrame.
+        # It could also be the case of PredictionFunctionModel that calls sklearn models.
+        if self.accepts_only_pd_series:
+            return data[data.columns[0]]
```
Hartorn, line 109: Maybe should we check the column name still ?

Hartorn (changes requested):

python-client/giskard/core/model_validation.py
rabah-khalek: I wanted to capture not any `IndexError` but specifically this one `"index 1 is out of bounds for axis 0 with size 1"`

python-client/giskard/models/base/wrapper.py
Hartorn: Shouldn'it failed if you train with a df with columns a, y (y being the target) and then you try to predict with a colname 'y', or 'z' ?
I think your df can have more than 1 col, but the used col is only one, it should still be the proper one no ?

Hartorn: Or is it done before ?

python-client/giskard/core/model_validation.py
Hartorn: Hmm, ok I see. I think you should still match on exception type also, in addition to checking message.

Small detail, but you can avoid the else by inversing your if 
```python
if "index 1 is out of bounds for axis 0 with size 1" not in str(e) or number_of_features != 1:
    raise ValueError(error_message) from e

try:
    model.accepts_only_pd_series = True
    prediction = model.predict(validation_ds)
except Exception as e2:
    raise ValueError(error_message) from e2

````

File: python-client/giskard/models/base/wrapper.py (hunk 2 above)
Comment (rabah-khalek) [reply python-client/giskard/models/base/wrapper.py]: this one is taken care of ✅ Thanks!

File: python-client/giskard/core/model_validation.py (hunk 1 above)
Comment (rabah-khalek) [reply python-client/giskard/core/model_validation.py]: Thanks! done ✅

**Review by mattbit — COMMENTED** [review]
@rabah-khalek, this is not an issue for me.

If the sklearn model expects a 1-dimensional `ndarray` or `pd.Series` as input, then calling
```py
model.predict_proba(df)
```
should break.

It’s up to the user to define the prediction function properly. If there is a bug it should break. I feel like we risk producing unexpected behaviour elsewhere if we try to outsmart the user.
Also, `sklearn` models may expect a `pd.DataFrame`, or in general a 2-dimensional array (even if N×1). We could cause unexpected consequences if we squeeze the dimension of the input.

What I would be okay with is to give a hint to the user about what could be wrong if the prediction function fails.

mattbit (approved):
It‘s good for me. Maybe add a little comment in the code to explain for posterity, it’s a quite special case.

rabah-khalek:
> It‘s good for me. Maybe add a little comment in the code to explain for posterity, it’s a quite special case.

comment added. Thanks

PR #1409, opened 2023-09-19
[GSK-852] Change Tensorflow to latest version, to enable arm64 test a…

Hartorn:
Needs https://github.com/Giskard-AI/giskard/pull/1401 to be merged
## Description

- Detoxify has been removed and changed by the HuggingFace model, but the prediction do change a bit...
I'm not sure why it's changing so much, but I guess we are not using the same weights exactly.
Is it

Notes : 
Pydantic >2 cannot be installed with tensorflow >=2.13, because of incompatible version typing-extensions

https://github.com/tensorflow/tensorflow/issues/60687#issuecomment-1655280362
Should be fine in 2.14


python-client/setup.py
Inokinoki, line 1: Nice work!
After getting rid of protobuf (where we had a build task in this file), this is also removed.

Does this removal impact any pip installation from source code, such as `pip install git+<url>`?

.github/workflows/nightly-test.yml
```diff
@@ -9,4 +9,9 @@ jobs:
...
     with:
       run-integration-tests: true
-      use-cache: false
\ No newline at end of file
+      use-cache: false
+  call-build-images:
+    uses: ./.github/workflows/build-images.yml
+    with:
+      run-tests: false
+      build_only: true
```
Inokinoki, line 17: How about having a nightly image?

Inokinoki:
Review finished.

I will try to test on Linux and macOS, with py3.11

rabah-khalek (approved):

PR #1410, opened 2023-09-20
Do not run integration tests on windows

Hartorn:
Slight improvement to not run integration tests on windows
Also avoid re-running not slow tests

### DIFF
--- a/.github/workflows/build_backend.yml
+++ b/.github/workflows/build_backend.yml
@@ -221,8 +221,8 @@ jobs:
           path: python-client/dist/*whl
           
       - name: Run integration tests for python
-        if: ${{ inputs.run-integration-tests }}
+        if: ${{ inputs.run-integration-tests && matrix.os != 'windows-2019' }}
         working-directory: python-client
         env:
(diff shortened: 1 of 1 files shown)


mattbit:
@Hartorn, why?

Hartorn:
> @Hartorn, why?

@mattbit Well, on windows runner, the nightly fails, because the whole workflow is taking more than 360minutes.
This is also linked to the fact we cannot used XDIST with more thant 1 on windows runner, else they run out of memory.

mattbit:
> Well, on windows runner, the nightly fails, because the whole workflow is taking more than 360minutes.

Wait, is it the workflow on Windows that takes >360 min or you mean the combined time of testing on all platforms & python versions?

Hartorn:
> > Well, on windows runner, the nightly fails, because the whole workflow is taking more than 360minutes.
> 
> Wait, is it the workflow on Windows that takes >360 min or you mean the combined time of testing on all platforms & python versions?

The workflow on Windows, including all integrations test, not the combined time.

andreybavt (approved):
as sub-task to re-enable "slow" tests for windows after investigating why they can't be run with GH runner:

https://linear.app/giskard/issue/GSK-1613/enable-slow-tests-on-windows

PR #1412, opened 2023-09-20
[GSK-1715] Filling per-test debug descriptions

rabah-khalek:
The default message where there's no `debug_description` in the test decorator is:
```
This debugging session opens one by one all the examples that make the test fail.
```

All the per test `debug_description`s will start with `debug_description_prefix`: 
```
This debugging session opens one by one all the examples 
```


jmsquare (approved):
Great work. I like the formulations:
- For the Theil U, maybe it's more explicit if we put:
`debug_description_prefix + "of the data slice that was given as an input of the test"`

PR #1416, opened 2023-09-22
Fix clear cache workflow

Hartorn:


.github/workflows/clear-cache.yml
```diff
@@ -12,7 +12,11 @@ jobs:
     name: Clear all cache
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout code
```
andreybavt, line 15: do we need to checkout the repo in order to clear its cache?

Hartorn: Yeah, or to give remote name

rabah-khalek (approved):

PR #1418, opened 2023-09-22
fixed gender_stereotype_detector issue

rabah-khalek:
The issue was due to the nested f-string contained in the `_prompt_template` in `gender_stereotype_detector`.


mattbit (approved):
Could have escaped the job instead of substituting it with `[job]`, but fine This detector is going to be dropped soon in any case, so as far as this solves you problem I’m good with that.

PR #1423, opened 2023-09-25
Minor visual fixes for the push feature (GSK-1667)

Googleton:


backend/src/main/java/ai/giskard/service/InspectionService.java
```diff
@@ -167,6 +167,16 @@ private Selection getSelection(Inspection inspection, Filter filter) throws File
                     DoubleColumn absDiff = calculatedTable.doubleColumn("absDiff");
                     selection = correctSelection.and(absDiff.isLessThanOrEqualTo(applicationProperties.getBorderLineThreshold()));
                 }
+                case OVERCONFIDENCE -> {
+                    DoubleColumn overconfidenceScore = calculatedTable.doubleColumn("absDiff");
+                    DoubleColumn overconfidenceThreshold = DoubleColumn.create("overconfidenceThreshold", overconfidenceScore.size());
+                    for (int i = 0; i < overconfidenceScore.size(); i++) {
+                        int n = predsTable.columnCount();
```
Hartorn, line 174: If the threshold is the same value for every row, this should be computed outside of the loop.

frontend/src/components/PushPopover.vue
```diff
@@ -122,6 +122,7 @@ async function applyCta(kind: CallToActionKind) {
   let action: PushActionDTO = (await pushStore.applyPush(push.value!.kind, kind)).action;
   switch (kind) {
     case CallToActionKind.CreateSlice:
+      await catalogStore.loadCatalog(projectStore.currentProjectId ?? 0);
```
Hartorn, line 125: Should not have been call or raise an error if currentProjectId is not defined, no ?
Or redirect ?

frontend/src/components/PushPopover.vue
```diff
@@ -152,6 +153,8 @@ async function applyCta(kind: CallToActionKind) {
       break;
     case CallToActionKind.OpenDebuggerOverconfidence:
       // Programmatically apply Overconfidence filter
+      //debuggingStore.setSelectedFilter({value: RowFilterType.OVERCONFIDENCE, label: 'Overconfidence', disabled: false});
```
Hartorn, line 156: Shouldn't be this one activated ?

frontend/src/stores/push.ts
```diff
@@ -28,24 +30,31 @@ export const usePushStore = defineStore('push', {
...
+            this.loading = true;
             let result = await api.getPushes(modelId, datasetId, rowNb, inputData);
-
+            this.loading = false;
             if (previous == result) {
                 return;
             }
 
             let currentIdentifier = JSON.stringify(this.identifier);
             if (currentIdentifier === identifierString) {
```
Hartorn, line 57: how could this be false ? because of async ?

frontend/src/components/PushPopover.vue
```diff
@@ -85,13 +85,13 @@ const show = computed(() => {
 
   switch (props.type) {
     case "contribution":
-      return value.value.contribution?.kind !== PushKind.Invalid && value.value.contribution?.key == props.column;
+      return value.value.contribution && value.value.contribution?.kind !== PushKind.Invalid && value.value.contribution?.key == props.column;
```
Hartorn, line 88: Maybe check before switch something like 

```js
if (!value.value[props.type] || value.value[props.type].kind === PushKind.Invalid) {
return false;
}
```

Hartorn: It will allow lighter switch case like
```js
return value.value.perturbation?.key == props.column;
```

frontend/src/views/main/project/InspectorWrapper.vue
```diff
@@ -296,7 +301,7 @@ async function submitValueVariationFeedback(userData: object) {
 
 const debouncedUpdateRow = _.debounce(async () => {
   await updateRow(false);
-}, 150);
+}, 250);
```
Hartorn, line 304: Maybe could add `leading` option ?

```js
}, 250, { leading: true});
```

frontend/src/views/main/project/InspectorWrapper.vue
```diff
@@ -254,9 +258,10 @@ function next() {
 }
 
 watch(() => inputData.value, async (newValue, oldValue) => {
-  console.log("RowNb changed");
-  const pushStore = usePushStore();
-  await pushStore.fetchPushSuggestions(inspection.value!.model.id, inspection.value!.dataset.id, rowNb.value, newValue, modelFeatures.value);
+  if (newValue !== undefined) {
+    const pushStore = usePushStore();
+    await pushStore.fetchPushSuggestions(inspection.value!.model.id, inspection.value!.dataset.id, rowNb.value, newValue, modelFeatures.value);
```
Hartorn, line 263: what does !. do ?

Hartorn (changes requested):

frontend/src/components/PushPopover.vue
Googleton: It is sorry forgot to push 🙏

frontend/src/views/main/project/InspectorWrapper.vue
Googleton: It's a "non null" operator. It'll check if the previous variable is null, and if so, return false directly. It allows chaining by doing

`if (value!.othervalue!.lastvalue)` in a single line rather than 3 separate checks

frontend/src/views/main/project/InspectorWrapper.vue
Googleton: The issue is that for long running models we want to avoid hitting them at all if the user starts fast navigation

frontend/src/views/main/project/InspectorWrapper.vue
Hartorn: 👍

PR #1447, opened 2023-10-02
Clean up everything not related to python

Hartorn:
## Description

- Removal of almost every files out of python-client folder (first commit)
- Moving file from python-client to root folder (second commit)
- Adapting build for new architecture


.prettierrc
mattbit, line 1: maybe we should keep this?

kevinmessiaen (approved):

mattbit (approved):
Ok for me!

.prettierrc
mattbit: actually not that important after all

PR #1472, opened 2023-10-10
Make worker action cancellable

Hartorn:
Closes #GSK-1863

## Type of Change

- [ ] 🔐 Security fix

## Checklist

### DIFF
--- a/giskard/ml_worker/websocket/listener.py
+++ b/giskard/ml_worker/websocket/listener.py
@@ -9,7 +9,7 @@ import sys
 import tempfile
 import time
 import traceback
-from concurrent.futures import Future
+from concurrent.futures import CancelledError, Future
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -69,7 +69,7 @@ class MLWorkerInfo:
         self.is_remote = worker.is_remote_worker()
 
 
-def websocket_log_actor(ml_worker: MLWorkerInfo, req: dict, *args, **kwargs):
+def websocket_log_actor(ml_worker: MLWorkerInfo, req: Dict, *args, **kwargs):
     param = req["param"] if "param" in req.keys() else {}
     action = req["action"] if "action" in req.keys() else ""
     logger.info(f"ML Worker {ml_worker.id} performing {action} params: {param}")
@@ -86,6 +86,9 @@ def wrapped_handle_result(action: MLWorkerAction, ml_worker: MLWorker, start: fl
 
         try:
             info: websocket.WorkerReply = future.result() if isinstance(future, Future) else future()
+        except CancelledError:
+            info: websocket.WorkerReply = websocket.Empty()
+            logger.warning("Task for %s has timed out and been cancelled", action.name)
         except Exception as e:
             info: websocket.WorkerReply = websocket.ErrorReply(
                 error_str=str(e), error_type=type(e).__name__, detail=traceback.format_exc()
@@ -168,7 +171,7 @@ def parse_and_execute(
     )
 
 
-def dispatch_action(callback, ml_worker, action, req, execute_in_pool):
+def dispatch_action(callback, ml_worker, action, req, execute_in_pool, timeout=None):
     # Parse the response ID
     rep_id = req["id"] if "id" in req.keys() else None
     # Parse the param
@@ -199,7 +202,7 @@ def dispatch_action(callback, ml_worker, action, req, execute_in_pool):
     result_handler = wrapped_handle_result(action, ml_worker, start, rep_id)
     # If execution should be done in a pool
     if execute_in_pool:
-        logger.debug("Submitting for action %s '%s' into the pool with %s", action.name, callback.__name__, params)
+        logger.debug("Submitting for action %s '%s' into the pool", action.name, callback.__name__)
         future = call_in_pool(
             parse_and_execute,
             callback=callback,
@@ -207,6 +210,7 @@ def dispatch_action(callback, ml_worker, action, req, execute_in_pool):
             params=params,
             ml_worker=MLWorkerInfo(ml_worker),
             client_params=client_params,
+            timeout=timeout,
         )
         future.add_done_callback(result_handler)
         log_pool_stats()
@@ -223,7 +227,7 @@ def dispatch_action(callback, ml_worker, action, req, execute_in_pool):
     )
 
 
-def websocket_actor(action: MLWorkerAction, execute_in_pool: bool = True):
+def websocket_actor(action: MLWorkerAction, execute_in_pool: bool = True, timeout: Optional[float] = None):
     """
     Register a function as an actor to an action from WebSocket connection
     """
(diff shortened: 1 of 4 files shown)


giskard/ml_worker/websocket/listener.py
```diff
@@ -86,6 +86,9 @@ def handle_result(future: Future):
 
         try:
             info: websocket.WorkerReply = future.result()
+        except CancelledError:
+            info: websocket.WorkerReply = websocket.Empty()
```
Inokinoki, line 90: According to @kevinmessiaen , the empty response could cause problem on Java side. Could it be better to merge https://github.com/Giskard-AI/giskard-hub/pull/134 first?

giskard/utils/__init__.py
```diff
@@ -21,19 +23,23 @@ class WorkerPool:
 
     def __init__(self):
         self.pool = None
+        self.nb_cancellable = 0
+        self.max_workers = 0
 
-    def start(self, *args, **kwargs):
+    def start(self, max_workers: int = None):
```
Inokinoki, line 29: I forgot the codestyle stuff, should it be something like `max_workers: int=None`?

giskard/ml_worker/websocket/listener.py
```diff
@@ -203,14 +206,15 @@ def dispatch_action(callback, ml_worker, action, req):
...
+        timeout=timeout,
     )
     future.add_done_callback(wrapped_handle_result(action, ml_worker, start, rep_id))
     log_pool_stats()
 
     return
 
 
-def websocket_actor(action: MLWorkerAction):
+def websocket_actor(action: MLWorkerAction, timeout: Optional[float] = None):
```
Inokinoki, line 217: Also the codestyle

giskard/ml_worker/websocket/listener.py
Hartorn: Sure about the PR.
It was looking fine from what I saw when testing. What should I send back ?

giskard/utils/__init__.py
Hartorn: I did the formatting with black, so should be fine as is

giskard/ml_worker/websocket/listener.py
Inokinoki: I think it's fine with @kevinmessiaen 's PR.
At the moment the empty response is just ignored. I guess it's OK with the push feature because there is no transaction.
Maybe we can add a comment here to remind people to handle the empty response when adding a transactional action.

kevinmessiaen: @Inokinoki Yes it's okay to return an `Empty` message in that case, the issue was if an error arose it wasn't intercepted because we didn't read the response DTO returned by the worker in the backend.

It caused issue as we didn't rolled back transactions when error wasn't intercepted properly. But there since we cancel  an action it's not an issue (and the issue will be fixed when the PR is merged anyway)

giskard/utils/__init__.py
```diff
@@ -21,19 +23,23 @@ class WorkerPool:
...
         if self.pool is not None:
             return
         LOGGER.info("Starting worker pool...")
-        self.pool = ProcessPoolExecutor(*args, **kwargs)
+        self.max_workers = min(max_workers, 2) if max_workers is not None else os.cpu_count()
```
andreybavt, line 33: small improvement: it'd be better to move `2` to the `giskard.settings.Settings`. In this case we'd get a possibility to change it from an env variable

giskard/utils/__init__.py
```diff
@@ -109,6 +104,64 @@ def wrapper(*args, **kwargs):
...
+
+    Returns:
+        Future: the promise of the results
+    """
+    if timeout is None:
+        return POOL.submit(fn, *args, **kwargs)
+    # Create independant process pool
+    # If we kill a running process, it breaking the Process pool, making it unusable
+    one_shot_executor = ProcessPoolExecutor(max_workers=1)
+    pid = one_shot_executor.submit(os.getpid).result(timeout=5)
```
andreybavt, line 157: I don't get it, the comment above said 
> .result(timeout=timeout) => Not working with WSL and python 3.10

here it's expected to work?

Hartorn: I should remove the timeout here anyway, no reason for basic os.getpid to fail or take a long time.

Basically, when I tested it, in the killer thread, the timeout was not respected, and it never stopped, so I had to do the loop.

Btw, the timeout was working out on MacOS with python 3.11.

giskard/utils/__init__.py
Hartorn: Done. I also changed `min` to `max` also, since we always want at least `min_workers`

giskard/utils/__init__.py
andreybavt: I think the overall killing mechanism could be simplified to avoid having a killer thread and a cancellable counter.

We could use an inter-process data structure to communicate pids between a spawned worker process and the main process: `multiprocessing.Queue` for example.

In this case the first thing the pool process will do is identify its pid and add it to queue, then do the actual work.
In the main process we could use `concurrent.futures._base.as_completed` to wrap the future and add a timeout to it and then call `.add_done_callback(result_handler)` as we do now.

When the timeout is expired we'd find a PID related to a given task and kill it from the main process.

WDYT?

Hartorn: > @Hartorn , actually, it looks like https://pypi.org/project/Pebble/ does exactly what we need:
> 
> ```
> from pebble import ProcessPool
> 
> pool = ProcessPool(max_workers=5)
> 
> pool.schedule(long_fn, args=(a,b,c), timeout=1).add_done_callback(done_callback)
> ```

Cancellable counter is only for logs, we can remove it if we want. 
I just wanted to check if that was working, or eventually to avoid having too many process launched.
For the callback of a future, you need everything to be pickable, so you cannot have lock, processPool, and so on.
So I don't know how to shutdown properly the one shot executor we ran.

Also, it cannot be the main executor, since killing a process is breaking the pool, which only raises BrokenPoolException after that.

I'm pretty sure we could avoid the thread is the whole code we were running was async, since a coroutine could do this job, but here I would not be confident.

Hartorn: > @Hartorn , actually, it looks like https://pypi.org/project/Pebble/ does exactly what we need:
> 
> ```
> from pebble import ProcessPool
> 
> pool = ProcessPool(max_workers=5)
> 
> pool.schedule(long_fn, args=(a,b,c), timeout=1).add_done_callback(done_callback)
> ```

Looking at it !

Hartorn: From what I saw, they are kinda doing the same stuff, when using a separate process, they have a handler thread for watching it and handling timeout.

Should we get this merged and change to use it ? Or want me to switch it ?

Although I'm a bit concerned it's not that much used, the code looks clean

andreybavt: Actually, while having a clean API and being easy to use Pebble is a LGPL library so we won't be able to use it (it might be the reason why it's not that widely adopted). 

I also read their code and found similarities with your implementation. I suggest we stick to your current code  (and merge it since we need these changes ASAP).

As an improvement, I think in a separate PR we could inspire from Pebble's API and also encapsulate `call_in_pool` into our custom ProcessPool implementation and have a `schedule` method, WDYT?
In the end Pebble is just a ~5k LOC library

Hartorn: Shame, the code was working well with it also.
Sure, seems a good idea to have our own worker implementation. I'm sure it will also help us to improve performance of the lib

PR #1478, opened 2023-10-13
GSK-1895 New worker pool

Hartorn:
## Description

Re-implement a worker pool, with possibility for timeout 

Also, had capacity to capture all loging for executed task

I removed python deps and related code, because it's was a bother for me under windows for testing.

## Type of Change

- [ ] 🔐 Security fix

## Checklist


giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+                pass
+            continue
+        future.logs = result.logs
+        if result.exception is None:
+            future.set_result(result.result)
+        else:
+            # TODO(Bazire): improve to get Traceback
+            future.set_exception(RuntimeError(result.exception))
+        try:
+            del executor._futures_mapping[result.id]
```
Inokinoki, line 264: Could this help get rid of the unusable executor error when getting exception from subprocess (executor)?
I faced a few such errors when testing with GPU (errors are raised due to the different devices of the tensors).

giskard/ml_worker/testing/registry/udf_repository.py
Inokinoki, line 1: What did it do before and why this is removed?

giskard/utils/worker_pool.py
Hartorn: Here exception a string with the error + the full traceback of the exception normally
Most exception are not really pickable, and so I serialized it.

What were you missing as information in your case ?

giskard/ml_worker/testing/registry/udf_repository.py
Hartorn: It is not used (git inner repository to handle code and so on)

I had some issue with it when running on windows, with git not configured or installed, so I removed it alltogether

giskard/utils/worker_pool.py
Inokinoki: I remembered it was something like:

```
concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```

Maybe not exactly the same exception, but it happens when the process is terminated due to an internal exception

giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+            sleep(0.01)
+        if executor._state in FINAL_STATES:
+            return
+        result = executor._tasks_results.get()
+        future = executor._futures_mapping.get(result.id)
+        if future.cancelled():
+            try:
+                del executor._futures_mapping[result.id]
+            except BaseException:
+                pass
```
andreybavt, line 255: we should log this unexpected exception

giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+            continue
+        future.logs = result.logs
+        if result.exception is None:
+            future.set_result(result.result)
+        else:
+            # TODO(Bazire): improve to get Traceback
+            future.set_exception(RuntimeError(result.exception))
```
andreybavt, line 262: why can't it by directly `future.set_exception(result.exception)` ?

andreybavt (changes requested):
A couple of general comments:

- the queue names are prefixed with `_` but being referenced from outside of executor, for example: `executor._pending_tasks_queue`. I think it's ok if they become public, at least we're using them this way anyway
- There are still untested corner cases, like "Future has been cancelled already". It'll be great to increase code coverage even more considering that this is a core part of the ML Worker

giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+    STARTED = 1
+    BROKEN = 2
+    STOPPING = 3
+    STOPPED = 4
+
+
+FINAL_STATES = [PoolState.STOPPING, PoolState.STOPPED, PoolState.BROKEN]
+
+
+class WorkerPoolExecutor(Executor):
```
andreybavt, line 121: I think we could also give pools optional names in case we decide to reuse it for other purposes. This name will be passed further to thread names as a prefix for example

giskard/utils/worker_pool.py
andreybavt: In any case there's no standard process pool anymore, so indeed this exception won't happen

andreybavt:
also let's get Sonar code smells fixed too

giskard/utils/worker_pool.py
Hartorn: Exception is a string in this case (the exception + traceback), because it needs to be pickable, and exception are linked to many objects + have circular references

giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+    def __init__(self, nb_workers: Optional[int] = None):
+        if nb_workers is None:
+            nb_workers = cpu_count()
```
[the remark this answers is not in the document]
Hartorn: I think it's done when starting, but will check up

giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+        self._state = PoolState.STOPPING
+        # Cancelling all futures we have
+        for future in self._futures_mapping.values():
+            if future.cancel() and not future.done():
+                future.set_exception(CancelledError("Executor is stopping"))
+        # Emptying running_tasks queue
+        while not self._running_tasks_queue.empty():
+            try:
+                self._running_tasks_queue.get_nowait()
+            except BaseException:
```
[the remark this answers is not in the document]
Hartorn: Yeah, probably. Plus I should reduce the exeception scope

giskard/utils/worker_pool.py
Hartorn: + reduce the scope

giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+        future.logs = result.logs
+        if result.exception is None:
+            future.set_result(result.result)
+        else:
+            # TODO(Bazire): improve to get Traceback
+            future.set_exception(RuntimeError(result.exception))
+        try:
+            del executor._futures_mapping[result.id]
+        except BaseException:
+            pass
```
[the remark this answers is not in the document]
Hartorn: Yup, I will add some warning logs

giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+            # TODO(Bazire): find a way to improve this ?
+            # Cannot use select, since we want to be windows compatible
+            sleep(0.01)
+        if executor._state in FINAL_STATES:
+            return
+        task = executor._pending_tasks_queue.get()
+        future = executor._futures_mapping.get(task.id)
+        if future is not None and future.set_running_or_notify_cancel():
+            executor._running_tasks_queue.put(task)
+        elif future is not None:
```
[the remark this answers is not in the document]
Hartorn: Will do

giskard/utils/worker_pool.py
```diff
@@ -0,0 +1,337 @@
...
+                        executor._spawn_worker()
+                except BaseException as e:
+                    LOGGER.warning("Unexpected error when killing a timed out process, pool is broken")
+                    LOGGER.exception(e)
+                    exception = e
+                    executor._state = PoolState.BROKEN
+                finally:
+                    clean_up.append(timeout_data)
+        for elt in clean_up:
+            executor._with_timeout_tasks.remove(elt)
```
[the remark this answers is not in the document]
Hartorn: Because it's the same list i'm iterating on, so it will raise an exception if I do that. Or I need ot iterate on a copy

Hartorn (commented):
I took in account all comment.
Also, I improved the threads, especially the killer one.

I added a quich health_check too

andreybavt (approved):

PR #1481, opened 2023-10-17
[GSK-1900] Allow to move PyTorch model from HF to the given device

Inokinoki:
## Description

TF model could also be PyTorch NN module. We need use `to(device)` to move and run the models to GPU.

Can be tested using HF DistilBert model.

## Related Issue

[GSK-1900] ML worker cannot predict on GPU with model from Hugging Face

## Type of Change

- [ ] 🔐 Security fix

## Checklist

### DIFF
--- a/giskard/models/huggingface.py
+++ b/giskard/models/huggingface.py
@@ -133,6 +133,7 @@ class HuggingFaceModel(WrapperModel):
         classification_labels: Optional[Iterable] = None,
         id: Optional[str] = None,
         batch_size: Optional[int] = 1,
+        device: Optional[str] = None,
         **kwargs,
     ) -> None:
         """Automatically wraps a HuggingFace model or pipeline.
@@ -184,6 +185,7 @@ class HuggingFaceModel(WrapperModel):
 
         self.huggingface_module = model.__class__
         self.pipeline_task = model.task if isinstance(model, pipelines.Pipeline) else None
+        self.device = device if device else model.device.type   # Allow to overwrite device
 
         try:
             if batch_size == 1 and model.device.type == "cuda":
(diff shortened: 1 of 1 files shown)


rabah-khalek (changes requested):
it's not a bad idea to have a `device` field for `HuggingFaceModel` but this needs to be done properly taking into account two things:
- `device` has to be saved in the meta files (see how it's done for `PyTorchModel`)
- the HF model could be a tensorflow one, so we should first check if `hasattr(model, "device")`

giskard/models/huggingface.py
```diff
@@ -184,6 +185,9 @@ def __init__(
 
         self.huggingface_module = model.__class__
         self.pipeline_task = model.task if isinstance(model, pipelines.Pipeline) else None
+        self.device = device if device else model.device.type   # Allow to overwrite device
+        if not self.device:
+            self.deivce = "cpu"
```
mattbit, line 190: Typo `deivce`

giskard/models/huggingface.py
```diff
@@ -226,6 +231,10 @@ def save_model(self, local_path):
         self.model.save_pretrained(local_path)
 
     def model_predict(self, data):
+        if isinstance(self.model, torch.nn.Module):
+            # Move model to the given device
+            self.model = self.model.to(self.device)
```
mattbit, line 236: - What if the device is not available?
- What if the model input is not on the same device?

mattbit (changes requested):

mattbit:
I would be careful with this. Saving the device in the metadata for me does not necessarily make sense, since it’s something that depends on the environment.
We should thoroughly test that we are not breaking normal usage.
E.g. I upload the model from my notebook, running on GPU. Then I download it somewhere else, e.g. in CICD where I don’t have a GPU. Would it break? Would it fallback to CPU? To be checked.

Also: the preprocessing could be specific to the device? If I move the tensors around in the preprocessing function, would that break model_predict? Is there a fallback?

rabah-khalek:
yeah but you'd need to change your `prediction_function` too if you do such transitions of env. It's not like you're protected if `device` is not there. You'd have to create a new model either way.

Inokinoki:
After the discussion with @rabah-khalek I think that providing a special prediction function is enough for those who are using GPU. So close this

PR #1502, opened 2023-10-24
GSK-1963: ignore mixpanel errors

andreybavt:


giskard/utils/analytics_collector.py
```diff
@@ -149,6 +146,7 @@ def track(self, event_name, properties=None, meta=None, force=False):
         return self._track(event_name, properties=properties, meta=meta, force=force)
 
     @threaded
+    @analytics_method
```
mattbit, line 149: If I remember correctly, this was removed to avoid infinite loops. Can we add a test to ensure that this is not happening  (since `_report_error` has been removed in the decorator, this should be the case) and make sure we will not reintroduce this in the future?

EDIT: maybe the test that is present already covers this case?

mattbit (approved):

tests/communications/test_analytics_collector.py
```diff
@@ -0,0 +1,9 @@
+from mixpanel import Consumer
+
+from utils.analytics_collector import GiskardAnalyticsCollector
```
mattbit, line 3:
```suggestion
from tests.utils.analytics_collector import GiskardAnalyticsCollector
```

giskard/utils/analytics_collector.py
andreybavt: it was actually the presence of 
```
 _report_error(e, error_type="tracking error")
```
that was causing an infinite loop

andreybavt:
@mattbit the test should be fine now. It was caused by the way we run tests from pdm (by disabling analytics). That's why I wasn't having it locally (by running pytest directly)

PR #1505, opened 2023-10-25
[GSK-1891][GSK-1967] W&B related updates

rabah-khalek:
- Removing the internal `run` initiation (including the feedback I got from Thomas).
- Improving the docs + notebook (including the feedback I got from Thomas).
- Adding LLM notebook to the doc


giskard/integrations/wandb/wandb_utils.py
```diff
@@ -9,16 +8,19 @@
...
-    yield wandb.run
+def get_wandb_run(run):
+    if run is None:
+        if wandb.run is None:
+            raise ValueError("There are currently no active wandb runs available. Please follow the following "
+                             "documentation: https://docs.wandb.ai/ref/python/run to initiate a run. Once initiated, "
+                             "you can choose to pass is as follows 'to_wandb(run)', if not we will retrieve internally "
+                             "the 'wandb.run'.")
+        run = wandb.run
+    return run
```
mattbit, line 23: Can we simplify this?
```py
if run is not None:
    return run  # should we do other checks here?

if wandb.run is not None:
    return wandb.run

raise ValueError(...)
```

Also, should we use signature `def get_wandb_run(run=None)`?

giskard/datasets/base/__init__.py
```diff
@@ -681,36 +681,34 @@ def to_mlflow(self, mlflow_client: MlflowClient = None, mlflow_run_id: str = Non
 
         return artifact_name
 
-    def to_wandb(self, **kwargs) -> None:
+    def to_wandb(self, run=None) -> None:
```
mattbit, line 684: woudl be nice to have a type hint if possible

giskard/models/shap_result.py
```diff
@@ -60,7 +60,7 @@ def _validate_wandb_config(self) -> None:
                 "classification models."
             )
 
-    def to_wandb(self, **kwargs) -> None:
+    def to_wandb(self, run) -> None:
```
mattbit, line 63: same as above

giskard/scanner/report.py
```diff
@@ -178,48 +178,49 @@ def to_mlflow(
...
+            )
+        except Exception as e:
+            analytics.track(
+                "wandb_integration:scan_result:error:unknown",
+                {
+                    "wandb_run_id": run.id,
+                    "error": str(e),
+                },
+            )
+            raise ValueError(
```
mattbit, line 220: This is probably a `RuntimeError`?

tests/integrations/test_wandb.py
```diff
@@ -64,19 +63,34 @@ def test_slow(dataset_name, model_name, request):
...
     test_suite_results = scan_results.generate_test_suite().run()
-    test_suite_results.to_wandb()
+    test_suite_results.to_wandb(run)
 
     # Verify that the logging of the SHAP explanation charts works.
     explanation_results = explain_with_shap(model, dataset)
-    explanation_results.to_wandb()
+    explanation_results.to_wandb(run)
+
+    assert re.match("^[0-9a-z]{8}$", str(run.id))
```
mattbit, line 83: Why do you check the format of `run.id`? This is set by `wandb`, right?

tests/integrations/test_wandb.py
```diff
@@ -64,19 +63,34 @@ def test_slow(dataset_name, model_name, request):
...
+    "dataset_name", ["german_credit_data"],)
+def test_error(dataset_name, request):
+    dataset = request.getfixturevalue(dataset_name)
 
-    assert re.match("^[0-9a-z]{8}$", str(wandb.run.id))
+    with pytest.raises(ValueError) as e:
+        dataset.to_wandb()
+    assert e.match(
+        r"There are currently no active wandb runs available.*"
+    )
```
mattbit, line 96: you can also do `pytest.raises(ValueError, match="There are ...")`

mattbit (commented):
Few minor comments

luca-martial (approved):
Docs LGTM

tests/integrations/test_wandb.py
mattbit: by the way, no need to add ".*" because this uses `re.search`. Thus the expression will match any portion of the string.

tests/integrations/test_wandb.py
rabah-khalek: I don't remember to be honest. Yes It's a bit of a moot check.

mattbit (approved):
Missing some `Optional[...]` in the type hints but otherwise good for me! Well done @rabah-khalek

PR #1512, opened 2023-10-26
GSK-1911 Improve performance of the push feature & make CTA almost instant

Googleton:
## Description

Includes a "SimpleCache" class that uses hashlib to cache anything we want inside it. It only stores the last 20 results. 

This greatly improves the usability of the insights feature by:

1. Due to being the last-seen example, all calls to CTA will be cached making them much faster
2. Since we store a few examples, if I'm navigating between a few examples to do comparisons, we always hit the cache and end up with ultra fast results

It also includes a small rewording that JM asked me to do, because it's only text changes I did not want to create yet another PR for it.


giskard/ml_worker/utils/cache.py
```diff
@@ -0,0 +1,28 @@
...
+
+class SimpleCache:
+    def __init__(self, max_results=20):
+        self.results = {}
+        self.order = []  # To maintain the order in which results were added
+        self.max_results = max_results
+
+    def add_result(self, obj, result):
+        # Calculate the hash of the object
+        obj_hash = hashlib.md5(repr(obj).encode()).hexdigest()
```
Hartorn, line 12: Can you add encoding explicitely to encode pls ?

giskard/ml_worker/utils/cache.py
```diff
@@ -0,0 +1,28 @@
+import hashlib
+
+
+class SimpleCache:
+    def __init__(self, max_results=20):
+        self.results = {}
```
Hartorn, line 6: Could you add typing hint here, for better readability ?

giskard/ml_worker/utils/cache.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def add_result(self, obj, result):
+        # Calculate the hash of the object
+        obj_hash = hashlib.md5(repr(obj).encode()).hexdigest()
+
+        # Store the result with the object's hash as the key
+        self.results[obj_hash] = result
+
+        # Add the object's hash to the order list and remove the oldest result if necessary
+        self.order.append(obj_hash)
```
Hartorn, line 18: I think you should use LRU (least recently used) algorithm here, instead of oldest.

giskard/ml_worker/websocket/listener.py
```diff
@@ -54,11 +55,9 @@
 from giskard.utils.analytics_collector import analytics
 
 logger = logging.getLogger(__name__)
-
-
+push_cache = SimpleCache(max_results=20)
```
Hartorn, line 58: Could you write it as PUSH_CACHE instead ?

giskard/ml_worker/websocket/listener.py
```diff
@@ -681,46 +680,13 @@ def get_push(
...
-
-    from giskard.push.contribution import create_contribution_push
-    from giskard.push.perturbation import create_perturbation_push
-    from giskard.push.prediction import create_borderline_push, create_overconfidence_push
-
-    contribs = create_contribution_push(model, dataset, df)
-    perturbs = create_perturbation_push(model, dataset, df)
-    overconf = create_overconfidence_push(model, dataset, df)
-    borderl = create_borderline_push(model, dataset, df)
+    cached_result_tuple = push_cache.get_result(params)
```
Hartorn, line 684: Why not using it on the overall get_push instead ?

giskard/ml_worker/websocket/listener.py
Hartorn: Also, 20 is not enough I think. Maybe 128 ?

Hartorn (changes requested):
You need to write some unit tests on you new class.
Also, I would prefer the caching to be done with a parametrize decorator, to be easily re-usable.
I think you can add this behaviour to websocket_actor, with a new parameter "cache" defaulting to false, using params as the object to hash

Hartorn:
> You need to write some unit tests on you new class. Also, I would prefer the caching to be done with a parametrize decorator, to be easily re-usable. I think you can add this behaviour to websocket_actor, with a new parameter "cache" defaulting to false, using params as the object to hash

And also, please try to use type hinting for variable and methods

PR #1486, opened 2023-10-20
[GSK-1383] Make dataset and model as suite input when generating the test suite from scan

Inokinoki:
## Description

`ScanReport` can receive `use_suite_input` as a param in `generate_test_suite`, which is `True` by default.

When set to `True`, it will overwrite the `model` and `dataset` args in the test function with `ModelInput` and `DatasetInput` placeholders, which are subclasses of `SuiteInput`. These two args need to be provided when the test or the test suites are executed.

When set to `False`, it will behave as before: `model` and `dataset` are set from the `Issue`.

## Related Issue

[GSK-1383] Make dataset and model as suite input, when generating the test suite from scan

## Type of Change

- [ ] 📚 Examples / docs / tutorials / dependencies update
- [x] 🔧 Bug fix (non-breaking change which fixes an issue)
- [x] 🥂 Improvement (non-breaking change which improves an existing feature)
- [ ] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🔐 Security fix

## Checklist

- [ ] I've read the [`CODE_OF_CONDUCT.md`](https://github.com/Giskard-AI/ai-inspector/blob/master/CODE_OF_CONDUCT.md) document.
- [ ] I've read the [`CONTRIBUTING.md`](https://github.com/Giskard-AI/ai-inspector/blob/master/CONTRIBUTING.md) guide.
- [ ] I've updated the code style using `make codestyle`.
- [ ] I've written tests for all new methods and classes that I created.
- [ ] I've written the docstring in Google format for all the methods and classes that I used.

PR #1522, opened 2023-10-30
LLM tests refactoring

mattbit:
- Clean up duplicate LLM tests
- Add clear documentation to each test
- Improve naming of the generated datasets


mattbit:
Depends on #1513

giskard/utils/display.py
```diff
@@ -9,3 +9,10 @@ def format_number(value, n=3):
         return f"{value:.{n}e}"
 
     return value
+
+
+def truncate(text, max_length=240, ellipsis="…"):
+    """Truncates a text to the given length, adding an ellipsis at the end if needed."""
+    if len(text) > max_length:
+        return text[: max_length - len(ellipsis)] + ellipsis
+    return text
```
rabah-khalek, line 18: Potential issues?
- multi-byte characters 
- `len(ellipsis) > max_length`
- `max_length < 0`

giskard/scanner/llm/llm_chars_injection_detector.py
```diff
@@ -101,7 +101,6 @@ def _generate_char_injection_tests(issue: Issue):
     feature = issue.features[0]
     return {
         f"Character injection ({issue.meta['special_char'].encode('unicode_escape').decode('ascii')}) in “{feature}”": test_llm_char_injection(
-            model=issue.model,
```
rabah-khalek, line 104: we don't want to fix the model? Is that also what we do in the traditional scan when we generate tests?

giskard/testing/tests/llm/output_requirements.py
```diff
@@ -1,249 +1,122 @@
...
+    return TestResult(
+        passed=eval_result.passed, output_df=output_ds, metric=len(eval_result.success_examples) / len(dataset)
+    )
 
 
 @test(
-    name="Validate LLM evaluation dataset using GPT-4",
-    tags=["llm", "GPT-4"],
+    name="Evaluation of model output using an LLM (LLM-as-a-judge)",
+    tags=["llm", "llm-as-a-judge"],
```
rabah-khalek, line 36: I know it's configurable/dynamic, but is there a quick/easy way to tell the user what underlying LLM is used as a judge?

Also we should probably add the options that the user can choose to evaluate, I think only support currently chatGPT based ones

rabah-khalek (approved):
looks good to me, minor comments

rabah-khalek:
you're missing the check of the `debug` bool, It should be `False` by default, and checked inside the test. If it's true, the `ouput_ds` should be filled

giskard/utils/display.py
mattbit: I wouldn’t really care, for something like this which is for internal usage I assume responsible usage. I put the default to 240 chars because I don’t know how the servers handles multi-byte, in principle it’s a varchar(255) but who knows the encoding…

giskard/scanner/llm/llm_chars_injection_detector.py
mattbit: After #1486 the default way is to let the scan set them at the test suite level. So that you can benchmark a different model on the same test suite (in particular from the Hub):

```py
report = gsk.scan(model)

suite = report.generate_test_suite()

suite.run(my_other_model)
```

giskard/utils/display.py
rabah-khalek: ok

giskard/testing/tests/llm/output_requirements.py
rabah-khalek: Okay, I thought it was still as Kevin had it in his POC. Clear. Thanks.

mattbit:
> you're missing the check of the `debug` bool, It should be `False` by default, and checked inside the test. If it's true, the `output_ds` should be filled

Since I have the dataset, I just return it regardless of whether `debug` is set or not. I don’t see the point in not setting it since I already have it. Also since the LLM-based tests are not deterministic, it’s best to have the debug dataset from the specific run (without having to run again with `debug=True`, which will give different results).

rabah-khalek:
I'm not sure this is well defined. The standard procedure today is that once you click debug in the hub, you re-run the test with debug=True, you save the output_ds. 

Not sure if the dataset is persisted only when the debug is clicked, or whether by always providing the dataset, the dataset gets saved twice.

But in any case, even with no check on debug, the debug click will produce a new dataset either way.

rabah-khalek:
In fact the mechanism of debug is also going to change with this https://github.com/Giskard-AI/giskard-hub/pull/172/files, where only the list of indices are now transfer to the backend, and the dataset gets retrieved and only filtered, not created...

mattbit:
> In fact the mechanism of debug is also going to change with this https://github.com/Giskard-AI/giskard-hub/pull/172/files, where only the list of indices are now transfer to the backend, and the dataset gets retrieved and only filtered, not created...

Nice! That’s a significant improvement, which also avoids polluting the project artifacts with tens of temporary datasets.

mattbit:
> Not sure if the dataset is persisted only when the debug is clicked, or whether by always providing the dataset, the dataset gets saved twice.

I will double check. It doesn’t seems it’s saved in the tests I’ve done, but I may have missed that.

mattbit:
@rabah-khalek I fixed a problem with the single value test and added a couple of unit tests. The debug output_df is not a problem (does not create duplicates).

rabah-khalek (approved):

PR #1524, opened 2023-10-31
[GSK-2020] Updated category threshold

rabah-khalek:
Setting category threshold to zero if `len(df) <= 2`


giskard/datasets/base/__init__.py
```diff
@@ -194,7 +194,12 @@ def __init__(
         self.column_dtypes = self.extract_column_dtypes(self.df)
 
         # used in the inference of category columns
-        self.category_threshold = round(np.log10(len(self.df))) if len(self.df) >= 100 else 2
+        df_size = len(self.df)
+        # if df_size >= 100     ==> category_threshold = floor(log10(df_size))
+        # if 2 < df_size < 100  ==> category_threshold = 2
+        # if df_size <= 2       ==> category_threshold = 0 (column is text)
+        # df_size != 0 to avoid <stdin>:1: RuntimeWarning: divide by zero encountered in log10
+        self.category_threshold = max(np.floor(np.log10(df_size)), 2) * (df_size > 2) if df_size != 0 else 0
```
mattbit, line 202: I don’t think `if df_size != 0 else 0` is required.

```py
max(np.floor(np.log10(df_size)), 2) * (df_size > 2)
```
should suffice.

rabah-khalek: > I don’t think if df_size != 0 else 0 is required.

It's to avoid the runtime error that you'd get when evaluating `np.log10(0)`. 

```
>>> import numpy as np; np.log10(0)
<stdin>:1: RuntimeWarning: divide by zero encountered in log10
-inf
```

mattbit: oh I see

mattbit (approved):

PR #1527, opened 2023-10-31
[GSK-1956] Suppressing target warning

rabah-khalek:
In the LLM use-cases, `Dataset.target` becomes irrelevant. We added the option `False` to suppress the warning related to validation.

@mattbit is there anything else to do? This has to be seen as a quick fix, until we refactor the model wrapping to individual flavours.

PR #1531, opened 2023-10-31
[GSK-1956] Moved target validation and silenced it for LLMs

rabah-khalek:
Supersedes https://github.com/Giskard-AI/giskard/pull/1527


andreybavt:
@rabah-khalek can you check the unit tests?

giskard/core/dataset_validation.py
```diff
@@ -8,18 +8,20 @@
...
-    else:
-        if ds.target not in ds.columns:
-            raise ValueError(
-                "Invalid target parameter:"
-                f" '{ds.target}' column is not present in the dataset with columns: {list(ds.columns)}"
-            )
+
+
+def validate_target_exists(ds: Dataset):
+    if ds.target and ds.target not in ds.columns:
```
mattbit, line 20: We must check explicitly for the value of `ds.target`. For example, 0 should be a valid target name (as far as pandas is concerned) but it would evaluate to False here.

mattbit: Explicit check for `ds.target is None`

giskard/core/model_validation.py
```diff
@@ -14,11 +14,14 @@
...
 from ..utils import fullname
 from ..utils.analytics_collector import analytics, get_dataset_properties, get_model_properties
+from .dataset_validation import validate_optional_target
 
 
 @configured_validate_arguments
 def validate_model(model: BaseModel, validate_ds: Optional[Dataset] = None, print_validation_message: bool = True):
     try:
+        if model.meta.model_type != SupportedModelTypes.TEXT_GENERATION and validate_ds is not None:
+            validate_optional_target(validate_ds)
```
mattbit, line 24: Why do we need this in `validate_model`???

mattbit (changes requested):
To rediscuss.

giskard/core/model_validation.py
rabah-khalek: we should refactor `model_validation` and `dataset_validation` into one common `validation`, but the way the code is structured right now, I needed `model.meta.model_type` to do `validate_optional_target` for non-LLM models.

giskard/core/dataset_validation.py
rabah-khalek: agreed

giskard/core/model_validation.py
rabah-khalek: Let's just fix the llm-warning issue for now and deal with the refactoring post-release, I added a card here: https://linear.app/giskard/issue/GSK-2064/refactor-model-and-dataset-validation

mattbit (approved):
Good but we should consider refactoring this soon.

PR #1542, opened 2023-11-02
[GSK-2037] Fix test for Benjamini-Hochberg filtering in PerformanceBiasDetector

mattbit:
Issues:

- the Titanic dataset in the fixtures is not initialized correctly: it is missing the categorical features, which are then considered "text"
- normally this should not be a problem because the text slicer would work equally well, creating the same slices when there are few unique values
- but by default, the vectorizer ignores words of less than 2 chars, and in this case the values are single-letter values "S", "Q", and they are ignored by the text vectorizer

The behaviour of the vectorizer makes sense. On the contrary, I fixed the fixture Dataset by putting the correct categorical features. This can possibly break other tests.

---

Side note: while checking how the slices were treated in the `LossBasedDetector`, I added an extra filter on slice size which should make the slicing detectors roughly 40% faster.

### DIFF
--- a/giskard/scanner/common/loss_based_detector.py
+++ b/giskard/scanner/common/loss_based_detector.py
@@ -5,13 +5,13 @@ from typing import Sequence
 
 import pandas as pd
 
-from ..issues import Issue
-from ..logger import logger
-from ..registry import Detector
 from ...datasets.base import Dataset
 from ...ml_worker.testing.registry.slicing_function import SlicingFunction
 from ...models.base import BaseModel
 from ...slicing.slice_finder import SliceFinder
+from ..issues import Issue
+from ..logger import logger
+from ..registry import Detector
 
 
 class LossBasedDetector(Detector):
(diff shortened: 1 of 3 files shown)


giskard/scanner/common/loss_based_detector.py
```diff
@@ -95,8 +95,11 @@ def _find_slices(self, model: BaseModel, dataset: Dataset, meta: pd.DataFrame):
         sliced = sf.run(dataset_with_meta, features, target=self.LOSS_COLUMN_NAME)
         slices = sum(sliced.values(), start=[])
 
-        # Keep only slices of size at least 5% of the dataset or 20 samples (whatever is larger)
-        slices = [s for s in slices if max(0.05 * len(dataset), 20) <= len(dataset_with_meta.slice(s))]
+        # Keep only slices of size at least 5% of the dataset or 20 samples (whatever is larger) and conversely exclude
+        # slices which are larger than 95% of the dataset
+        slices = [
+            s for s in slices if max(0.05 * len(dataset), 20) <= len(dataset_with_meta.slice(s)) <= 0.95 * len(dataset)
```
rabah-khalek, line 101: What's the harm in detecting issues with slices of > 95%?
wasn't the case of a user who detected a 100% spurious correlation in his model he wasn't aware about, and was super happy about it?

mattbit: - this is just for loss-based detectors, spurious correlation has impact here
- even if it was included, for how spurious correlation is calculated, 100% would just mean that the target had a unique constant value, and it would not make sense to talk about correlation
- for other loss-based detectors, as we filter out slices < 5% of the dataset, it makes sense to also filter out slices which are larger than 95% because they would be compared against a < 5% complementary data slice
- in the end, this filter does not change the results but it allows us to skip a good number of useless comparisons which would have been dropped later anyway

rabah-khalek: Ok clear, Thanks!

rabah-khalek: 100% correlation, doesn't necessarily imply a target having a unique constant value no?
It's been a while but wouldn't this:
featureA featureB target
1 1 0
2 1 0
3 0 1
4 0 1
5 1 0
be considered that featureB is 100% correlated with target?

rabah-khalek (approved):
good to go

PR #1543, opened 2023-11-02
Add scope "session" to fixtures to avoid expensive computation

mattbit:
Draft


mattbit:
Some tests fail because they don’t expect fixtures to the singletons, in particular for `giskard.Model` or `giskard.Dataset` instances.

Hartorn:
Nightly run here : https://github.com/Giskard-AI/giskard/actions/runs/6783970675/job/18439314300

rabah-khalek (approved):

PR #1545, opened 2023-11-03
Import lightgbm in test init to avoid bug in macOS build

mattbit:


Hartorn:
Test launched here : https://github.com/Giskard-AI/giskard/actions/runs/6745784902

Hartorn:
Fixed with https://github.com/Giskard-AI/giskard/pull/1555

PR #1552, opened 2023-11-04
GSK-2060: take into account `debug` argument when running LLM tests

andreybavt:


rabah-khalek:
@andreybavt If the absence of this check was causing issues, `test_llm_char_injection` should also cause the same issue. I added it in https://github.com/Giskard-AI/giskard/pull/1556

PR #1569, opened 2023-11-06
Support for openai>=1.0

mattbit:
Adding support for the new API following the release of `openai==1.0.0`.


mattbit:
@Hartorn I fixed openai>=1 in for the test environment because I create instances of classes defined there. Since this is going to be the more supported version, I think it’s safe to keep it like this.
I still test the legacy version, but in that case I pass `dict`s, since the objects in openai==0.28.1 are compatible with `dict`.

Hartorn:
> @Hartorn I fixed openai>=1 in for the test environment because I create instances of classes defined there. Since this is going to be the more supported version, I think it’s safe to keep it like this. I still test the legacy version, but in that case I pass `dict`s, since the objects in openai==0.28.1 are compatible with `dict`.

👍 I regenerated the lock file, for the version to be used.

Should we run some tests with version <1 anyway ?

mattbit:
> Should we run some tests with version <1 anyway ?

We don’t need that, I added tests to mock the two versions of the API. Currently we don’t have any end-to-end test that calls OpenAI API, it’s all mocked.

Hartorn (approved):

rabah-khalek (approved):

PR #1596, opened 2023-11-09
[GSK-2139] Debug memory leak

Hartorn:
## Type of Change

- [ ] 🔐 Security fix

## Checklist


mattbit:
@Hartorn some fixes for `LLMCharInjector` in #1600 should slightly reduce memory usage.

.github/workflows/build-python.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         pydantic_v1: [false]
         pandas_v1: [false]
         langchain_minimal: [false]
-        # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+        # # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
```
mattbit, line 61: double comment

.github/workflows/build-python.yml
```diff
@@ -160,28 +160,29 @@ jobs:
...
+        if: ${{ startsWith(matrix.os, 'windows') }}
+        run: pdm test-ram tests/models tests/testing tests/test_metamorphic_invariance.py
 
-      - name: Tests integrations/
-        env:
-          PYTEST_XDIST_AUTO_NUM_WORKERS: ${{ startsWith(matrix.os,'windows-') && 1 || 2 }}
-        run: pdm test-integrations
+      - name: Test code (RAM expensive, /tests/models and /tests/testing)
+        if: ${{ !startsWith(matrix.os, 'windows') }}
+        run: pdm test-ram --memray tests/models tests/testing tests/test_metamorphic_invariance.py
```
mattbit, line 169: should we remove the `--memray` by default or you think it’s useful too keep it?

mattbit: In the future: can find some way to mark tests and better distribute them (e.g. avoiding running in parallel certain tests), instead of this manual setting at runtime?

conftest.py
mattbit, line 1: Do we need this except for debugging?

conftest.py
```diff
@@ -0,0 +1,65 @@
...
+import logging
+import os
+import time
+from pathlib import Path
+
+import psutil
+import pytest
+from _pytest.python import Function
+
+LOGGER = logging.getLogger(__name__)
```
mattbit, line 11: lowercase?

.github/workflows/build-python.yml
Hartorn: I think it can stays. This kind of insights maybe useful
image

conftest.py
Hartorn: Yeah, let's keep it.
Since I've added an artifact into the build, we can survey the memory usage from time to time.
Also, if we have any crash, we'll still get the artifacts to debug

conftest.py
Hartorn: Well, constant should be in uppercase in python normally.
https://peps.python.org/pep-0008/#constants

.github/workflows/build-python.yml
Hartorn: For the marking, I did add a pytest mark to not run memory intensive tests in parallel, but still had to split them, else it was too much

maybe we can find a way to run each RAM intensive tests independently with a sub process and kill it everytime ?

mattbit: What is the difference then? The weight of the fixture?

conftest.py
mattbit: Yeah, but that’s a mutable instance of `logging.Logger`. `LOGGER` here is not a constant, but a global instance.

mattbit: E.g. similar case: https://github.com/pytorch/pytorch/blob/44f1c6e41c3cf636ba8f691fafbe3b4be621e718/torch/utils/_cuda_trace.py#L6C37-L6C37

mattbit: By the way, where do we actually use this?

.github/workflows/build-python.yml
Hartorn: Done

conftest.py
Hartorn: Removed, not used.

mattbit (approved):

.github/workflows/build-python.yml
Hartorn: No, but when running one test in a process and killing it, all the memory used by the process is going to be freed.
Let's say we use xdist, but recreating a new process every N tests => all the lost memory of our tests will be free as soon as the process is killed.

PR #1580, opened 2023-11-07
Improve docstring parsing for Hub catalog

mattbit:
- Reformat all docstrings in unifrom numpydoc style.
- Add a simple parser from `numpydoc` to parse the different sections of the docstring.

PR #1599, opened 2023-11-10
GSK-2092 Fix LLM test doc to properly render in frontend

Googleton:
## Description

Reformatted the docstrings of the LLM tests so that the parsing on the frontend can read the properly and set hints instead of it being shown as a description in the test.


mattbit:
We are standardizing to numpydoc and improving parsing here: #1580 #1581

PR #1616, opened 2023-11-17
How to export scan report in quickstarts [GSK-2151]

mattbit:
Added instructions about how to export scan report to HTML, useful if the scan is not performed in a notebook.


luca-martial (approved):
Grazie!

PR #1620, opened 2023-11-20
[GSK-2135] Show warning if dependencies or python versions are mismatched

Inokinoki:
## Type of Change

- [ ] 🔐 Security fix

## Checklist

### DIFF
--- a/giskard/ml_worker/websocket/__init__.py
+++ b/giskard/ml_worker/websocket/__init__.py
@@ -1,9 +1,9 @@
 from enum import Enum
-from typing import Dict, List, Optional
 
 import pydantic
 from packaging import version
 from pydantic import Field
+from typing import Dict, List, Optional
 
 from giskard.core.validation import ConfiguredBaseModel
 
@@ -323,6 +323,7 @@ class TestSuite(WorkerReply):
 
 
 class TestSuiteParam(ConfiguredBaseModel):
+    projectKey: str
     tests: Optional[List[SuiteTestArgument]] = None
     globalArguments: Optional[List[FuncArgument]] = None
 
--- a/giskard/ml_worker/websocket/listener.py
+++ b/giskard/ml_worker/websocket/listener.py
@@ -10,12 +10,12 @@ from concurrent.futures import CancelledError, Future
 from copy import copy
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Dict, Optional, Union
 
 import numpy as np
 import pandas as pd
 import pkg_resources
 import psutil
+from typing import Any, Callable, Dict, Optional, Union
 
 import giskard
 from giskard.client.giskard_client import GiskardClient
@@ -562,7 +562,7 @@ def run_test_suite(
             identifier_single_test_results.append(
                 websocket.IdentifierSingleTestResult(
                     id=identifier,
-                    result=map_result_to_single_test_result_ws(result, datasets),
+                    result=map_result_to_single_test_result_ws(result, datasets, client, params.projectKey),
                     arguments=function_argument_to_ws(arguments),
                 )
             )
--- a/tests/communications/fixtures/with_alias.json
+++ b/tests/communications/fixtures/with_alias.json
@@ -2866,10 +2866,12 @@
   ],
   "TestSuiteParam": [
     {
+      "projectKey": "my_project",
       "tests": [],
       "globalArguments": []
     },
     {
+      "projectKey": "awesome_project",
       "tests": [
         {
           "id": 2201,
@@ -2942,6 +2944,7 @@
       "globalArguments": []
     },
     {
+      "projectKey": "amazing_project",
       "tests": [
         {
           "id": 5848,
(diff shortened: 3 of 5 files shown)


giskard/ml_worker/exceptions/giskard_exception.py
```diff
@@ -1,3 +1,40 @@
...
+class GiskardPythonEnvException(GiskardException):
+    def __init__(self, *args: object) -> None:
+        super().__init__(*args)
+
+
+class GiskardPythonVerException(GiskardPythonEnvException):
+    def __init__(self, cls_name, e, *args: object) -> None:
+        super().__init__(
+            f"Failed to load '{cls_name}' due to {e.__class__.__name__}.\n"
+            "Make sure you are loading it in the environment with matched Python version.",
```
kevinmessiaen, line 15: Can we add the Python version from the pickle meta, it will help user to fix issue

giskard/ml_worker/exceptions/giskard_exception.py
```diff
@@ -1,3 +1,40 @@
...
+            *args,
+        )
+
+
+def python_env_exception_helper(cls_name, e: Exception):
+    if isinstance(e, TypeError):
+        # Python 3.9: code() takes at most 16 arguments (18 given)
+        # Python 3.10: code expected at most 16 arguments, got 18
+        # Python 3.11: code() argument 13 must be str, not int
+        if "at most 16 arguments" in e.args[0] or "code() argument 13 must be str, not int" in e.args[0]:
```
kevinmessiaen, line 34: We should probably compare the Python version meta with the current python version to be independent of the format (ei. Python 3.12 might break this code)

kevinmessiaen (changes requested):
LGTM but we should leverage the Python version in meta to give more insight to the user when raising the exception

giskard/ml_worker/exceptions/giskard_exception.py
Inokinoki: Yes, that's what we are considering as well

giskard/ml_worker/exceptions/giskard_exception.py
Inokinoki: With the version meta, it should be feasible

Inokinoki: Do you think whether we should make another PR for the Python versions of the callable function?
I am drafting one for the Giskard Hub DB and backend changes: https://github.com/Giskard-AI/giskard-hub/pull/256

kevinmessiaen: Not sure if it's worth since we will probably get ride of the `pickle` serialization to just saving python code.

Inokinoki: OK, let me keep this way at the moment. And add the Python version to the code later

giskard/ml_worker/exceptions/giskard_exception.py
```diff
@@ -1,3 +1,48 @@
...
+    def __init__(self, cls_name, e, *args: object) -> None:
+        super().__init__(
+            f"Failed to load '{cls_name}' due to {e.__class__.__name__}.\n"
+            "Make sure you are loading it in the environment with matched dependencies.",
+            *args,
+        )
+
+
+def python_env_exception_helper(cls_name, e: Exception, required_py_ver: Optional[Tuple[str, str, str]] = None):
+    if required_py_ver is not None and required_py_ver[:2] != platform.python_version_tuple[:2]:
```
Inokinoki, line 34: I also worry about the false positive here: it's actually caused by dependency issues, but the Python version mismatch leads to this branch. @kevinmessiaen WDYT

kevinmessiaen: That's true that it'll create false positive, however for now we only use pickle so that's not a big deal. We will probably handle Python code differently once we improve the code serialization

Inokinoki: How about `TypeError` and `minor` version not matched?

kevinmessiaen: Well we should probably keep track of the Python version that we used to generate pickle since we used [cloudpickle](https://github.com/cloudpipe/cloudpickle)

Has stated in there readme:
image

Here is an explanation of why it is the case: https://github.com/cloudpipe/cloudpickle/issues/293#issuecomment-502011167

Catching `TypeError` is not reliable enough IMO, better keeping track of the version internally

Inokinoki: Done!

kevinmessiaen (approved):

giskard/ml_worker/exceptions/giskard_exception.py
```diff
@@ -1,3 +1,39 @@
...
```
kevinmessiaen, line 34: Missing function call

```suggestion
    if required_py_ver is not None and required_py_ver[:2] != platform.python_version_tuple()[:2]:
```

Inokinoki: Comparing the whole tuple might result in more false-positives. And it should work well across different patch levels.

Issue #1614, opened 2023-11-16
error while running LLM-as-a-judge test: Must provide an 'engine' or 'deployment_id' parameter to create a <class 'openai.api_resources.chat_completion.ChatCompletion'>


younes-io:
### Issue Type

Bug

### Source

source

### Giskard Library Version

2.0.3

### Giskard Hub Version

N/A

### Python version

3.11

### Current Behaviour?

```shell
I'm trying to use LLM-as-a-judge methodology to test my chatbot (https://docs.giskard.ai/en/latest/reference/tests/llm.html#llm-as-a-judge).

When I run run the code, I get the error log below.

I don't know why it asks for a deployment, even though it's provided when I create the `AzureChatOpenAI` :


llm = AzureChatOpenAI(
        openai_api_key = openai_api_key,
        openai_api_base = openai_api_base,
        openai_api_type = openai_api_type,
        openai_api_version = openai_api_version,
        deployment_name = model_name,
        temperature=0,
    )
```

### Standalone code OR list down the steps to reproduce the issue

```shell
def call_qa(user_profile):
    print("calling call_qa..")
    # Define Azure OpenAI component
    llm = AzureChatOpenAI(
        openai_api_key = openai_api_key,
        openai_api_base = openai_api_base,
        openai_api_type = openai_api_type,
        openai_api_version = openai_api_version,
        deployment_name = model_name,
        temperature=0,
    )
    print("llm._default_params : ", llm._default_params)
    
    qa = RetrievalQAWithSourcesChain.from_chain_type(
        memory=memory,
        llm=llm,
        chain_type="stuff",
        retriever=doc_retriever,
        return_source_documents=True,
        verbose=True,
        chain_type_kwargs=chain_type_kwargs,
    )
    return qa
    
    
 
qa = call_qa({"Country": "UK"})
response = qa({"question": user_input}) ## this is working



import os

import ast
import openai
import tiktoken
import pandas as pd
from scipy import spatial

from giskard import scan, Dataset, Model

TEXT_COLUMN_NAME = "text"

# Optional: Wrap a dataframe of sample input prompts to validate the model wrapping and to narrow specific tests' queries.
corpus = [
    "What should I do in case of sickness?",
    "How does the company take care of the problems cause by increased stressed at work?",
]

raw_data = pd.DataFrame(data={TEXT_COLUMN_NAME: corpus})
giskard_dataset = Dataset(raw_data)

# Wrap the model.
def prediction_function(df):
[...]

PR #1622, opened 2023-11-21
GSK-2177 & GSK-2175: Allow to run the LLM scan using Azure

kevinmessiaen:
## Description

This PR introduce a way to execute the LLM scan using AzureOpenAI. 

AzureOpenAI is used if any `AZURE_OPENAI_API_KEY` has been provided.

An environment variable called `GISKARD_SCAN_LLM_MODEL` has been introduced, this variable allow setting up the model being used by AzureOpenAI. This can also be used to override default model used by OpenAI, however this is *not recommended*

Lastly we can keep using OpenAI even if `AZURE_OPENAI_API_KEY` has been provided by setting `GISKARD_SCAN_API` to `openai`

## Related Issue

- https://github.com/Giskard-AI/giskard/issues/1614
- [GSK-2177 (available on linear)](https://linear.app/giskard/issue/GSK-2177/add-a-way-to-support-azure-on-llm-scan)
- [GSK-2175 (available on linear)](https://linear.app/giskard/issue/GSK-2175/add-a-disclaimer-stating-that-the-llm-scan-use-openai-api)

## Type of Change

- [ ] 🔐 Security fix


giskard/llm/client/utils.py
```diff
@@ -0,0 +1,38 @@
+import logging
+import os
+from enum import Enum
+from typing import Optional
+
+
+class ApiType(Enum):
+    openai = (0,)
+    azure = 1
```
mattbit, line 9: - constants should be uppercase
- enums start at 1 by default
- let’s find a more descriptive name
- acronyms in uppercase
see [PEP8](https://peps.python.org/pep-0008/).

```suggestion
class LLMClientAPI(Enum):  # maybe can find a better name
    OPEN_AI = 1
    AZURE = 2
```  

Also, this enum shold be moved in a more appropriate namespace than `utils`.

giskard/llm/client/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def autodetect_client():
+    if "GISKARD_SCAN_LLM_API" in os.environ:
+        try:
+            return ApiType[os.environ["GISKARD_SCAN_API"].lower()]
```
mattbit, line 15: Mismatch in the env variable name. I would not link it to the scan only (note that we are not in the `scanner` module), because it’s also used in the tests.

`GSK_LLM_API`? (We use `GSK` instead of `GISKARD` for other env variables)

mattbit: As we do with other settings, I would put a global python variable somewhere instead of only relying on the environment variable. We should always allow to set things programmatically via Python.

```py3
import giskard as gsk

gsk.llm.set_llm_api(LLMClientAPI.OPEN_AI)
```

The environment should only be used to set some default.

I would put something like this in `__init__.py`:
```py3
_default_client_api = os.getenv("GSK_LLM_API", LLMClientAPI.AZURE if "AZURE_OPENAI_API_KEY" in os.environ else LLMClientAPI.OPEN_AI)
```

giskard/llm/client/utils.py
```diff
@@ -0,0 +1,38 @@
...
+            )
+
+    if "AZURE_OPENAI_API_KEY" in os.environ:
+        return ApiType.azure
+
+    return ApiType.openai
+
+
+def get_scan_model(api_type: ApiType) -> Optional[str]:
+    scan_model = os.environ.get("GISKARD_SCAN_LLM_MODEL")
```
mattbit, line 28: This is tricky because we refer to GPT-4 when using the various clients. That was over-optimization, could you refactor that to get rid of the model choice at the Generator/Evaluator level? This would also include the `LLMClient` interface.

giskard/llm/client/openai.py
```diff
@@ -147,7 +149,7 @@ def _completion(
 
         try:
             completion = self._client.chat.completions.create(
-                model=model,
+                model=self._model or model,
```
mattbit, line 152: Wrote also a comment below, this is bad because the Evaluators/Generators are also specifying the model, and here you just overwrite it.
I think you are right and it’s a problem with the current implementation. Let’s bind the model at the `LLMClient` level, passing it in init as you did here.
Let’s get rid of the `model` parameter when calling the `complete` method and let’s remove it from all generators/detectors. It will make things easier in the future.

mattbit (changes requested):

giskard/llm/client/__init__.py
```diff
@@ -1,14 +1,47 @@
...
+
+def set_llm_model(llm_model: str):
+    global _default_llm_model
+    _default_llm_model = llm_model
+
+
+def get_default_llm_api() -> LLMClientAPI:
+    global _default_llm_api
+    if _default_llm_api is None:
+        _default_llm_api = os.getenv(
```
mattbit, line 38: I’m just realizing that `StrEnum` is only supported from Python 3.11. The code here would now work as `os.getenv` returns a string and we would need to compare.

Maybe I would entirely drop the enum for now, and just use pure strings, so that we don’t have to come up with magic ways to support 3.9 and 3.10.

mattbit (changes requested):
Thanks @kevinmessiaen, very good! Just one minor bug with the enum that I realized late

mattbit (commented):
Thanks @kevinmessiaen. Looks good, I will test this locally as soon as possible.

mattbit (approved):
Works perfectly!

PR #1624, opened 2023-11-21
Exclude anonymous dataset from AVID report

mattbit:
- Fixed an issue with AVID report generation when the `Dataset` object has no name
- Same for `Model` object with no name (we take the class name)
- Added more tests

### DIFF
--- a/giskard/integrations/avid.py
+++ b/giskard/integrations/avid.py
@@ -47,8 +47,7 @@ def create_report_from_issue(issue: Issue, model: BaseModel, dataset: Dataset = 
         AVID Report created from the issue.
     """
     # Prepare artifacts
-    model_name = model.meta.name or model.__class__.__name__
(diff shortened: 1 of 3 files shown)


giskard/integrations/avid.py
```diff
@@ -47,8 +47,9 @@ def create_report_from_issue(issue: Issue, model: BaseModel, dataset: Dataset =
         AVID Report created from the issue.
     """
     # Prepare artifacts
-    artifacts = [Artifact(type=ArtifactTypeEnum.model, name=model.meta.name)]
-    if dataset:
+    model_name = model.meta.name or model.__class__.__name__
```
rabah-khalek, line 50: Your previous implementation was fine. This is already done for you in the BaseModel class. I'd leave it as it was.

rabah-khalek (approved):
good to go, only one minor comment to take care of.

PR #1625, opened 2023-11-21
Running tests in threads to avoid issues due to memory leaks

kevinmessiaen:
## Description

Running detectors in pool to avoid issues due to memory leaks

## Type of Change

- [ ] 🔐 Security fix

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,16 +1,14 @@
 import os
-import threading
 from glob import glob
 from io import TextIOWrapper
 from pathlib import Path
-from typing import List, Tuple
 from uuid import uuid4
 
-import pytest
 from _pytest.config import Config, ExitCode
 from _pytest.reports import BaseReport
 from _pytest.terminal import TerminalReporter
 from pytest import CollectReport, TestReport
+from typing import List, Tuple
 
 pytest_plugins = []
 for f in glob("**/fixtures/**/*.py", recursive=True):
@@ -96,30 +94,3 @@ def pytest_terminal_summary(terminalreporter: TerminalReporter, exitstatus: Exit
         writer.write("\n### Failures\n\n")
         for failure in failures:
             _write_report(writer, failure)
-
-
-@pytest.hookimpl(tryfirst=True)
-def pytest_runtest_protocol(item):
-    ihook = item.ihook
-    ihook.pytest_runtest_logstart(nodeid=item.nodeid, location=item.location)
-    reports = threaded_run_report(item)
-    for rep in reports:
-        ihook.pytest_runtest_logreport(report=rep)
-    ihook.pytest_runtest_logfinish(nodeid=item.nodeid, location=item.location)
-    return True
-
-
-def threaded_run_report(item) -> List[TestReport]:
-    # for now, we run setup/teardown in the subprocess
-    # XXX optionally allow sharing of setup/teardown
-    from _pytest.runner import runtestprotocol
-
-    def run_threaded(reports: List[TestReport]):
-        reports.extend(runtestprotocol(item, log=False))
-
-    reports = []
-    thread = threading.Thread(target=run_threaded, args=(reports,))
-    thread.start()
-    thread.join()
-
-    return reports
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,11 +1,12 @@
-from typing import List, Tuple
-
 import os
+import threading
 from glob import glob
 from io import TextIOWrapper
 from pathlib import Path
+from typing import List, Tuple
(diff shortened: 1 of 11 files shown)


tests/conftest.py
```diff
@@ -95,3 +96,30 @@ def pytest_terminal_summary(terminalreporter: TerminalReporter, exitstatus: Exit
         writer.write("\n### Failures\n\n")
         for failure in failures:
             _write_report(writer, failure)
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_runtest_protocol(item):
+    ihook = item.ihook
```
Hartorn, line 103: Would be nice to only run for marked test, using "memory_expensive"

tests/conftest.py
```diff
@@ -95,3 +96,30 @@ def pytest_terminal_summary(terminalreporter: TerminalReporter, exitstatus: Exit
...
+def threaded_run_report(item) -> List[TestReport]:
+    # for now, we run setup/teardown in the subprocess
+    # XXX optionally allow sharing of setup/teardown
+    from _pytest.runner import runtestprotocol
+
+    def run_threaded(reports: List[TestReport]):
+        reports.extend(runtestprotocol(item, log=False))
+
+    reports = []
+    thread = threading.Thread(target=run_threaded, args=(reports,))
```
Hartorn, line 121: Running in a thread will not help regarding memory leak, as it stays shared with main process.
To be sure, I ran `pdm run pytest tests/ -m "memory_expensive"` before and after, and it did not change anything (memory usage goes up to 7go, you can see it locally by `checking _memory_tests.csv`)

You should use subprocess and not threading, so that way memory is actually cleared.

You can do that either using a process pool executor 
https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor-example
especially instiating one and using `max_tasks_per_child=1`.
That way, task would be executed in a separated process and cleaned up

Or you could spawn manually a process and handle, but it will be much more difficult.

Hartorn: Honestly, we could kind of replace xdist at the same time, by handling with several pools and tasks, to handle parallelism as we want. But that's another question :)

Hartorn (changes requested):

tests/conftest.py
kevinmessiaen: The issue is that `item`, is not pickable. Therefore it's not that easy to start a subprocess, while the fork method works because we don't need to pickle

Hartorn: What is the type of item ? maybe we can find some way to serialize/unserialize it ?

kevinmessiaen: > It's a Node containing `_Pyhtest.python.Function` which is very hard to serialize because it contains references to a bunch of objects which are not simple to serialize, like the FixtureManager.

https://github.com/pytest-dev/pytest-xdist/issues/675#issuecomment-868400332

kevinmessiaen:
After lot of tests I decided to use `dill` to serialize function executed in pools (it is more consistent that pickle since it handle lambas, ... that we use in Models, Datasets and Issues).

I did it in test level because hooks are not pickables.

The issue is that bert model use a lot of memory and is not released, I could not clean it manually with neither `torch.cuda.empty_cache()` nor `with tock.no_grad():`. We will probably need to fix the issue in worker by running tasks in thread pool

Hartorn:
> After lot of tests I decided to use `dill` to serialize function executed in pools (it is more consistent that pickle since it handle lambas, ... that we use in Models, Datasets and Issues).
> 
> I did it in test level because hooks are not pickables.
> 
> The issue is that bert model use a lot of memory and is not released, I could not clean it manually with neither `torch.cuda.empty_cache()` nor `with tock.no_grad():`. We will probably need to fix the issue in worker by running tasks in thread pool

It's a bit of a shame to have to change each test one by one. 
Nice work with dill, will check if I can have another way to avoid thoses changed

kevinmessiaen:
> > After lot of tests I decided to use `dill` to serialize function executed in pools (it is more consistent that pickle since it handle lambas, ... that we use in Models, Datasets and Issues).
> > I did it in test level because hooks are not pickables.
> > The issue is that bert model use a lot of memory and is not released, I could not clean it manually with neither `torch.cuda.empty_cache()` nor `with tock.no_grad():`. We will probably need to fix the issue in worker by running tasks in thread pool
> 
> It's a bit of a shame to have to change each test one by one. Nice work with dill, will check if I can have another way to avoid thoses changed

I tried with adding an decorator to the tests to automatically wrap but it mess up with the pytest framework

Hartorn:
done in https://github.com/Giskard-AI/giskard/pull/1640

PR #1628, opened 2023-11-22
Fixed Sequence import (test failed on Windows)

luca-rossi:


giskard/llm/client/openai.py
```diff
@@ -2,7 +2,7 @@
 from abc import ABC, abstractmethod
 from typing import Dict, Optional
 
-from git import Sequence
+from typing import Sequence
```
mattbit, line 5: please remember to run ruff to fix the imports styling
```py
from typing import Dict, Optional, Sequence
```

mattbit (changes requested):
Fix import code style

mattbit (approved):

PR #1629, opened 2023-11-22
[GSK-1949] Maintain class proportions after slicing in classification models

luca-rossi:


giskard/scanner/common/loss_based_detector.py
```diff
@@ -73,6 +70,30 @@ def run(self, model: BaseModel, dataset: Dataset, **kwargs):
     def _numerical_slicer_method(self):
         return "tree"
 
+    def _slice_dataset(self, model: BaseModel, dataset: Dataset):
```
mattbit, line 73: Can we move this function out of the class and give it a more descriptive name?

giskard/scanner/common/loss_based_detector.py
```diff
@@ -73,6 +70,30 @@ def run(self, model: BaseModel, dataset: Dataset, **kwargs):
...
+            num = dataset.df[dataset.target].value_counts()[i]
+            props.append(num / len(dataset))
+        # sample the dataset while keeping the original proportion of each class
+        sliced_dataset = Dataset(
+            pd.DataFrame(), target=dataset.target, column_types=dataset.column_types, validation=False
+        )
+        for i in range(len(values)):
+            part_dataset = dataset.df[dataset.df[dataset.target] == values[i]]
+            part_dataset = part_dataset.sample(int(max_data_size * props[i]), random_state=42)
+            sliced_dataset.df = pd.concat([sliced_dataset.df, part_dataset])
```
mattbit, line 94: I would do something like
```py3
keep_ratio = max_data_size / len(dataset)

sliced = dataset.slice(
    lambda df: df.groupby(dataset.target, group_keys=False, sort=False).sample(frac=keep_ratio),
    row_level=False,
)

return sliced
```

mattbit (changes requested):

mattbit:
Also note that some test failed with a “division by 0 error“

mattbit (commented):
Good for me, need to add tests

mattbit (approved):

PR #1635, opened 2023-11-24
Only scan a subset of features [GSK-2173]

mattbit:
The Giskard scan can now be limited to an arbitrary subset of features:

```py3
import giskard as gsk
...
report = gsk.scan(model, dataset, features=["feature_1", "feature_2"])
```

### DIFF
--- a/giskard/scanner/scanner.py
+++ b/giskard/scanner/scanner.py
@@ -240,6 +240,12 @@ class Scanner:
                     f"Valid features for this model are: {', '.join(_default_features)}."
                 )
 
+        if len(features) < 1:
+            raise ValueError(
+                "No features to scan. Please provide a non-empty list of features to scan,"
+                "and ensure that you correctly set the `features_names` argument when wrapping your model."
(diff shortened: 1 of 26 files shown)


giskard/scanner/scanner.py
```diff
@@ -213,16 +223,33 @@ def _validate_model_and_dataset(self, model, dataset):
...
+            features = _default_features
+        else:
+            if not set(features).issubset(_default_features):
+                raise ValueError(
+                    "The `features` argument contains invalid feature names: "
+                    f"{', '.join(set(features) - set(_default_features))}. "
+                    f"Valid features for this model are: {', '.join(_default_features)}."
+                )
+
+        if len(features) > 100:
```
kevinmessiaen, line 243: We should probably also validate empty features since some scanner fail due to / by 0
image

kevinmessiaen (approved):
Great work!

kevinmessiaen (approved):

PR #1636, opened 2023-11-24
[GSK-1953] Max data size as param

luca-rossi:
This PR allow users to set maximum number of samples (`max_dataset_size`) in loss-based scan detectors.

### DIFF
--- a/giskard/scanner/common/loss_based_detector.py
+++ b/giskard/scanner/common/loss_based_detector.py
@@ -23,8 +23,8 @@ class LossBasedDetector(Detector):
 
     _needs_target = True
 
-    def __init__(self, max_dataset_size: int = None):
-        self.max_dataset_size = max_dataset_size or self.MAX_DATASET_SIZE
+    def __init__(self, max_dataset_size: int = MAX_DATASET_SIZE):
+        self.max_dataset_size = max_dataset_size
 
     def run(self, model: BaseModel, dataset: Dataset, **kwargs):
         if self._needs_target and dataset.target is None:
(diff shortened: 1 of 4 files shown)


giskard/scanner/common/loss_based_detector.py
```diff
@@ -23,6 +23,9 @@ class LossBasedDetector(Detector):
 
     _needs_target = True
 
+    def __init__(self, max_dataset_size: int = None):
```
mattbit, line 26: Set default directly

giskard/scanner/common/loss_based_detector.py
```diff
@@ -39,7 +42,7 @@ def run(self, model: BaseModel, dataset: Dataset, **kwargs):
             return []
 
         # If the dataset is very large, limit to a subsample
-        max_data_size = self.MAX_DATASET_SIZE // len(model.meta.feature_names or dataset.columns)
+        max_data_size = self.max_dataset_size // len(model.meta.feature_names or dataset.columns)
```
mattbit, line 45: skip the division as discussed

giskard/scanner/common/loss_based_detector.py
```diff
@@ -23,6 +23,9 @@ class LossBasedDetector(Detector):
 
     _needs_target = True
 
+    def __init__(self, max_dataset_size: int = MAX_DATASET_SIZE):
```
mattbit, line 26: inline this and probably divide by ~10?

mattbit (changes requested):
All good, one small change

giskard/scanner/common/loss_based_detector.py
luca-rossi: Ok, also maybe it still makes sense to divide by n_features if the param is not specified?

mattbit: yep, you could set `Optional[int] = None` and calculate the default at runtime if not provided

mattbit:
Broken test!

mattbit (commented):
Ok, but please fix conflicts

mattbit (approved):

Issue #1605, opened 2023-11-13
[GSK-2179] Slicing function DB insert fails because of the long name


martjushev:
### Issue Type

Bug

### Source

docker

### Giskard Library Version

2.0.2

### Giskard Hub Version

2.0.2

### OS Platform and Distribution

macOS Ventura 13.4

### Python version

3.10

### Current Behaviour?

```shell
I'm using featuretools lib to generate features. It can generate pretty long names for the features, like `TIME_SINCE_LAST_MIN(my_date_column123, my_numeric_column123)`. As a result, the automatic slicing function name is longer than 255 characters and a DB insert fails. I believe it makes sense to increase the limit.
```

### Standalone code OR list down the steps to reproduce the issue

```shell
results = scan(giskard_model, giskard_dataset, only=["performance"])
test_suite = results.generate_test_suite("performance")
test_suite.upload(client, project_name)
```

### Relevant log output

```shell
[...]

PR #1639, opened 2023-11-27
[GSK-2179] Limit size of artifact name and display_name to 255 chars when uploading to Giskard Hub

kevinmessiaen:
## Description

Limit size of artifact name and display_name to 255 chars when uploading to Giskard Hub

## Related Issue

- https://github.com/Giskard-AI/giskard/issues/1605
- [GSK- 2179 (Available on linear)](https://linear.app/giskard/issue/GSK-2179/slicing-function-db-insert-fails-because-of-the-long-name-1605)

## Type of Change

- [ ] 🔐 Security fix

## Checklist


giskard/client/giskard_client.py
```diff
@@ -74,6 +73,15 @@ def __call__(self, r):
         return r
 
 
+def _limit_str_size(json, field, limit=255):
+    if field not in json or not isinstance(json[field], str):
+        return
+
+    if len(json[field]) > limit:
```
Inokinoki, line 80: I tested emoji in the name.

In Python 3, `len`  counts the number of utf-8 characters:

```
>>> len("123✋56")
6
>>> len("123✋🔥56")
7
```

while in the Java backend and database, the case could be different:

image

image

Could we check the number of utf-16 characters instead?
Ref: https://docs.oracle.com/javase/8/docs/api/java/nio/charset/Charset.html#:~:text=The%20native%20character%20encoding%20of,chars)%20and%20sequences%20of%20bytes.

Inokinoki (changes requested):

giskard/client/giskard_client.py
Inokinoki: In addition, the case could also be different in PostgreSQL in our production env. We may have to check deeper the encoding used in production.

kevinmessiaen: Seems that we use the EN_UTF-8 local in our env since we use `12-alpine` without LANG variable: https://github.com/docker-library/postgres/blob/038c4c577a3c58dddf9ec2ccaa643009b8ba414b/12/alpine3.18/Dockerfile

kevinmessiaen: I fixed the truncate logic

giskard/client/giskard_client.py
```diff
@@ -74,6 +73,33 @@ def __call__(self, r):
         return r
 
 
+def _cut_str_to_real_char(s, max_chars):
+    result = ""
+    length = 0
+
+    for char in s:
+        b = bytes(char, "utf-8")
+        length += len(b) // 2 + (len(b) % 2 > 0)
```
Inokinoki, line 82: Do you have a reference for this? UTF-8 could have 1-4 bytes. From my understanding of the code:
- for 1 byte (generally ASCII): `length += 0 + 1`
- for 2 bytes: `length += 1 + 0`
- for 3 bytes: `length += 1 + 1`
- for 4 bytes: `length += 2 + 0`

How do we determine the total length?

giskard/client/giskard_client.py
Inokinoki: Thanks very much!

To confirm:
1. we have utf-8 between client/front-end and backend.
2. In Java backend, they are stored with utf-16 in `String`.
3. And to the DB, we have utf-16 for H2 and utf-8 for PostgreSQL.

I am not sure where the length is validated: is it in the DB or in the Hibernate? Could the requirement be different?

giskard/client/giskard_client.py
kevinmessiaen: Well Python count surrogate pairs has one character while Java, js, h2 and postgres count each individual character: https://en.wikipedia.org/wiki/UTF-16 

I improved the code as it wasn't 100% correct

Inokinoki (approved):
LGTM

Tested with titanic (even though without special characters or exceeded length)

kevinmessiaen:
> LGTM
> 
> Tested with titanic (even though without special characters or exceeded length)

You can try with this code:

```python
import giskard

@giskard.slicing_function(name="🏃" * 500)
def sf_long_name(row):
    return True

client = giskard.GiskardClient(
    url="http://localhost:9000",  # URL of your Giskard instance
    key="gsk-...")

sf_long_name.upload(client, "project_key")
```

I tried with different emojis and it worked (🧑‍🚒 counted as 5 characters, 👩‍👩‍👧‍👦 counted as 11, ...).

PR #1645, opened 2023-11-28
[GSK-2236] Fixing docs

rabah-khalek:
- added text generations to the test catalog
- standardizing index.rst for all main modules
- 1:1 mapping between cells and TOC
- introduced community cell


docs/open_source/integrate_tests/index.md
```diff
@@ -8,23 +8,14 @@ Once you've created a test suite for a model, integrating it into existing workf
 :hidden:
 
 ../../integrations/cicd/index
-../../integrations/cicd/pipeline.ipynb
```
Hartorn, line 11: This should be here, and cicd/index should not

docs/integrations/index.rst
```diff
@@ -0,0 +1,12 @@
+Integrations
+===============
+
+.. toctree::
+   :maxdepth: 1
+
+   cicd/index
```
Hartorn, line 7: I think it may be better to hide this toctree with :hidden; and display cards to have something overall disparate

Hartorn: It may also fix this 
image
(double integrations displayed)

docs/giskard_hub/index.rst
```diff
@@ -0,0 +1,12 @@
+Giskard Hub
+===============
+
+.. toctree::
```
Hartorn, line 4: Same as for integrations :

Should we use :hidden: here, and use cards instead ?

docs/community/index.rst
```diff
@@ -0,0 +1,9 @@
+Community
+===============
+
+.. toctree::
```
Hartorn, line 4: Same as for integrations :

Should we use :hidden: here, and use cards instead ?

docs/knowledge/index.rst
```diff
@@ -0,0 +1,9 @@
+Knowledge
+===============
+
+.. toctree::
```
Hartorn, line 4: Same as for integrations :

Should we use :hidden: here, and use cards instead ?

docs/open_source/index.rst
```diff
@@ -0,0 +1,10 @@
+Open Source
+===============
+
+.. toctree::
```
Hartorn, line 4: Same as for integrations :

Should we use :hidden: here, and use cards instead ?

Hartorn (commented):
```

checking consistency... /home/bazire/giskard/docs/community/index.rst: WARNING: document isn't included in any toctree
/home/bazire/giskard/docs/giskard_hub/index.rst: WARNING: document isn't included in any toctree
/home/bazire/giskard/docs/knowledge/index.rst: WARNING: document isn't included in any toctree
/home/bazire/giskard/docs/open_source/index.rst: WARNING: document isn't included in any toctree

````

Could you fix these warnings also ?

Hartorn (changes requested):
forgot to tick that

docs/open_source/integrate_tests/index.md
rabah-khalek: to be conform with the others, it should be cicd/index in both the toc and the cell. I didn't like the idea of pointing to a notebook from here.

docs/integrations/index.rst
rabah-khalek: For this, I display now only the child pages (Github, Execute, etc.)

docs/community/index.rst
rabah-khalek: done

docs/knowledge/index.rst
rabah-khalek: done

docs/open_source/index.rst
rabah-khalek: done

rabah-khalek:
@Hartorn do you have an idea on how to silence these warnings in case I don't want to include some indexes in tocs (the reason being, they're alone in the category)

Hartorn:
> @Hartorn do you have an idea on how to silence these warnings in case I don't want to include some indexes in tocs (the reason being, they're alone in the category)

I would add them to an octree but with :hidden:

Also, if they are not in any octree, they are not accessible, except if you have some links in card. So i would create the hidden octree at the same place maybe

Hartorn:
Another small issue : highlighted part is always in integrations
image
image

Hartorn:
Also this is not coherent
image

Hartorn:
> Also this is not coherent image

fixed

Hartorn (approved):

Issue #1601, opened 2023-11-10
Feature: add data quality tests in Giskard


luca-martial:
## 🚀 Feature Request

Giskard currently focuses on model quality testing, but since ML models are heavily dependent on the data they are trained on, data quality testing is of high interest. We are looking to implement various data quality tests and are open to community contributions.

Examples of tests to add:

**1. Data Completeness Test**
- **Description:** Checks for missing values in the dataset.
- **Implementation Hint:** Calculate the percentage of missing values in each column and flag columns exceeding a threshold.

**2. Data Uniqueness Test**
- **Description:** Ensures no duplicate entries in the dataset.
- **Implementation Hint:** Detect and report duplicate rows or values in specified columns.

**3. Data Range and Validity Test**
- **Description:** Ensures numerical or categorical data falls within expected ranges or sets of values.
- **Implementation Hint:** Check if data values are within specified ranges or lists of valid values.

**4. Data Correlation Test**
- **Description:** Analyzes correlations between different features.
- **Implementation Hint:** Calculate and report the correlation matrix of the dataset's features.

**5. Data Anomaly Detection Test**
- **Description:** Identifies outliers or anomalies in the dataset.
- **Implementation Hint:** Use statistical methods or anomaly detection algorithms to flag significant deviations.

**6. Data Integrity Test**
- **Description:** Ensures relationships between different data tables or datasets are maintained.
- **Implementation Hint:** Check for foreign key relationships and cross-references.

**7. Label Consistency Test**
- **Description:** Checks that labels are consistent and correctly assigned.
- **Implementation Hint:** Audit and validate label assignments.

**8. Class Imbalance Test**
- **Description:** Assesses the distribution of classes in classification problems.
- **Implementation Hint:** Calculate and report the proportion of each class.

**9. Feature Importance Test**
- **Description:** Evaluates the relevance of each feature to the target variable.
- **Implementation Hint:** Use feature importance scores or coefficients to rank features.

**10. Label Noise Detection Test**
- **Description:** Detects errors in the labeling of data.
- **Implementation Hint:** Use anomaly detection or clustering to identify mislabeled data points.

## 🔈 Motivation

This will enhance the completeness of Giskard's testing capabilities.

Kranium2002:
Hi, I would like to work on this issue

Kranium2002:
I had a question, should I work in giskard/utils and create a file for data quality tests for user? User will pass a pandas df and then the system will check its quality. How does this sound?

[...]

PR #1651, opened 2023-12-01
Add data quality tests

Kranium2002:
## Description

This pull request addresses a crucial aspect of machine learning model development by introducing a comprehensive suite of data quality tests to Giskard. While Giskard currently excels in model quality testing, enhancing its capabilities to evaluate the quality of training data is imperative.

List of tests to be added:
- [x] Data Completeness Test
- [x] Data Uniqueness Test
- [x] Data Range Test
- [x] Validity Test
- [x] Data Correlation Test
- [x] Data Anomaly Detection Test
- [x] Data Integrity Test
- [x] Label Consistency Test
- [x] Class Imbalance Test
- [x] Feature Importance Test
- [x] Label Noise Detection Test

## Related Issue

This PR is related to issue #1601 

## Type of Change

- [ ] 🔐 Security fix

## Checklist


Kranium2002:
I think some of the tests can be removed from this list like Data Consistency Test as Giskard already tests this when creating a Giskard dataset. What do you think @kevinmessiaen ?

giskard/testing/tests/data_quality.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    completeness_scores = {}
+    for column in dataset.df.columns:
+        column_data = dataset.df[column]
+        completeness_ratio = len(column_data.dropna()) / len(column_data)
+        completeness_scores[column] = completeness_ratio
+    return completeness_scores
+
+@test(name="Data Range and Validity Test")
+def range_validity_test(dataset: Dataset, column: str, min_value=None, max_value=None, valid_values=None):
```
kevinmessiaen, line 43: In my opinion it would make more sense to split into two tests: 

- `range_validity_test(dataset: Dataset, column: str, min_value: Optional[float] = None, max_value: Optional[float] = None)`
- `allowed_values_validity_test(dataset: Dataset, valid_values: Sequence[Any])`.

giskard/testing/tests/data_quality.py
```diff
@@ -0,0 +1,64 @@
...
+    Returns:
+        TestResult: The result of the test.
+    """
+    column_data = dataset.df[column]
+    if valid_values is not None:
+        # Check if all values are in the list of valid values
+        test_passed = all(x in valid_values for x in column_data.dropna())
+    else:
+        # Check if all values are within the specified range
+        test_passed = all(min_value <= x <= max_value for x in column_data.dropna())
```
kevinmessiaen, line 63: We might want to check for only `min_value`, only `max_value` and both:

```suggestion
        if min_value is not None and max_value is not None:
            test_passed = all(min_value <= x <= max_value for x in column_data.dropna())
        elif min_value is not None:
            test_passed = all(min_value <= x for x in column_data.dropna())
        elif max_value is not None:
            test_passed = all(x <= max_value for x in column_data.dropna())
        else:
            raise ValueError("Neither min_value nor max_value were provided")
```

giskard/testing/tests/data_quality.py
```diff
@@ -0,0 +1,64 @@
+"""
+Module for data quality tests.
+"""
+from giskard.ml_worker.testing.test_result import TestResult
+from giskard.ml_worker.testing.registry.decorators import test
+from giskard.datasets.base import Dataset
+
+@test(name="Data Uniqueness Test")
+def uniqueness_test(dataset: Dataset, column: str):
```
kevinmessiaen, line 9: We might want to have a threshold for flexibility

```suggestion
def uniqueness_test(dataset: Dataset, column: str, threshold: float = 0.8):
```

giskard/testing/tests/data_quality.py
```diff
@@ -0,0 +1,64 @@
...
+    Args:
+        dataset (Dataset): The dataset to test.
+        column (str): The column to check for uniqueness.
+
+    Returns:
+        TestResult: The result of the test.
+    """
+    column_data = dataset.df[column]
+    uniqueness_ratio = len(column_data.unique()) / len(column_data)
+    return TestResult(passed=uniqueness_ratio == 1)
```
kevinmessiaen, line 22: Nit picking here but in Giskard you can put a metric, this way when a test fail we can easily know what was our actual score

```suggestion
    return TestResult(passed=uniqueness_ratio == 1, metric=uniqueness_ratio)
```

giskard/testing/tests/data_quality.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Returns:
+        dict: A dictionary with the completeness score for each column.
+    """
+    completeness_scores = {}
+    for column in dataset.df.columns:
+        column_data = dataset.df[column]
+        completeness_ratio = len(column_data.dropna()) / len(column_data)
+        completeness_scores[column] = completeness_ratio
+    return completeness_scores
```
kevinmessiaen, line 40: Giskard test must return a `TestResult`:  https://docs.giskard.ai/en/latest/open_source/customize_tests/test_model/index.html#create-execute-your-own-test

kevinmessiaen (changes requested):
Thanks for the tests and your contribution @Kranium2002 ! 

I left a few comments, let me knows if you need some help.

kevinmessiaen:
> I think some of the tests can be removed from this list like Data Consistency Test as Giskard already tests this when creating a Giskard dataset. What do you think @kevinmessiaen ?

You're right about Data Consistency.

@luca-martial Do you have something specific in mind about it or can we remove this type of test?

luca-martial:
@kevinmessiaen I removed Data Consistency!

giskard/testing/tests/data_quality.py
Kranium2002: I want to return a dictionary, which has completeness ratio for each column. Should I just add a parameter for column name and a threshold for completeness ratio. Then the test returns if it passed or not with the metric as the completeness ratio.

Kranium2002:
For anomaly detection I was thinking about using z-score what do you suggest @kevinmessiaen ?

Kranium2002:
I am having some difficulty in implementing Label Noise Detection test.

I tried using isolation forests for this and they work pretty good for the sample data you provided but it starts failing as soon as we give more than one inconsistent label and also it gives false negatives when there are no errors in the labels. @kevinmessiaen

Kranium2002:
I guess if we have a large sample then we may get good results using isolation forests.

Kranium2002:
@kevinmessiaen I have added all the tests. Would you please review my PR?

giskard/testing/tests/data_quality.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    # Convert classes to strings and calculate the class proportions
+    class_counts = Counter(dataset.df[target_column].astype(str))
+    total_count = len(dataset.df)
+    class_proportions = {cls: count / total_count for cls, count in class_counts.items()}
+
+    # Create a message containing the class proportions
+    message = f"Class proportions: \n{class_proportions}"
+
+    return TestResult(passed=True,
```
kevinmessiaen, line 296: This test always return `True`, we should probably add some check to ensure we don't have over or under represented classes. 

Maybe we can introduce some thresholds

giskard/testing/tests/data_quality.py
```diff
@@ -0,0 +1,299 @@
...
+    Returns:
+        TestResult: The result of the test,
+        containing the correlation between the two columns
+        (if provided) or the full correlation matrix.
+    """
+    correlation_matrix = dataset.df.corr()
+
+    if column1 is not None and column2 is not None:
+        correlation = dataset.df[[column1, column2]].corr().iloc[0, 1]
+        return TestResult(passed=True,
```
kevinmessiaen, line 113: We always return `True`, we might have a param `should_correlate` to do this check.

Example cases:
- We find a correlation between sexe feature and higher salary which indicate some bias in our dataset. 
- We find correlation between the house area and it's price which is totally expected.

giskard/testing/tests/data_quality.py
```diff
@@ -0,0 +1,299 @@
...
+    # Train the Random Forest model
+    model = RandomForestClassifier()
+    model.fit(x, y)
+    # Get the feature importances
+    importances = model.feature_importances_
+    feature_importances = dict(zip(features, importances))
+    # Create a message containing the feature importances
+    message = f"Feature importances: \n{feature_importances}"
+
+    return TestResult(passed=True, metric_name="feature_importance",
```
kevinmessiaen, line 273: Same here, we should implement some kind of logic in order to decide whenever the test pass or fail

kevinmessiaen (changes requested):
Thanks for the great work, some tests always pass which should not be the case.

It would be nice also if you can add some unit tests (I can help you with that if you need), we use pytest and you can find example under `tests/testing/*.py`

giskard/testing/tests/data_quality.py
Kranium2002: @kevinmessiaen Could you re-assign this PR to me? I would then continue with the work.

kevinmessiaen (commented):
That's really good!

Just a small comment. I saw that you removed the `@test` annotation.

Currently the `@test` annotation generate a `GiskardTest` instance that need to be initialized and then executed as following:

```python
@test
def example_test(will_pass: bool):
  return TestResult(passed=will_pass)

result = example_test(True).execute()
assert result.passed
```

Kranium2002:
> That's really good!
> 
> Just a small comment. I saw that you removed the `@test` annotation.
> 
> Currently the `@test` annotation generate a `GiskardTest` instance that need to be initialized and then executed as following:
> 
> ```python
> @test
> def example_test(will_pass: bool):
>   return TestResult(passed=will_pass)
> 
> result = example_test(True).execute()
> assert result.passed
> ```

While using @test in the tests I was having trouble with writing unit tests for the same. It was giving errors and warnings that we need to call .execute in unit tests, but when I called .execute it still gave the same errors.

Could you give me a sample on how to write a unit test for the data quality tests with @test annotation? @kevinmessiaen

Kranium2002:
Added the @ test decorator back. @kevinmessiaen

kevinmessiaen (approved):
Thank you so much for your contribution 🎉 

I've done some modification to have consistency and improve debugging

PR #1661, opened 2023-12-05
[GSK-2292 GSK-2293] Add args kwargs to the public methods of the model and all

kevinmessiaen:
## Description

- Add `args` and `kwargs` to the public methods of the model class and subclass
- Add backward compatibility tests

## Type of Change

- [ ] 🔐 Security fix


tests/models/test_function_model.py
```diff
@@ -75,20 +76,48 @@ def test_single_feature():
     "3.10": ["3.9", "3.10"],
     "3.11": ["3.11"],
 }
+PYTHON_MAJOR_VERSION = ".".join(platform.python_version_tuple()[:2])
+BACKWARD_COMPATIBILITY_MODEL_VERSIONS = {"3.9": "3.10", "3.10": "3.10", "3.11": "3.11"}
```
Inokinoki, line 80: How do we ensure the backward compatibility here?

Inokinoki (approved):
LGTM!

My only concern is on the unused args and kwargs, which impacts the code smell.
But I think we have to do it if without a better solution...

kevinmessiaen:
> LGTM!
> 
> My only concern is on the unused args and kwargs, which impacts the code smell. But I think we have to do it if without a better solution...

Let me change to `def method(..., *_args, **_kwargs)` to prevent code smells

PR #1664, opened 2023-12-06
Add language support in LLM generators [GSK-2152]

pierlj:
## Description

Add multilanguages support inside the scanner for LLMs. Prompt the generator model to output evaluation queries in specified languages.

## Related Issue

GSK-2152

## Type of Change

- Adds a method inside the Dataset class to extract languages from the 'text' columns of the datasets. 
- Add language requirement inside the generator prompts.


giskard/datasets/base/__init__.py
```diff
@@ -507,6 +507,19 @@ def upload(self, client: GiskardClient, project_key: str):
             )
         return dataset_id
 
+    def extract_languages(self):
+        """
+        Extracts all languages present in the dataset 'text' column.
+
+        Returns:
+            list[str]: a list of language codescontaining all languages in the dataset.
```
mattbit, line 515: typo. What is language code? There are several standards. In our dataset meta, it's the [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1), maybe we can clarify it here.

giskard/datasets/base/__init__.py
```diff
@@ -507,6 +507,19 @@ def upload(self, client: GiskardClient, project_key: str):
...
+        """
+        languages_set = []
```
mattbit, line 517: Naming? `language_set` is a `list`. This may be a bit confusing. You can actually use directly a `set`.
Also, no need to include the type in the name of the variable, the code should be explicit enough:
```py3
languages = set()
```

giskard/datasets/base/__init__.py
```diff
@@ -507,6 +507,19 @@ def upload(self, client: GiskardClient, project_key: str):
...
+        for col in self.columns:
+            if self.column_types[col] == "text":
+                languages_set = languages_set + self.column_meta[col, self.column_types[col]]['language'].unique().tolist()
```
mattbit: May want to make this a bit more idiomatic, avoiding the for/if nesting and list appending.

E.g. you can get all langs
```py3
langs_per_feature = [
    dataset.column_meta[col, "text"]["language"].dropna().unique()
    for col, col_type in dataset.column_types.items()
    if col_type == "text"
]

return set().union(*langs_per_feature)
```

P.S. `nan` are not handled in your code.

giskard/llm/generators/base.py
```diff
@@ -31,10 +32,12 @@ def __init__(
         llm_temperature: Optional[float] = None,
         llm_client: LLMClient = None,
         prompt: Optional[str] = None,
+        languages_list: list[str] = ['en'],
```
mattbit: Better not to set a default argument with a mutable, see for example [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). In this case, the usual way is to set it as optional with a default to `None` and then use `self.languages = languages or ["en"]` (or similar).

Also, naming: no need to specify the type in the variable name. `languages: list[str]` is already very explicit.

mattbit: Maybe if there are no languages we should just avoid including the extra line in the prompt, no?

giskard/datasets/base/__init__.py
```diff
@@ -507,6 +507,19 @@ def upload(self, client: GiskardClient, project_key: str):
             )
         return dataset_id
 
+    def extract_languages(self):
```
mattbit, line 510: We may want to specify the features we want to consider. In some cases, the model only uses a subset of the columns.

pierlj: Yes it makes sense, I added a 'columns' argument (maybe you want to name it 'features'?) and I filter the dataset columns with it.

giskard/llm/generators/base.py
```diff
@@ -31,11 +33,16 @@ def __init__(
...
         llm_client: LLMClient = None,
         prompt: Optional[str] = None,
+        languages: list[str] = None,
     ):
         self.llm_temperature = llm_temperature if llm_temperature is not None else self._default_temperature
         self.llm_client = llm_client or get_default_client()
+        self.languages = languages if languages is not None else []
         self.prompt = prompt if prompt is not None else self._default_prompt
 
+        if isinstance(self.languages, list) and self.languages != []:
```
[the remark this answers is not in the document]
pierlj: I am not sure about this, we need to check that self.languages is actually a list otherwise it won't generate a correct prompt. I would keep it like this: 
```
if self.languages and isinstance(self.languages, list):
```

giskard/llm/generators/base.py
```diff
@@ -31,11 +33,16 @@ def __init__(
...
+            self.prompt = self.prompt + LANGUAGE_REQUIREMENT_PROMPT.format(languages=self.languages)
```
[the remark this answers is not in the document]
pierlj: I agree it would be better, but since some of the `Generator` override the `_make_generate_input_prompt` method it will require duplicating this in all these classes. Should I proceed anyway and move it to  `_make_generate_input_prompt` ?

giskard/scanner/llm/llm_basic_sycophancy_detector.py
```diff
@@ -78,7 +78,8 @@ def get_cost_estimate(self, model: BaseModel, dataset: Dataset) -> dict:
 
     def run(self, model: BaseModel, dataset: Dataset, features=None) -> Sequence[Issue]:
         # Prepare datasets
-        generator = SycophancyDataGenerator()
+        languages = dataset.extract_languages()
```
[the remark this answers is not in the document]
pierlj: Should I use `model.meta.feature_names` as argument for the `extract_languages` method or directly `features` ? 

If I understand correctly `features` from the scanner call, so it should be the intersection of the columns instead. 

Here is what I propose:

```
languages_features = list(set(model.meta.feature_names) & set(features))
languages = dataset.extract_languages(columns=languages_features)
```

mattbit (approved):

PR #1669, opened 2023-12-06
Added test_llm_ground_truth_similarity

kevinmessiaen:
## Description

Added `test_llm_ground_truth_similarity` and `test_llm_ground_truth`

## Type of Change

- [ ] 🔐 Security fix

### DIFF
--- a/giskard/testing/tests/llm/ground_truth.py
+++ b/giskard/testing/tests/llm/ground_truth.py
@@ -49,7 +49,6 @@ def test_llm_ground_truth_similarity(
         predictions=pred.prediction,
         references=dataset.df[dataset.target],
         model_type="distilbert-base-multilingual-cased",
-        idf=True,
     )
     passed = np.array(score["f1"]) > 1 - output_sensitivity
     metric = len([p for p in passed if p]) / len(passed)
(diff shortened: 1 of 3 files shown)


rabah-khalek (changes requested):
the tests look good to me @kevinmessiaen, could we just reduce code duplication in the unit-tests?

kevinmessiaen:
> the tests look good to me @kevinmessiaen, could we just reduce code duplication in the unit-tests?

It's done

giskard/testing/tests/llm/ground_truth.py
```diff
@@ -0,0 +1,58 @@
...
+        import evaluate
+    except ImportError as err:
+        raise LLMImportError() from err
+
+    scorer = evaluate.load("bertscore")
+    score = scorer.compute(
+        predictions=pred.prediction,
+        references=dataset.df[dataset.target],
+        model_type="distilbert-base-multilingual-cased",
+        idf=True,
```
rabah-khalek, line 52: > Using inverse document frequency (idf) on the reference sentences to weigh word importance may correlate better with human judgment. However, when the set of reference sentences become too small, the idf score would become inaccurate/invalid. We now make it optional. To use idf, please set --idf when using the CLI tool or idf=True when calling bert_score.score function.
> -- https://github.com/Tiiiger/bert_score

Let's keep it to default (False). This way it would also be consistent with the evaluation we do in `LLMCharInjector`

rabah-khalek (changes requested):
the unit-tests looks way better now.

sorry I didn't notice the idf parameter in my first review.

PR #1671, opened 2023-12-06
[GSK-1279] Fisher's exact test and permutation test for slice metrics significance testing

luca-rossi:
Added Fisher's exact test and permutation test for slice metrics significance testing. Todo: unit tests.

### DIFF
--- a/giskard/scanner/performance/metrics.py
+++ b/giskard/scanner/performance/metrics.py
@@ -15,6 +15,7 @@ class MetricResult:
     value: float
     affected_samples: int
     raw_values: Optional[np.ndarray] = None
+    ctable_values: Optional[list[list[int]]] = None
 
     @property
     def name(self):
@@ -27,6 +28,7 @@ class MetricResult:
 class PerformanceMetric(ABC):
     name: str
     greater_is_better = True
+    has_contingency_table = False
 
     @abstractmethod
     def __call__(self, model: BaseModel, dataset: Dataset) -> MetricResult:
@@ -43,7 +45,9 @@ class ClassificationPerformanceMetric(PerformanceMetric, metaclass=ABCMeta):
 
         value = self._calculate_metric(y_true, y_pred, model)
         num_affected = self._calculate_affected_samples(y_true, y_pred, model)
-        return MetricResult(self, value, num_affected)
+        ctable_values = self._calculate_ctable_values(value, num_affected) if self.has_contingency_table else None
+
+        return MetricResult(self, value, num_affected, ctable_values=ctable_values)
 
     @abstractmethod
     def _calculate_metric(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel) -> MetricResult:
@@ -52,10 +56,16 @@ class ClassificationPerformanceMetric(PerformanceMetric, metaclass=ABCMeta):
     def _calculate_affected_samples(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel) -> int:
         return len(y_true)
 
+    def _calculate_ctable_values(self, value, num_affected) -> list[list[int]]:
+        x = round(value * num_affected)
+        y = num_affected - x
+        return [x, y]
+
 
 class Accuracy(ClassificationPerformanceMetric):
     name = "Accuracy"
     greater_is_better = True
+    has_contingency_table = True
 
     def _calculate_metric(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel):
         return sklearn.metrics.accuracy_score(y_true, y_pred)
@@ -64,6 +74,7 @@ class Accuracy(ClassificationPerformanceMetric):
 class BalancedAccuracy(ClassificationPerformanceMetric):
     name = "Balanced Accuracy"
     greater_is_better = True
+    has_contingency_table = False
 
     def _calculate_metric(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel):
         return sklearn.metrics.balanced_accuracy_score(y_true, y_pred)
@@ -89,6 +100,7 @@ class SklearnClassificationScoreMixin:
 class F1Score(SklearnClassificationScoreMixin, ClassificationPerformanceMetric):
     name = "F1 Score"
     greater_is_better = True
+    has_contingency_table = False
     _sklearn_metric = "f1_score"
 
     def _calculate_affected_samples(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel) -> int:
@@ -104,6 +116,7 @@ class F1Score(SklearnClassificationScoreMixin, ClassificationPerformanceMetric):
 class Precision(SklearnClassificationScoreMixin, ClassificationPerformanceMetric):
     name = "Precision"
     greater_is_better = True
+    has_contingency_table = True
     _sklearn_metric = "precision_score"
 
     def _calculate_affected_samples(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel) -> int:
@@ -116,6 +129,7 @@ class Precision(SklearnClassificationScoreMixin, ClassificationPerformanceMetric
 class Recall(SklearnClassificationScoreMixin, ClassificationPerformanceMetric):
     name = "Recall"
     greater_is_better = True
+    has_contingency_table = True
     _sklearn_metric = "recall_score"
 
     def _calculate_affected_samples(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel) -> int:
@@ -128,6 +142,7 @@ class Recall(SklearnClassificationScoreMixin, ClassificationPerformanceMetric):
 class AUC(PerformanceMetric):
     name = "ROC AUC"
     greater_is_better = True
+    has_contingency_table = False
 
     def __call__(self, model: BaseModel, dataset: Dataset) -> MetricResult:
         y_true = dataset.df[dataset.target]
--- a/giskard/scanner/performance/metrics.py
+++ b/giskard/scanner/performance/metrics.py
@@ -15,7 +15,7 @@ class MetricResult:
     value: float
     affected_samples: int
     raw_values: Optional[np.ndarray] = None
-    ctable_values: Optional[list[list[int]]] = None
+    binary_counts: Optional[list[int]] = None
 
     @property
     def name(self):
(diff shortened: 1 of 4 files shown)


giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -303,36 +303,88 @@ def _is_unbalanced_target(classes: pd.Series):
...
+def _calculate_pvalue_from_permutation_test(
+    slice_dataset, comp_dataset, dataset, model, metric, perm_test_resamples=1000
+):
+    logger.info("PerformanceBiasDetector: permutation test")
+
+    def statistic(slice_ids, comp_ids):
+        perm_slice_dataset = Dataset(
+            dataset.df.loc[slice_ids],
+            target=dataset.target,
+            column_types=dataset.column_types.copy(),
```
mattbit, line 332: I don't think we need to make a copy here

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -303,36 +303,88 @@ def _is_unbalanced_target(classes: pd.Series):
...
+        logger.info("PerformanceBiasDetector: Fisher's exact test")
+        return scipy.stats.fisher_exact(ctable, alternative="two-sided")[1]
+    logger.info("PerformanceBiasDetector: chi-square test")
+    return scipy.stats.chi2_contingency(ctable, correction=False, lambda_="log-likelihood")[1]
+
+
+def _calculate_pvalue_from_permutation_test(
+    slice_dataset, comp_dataset, dataset, model, metric, perm_test_resamples=1000
+):
+    logger.info("PerformanceBiasDetector: permutation test")
```
mattbit, line 326: let’s keep this as debug

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -303,36 +303,88 @@ def _is_unbalanced_target(classes: pd.Series):
...
+    slice_y_cnt = slice_metric.affected_samples - slice_x_cnt
+
+    comp_x_cnt = round(comp_metric.value * comp_metric.affected_samples)
+    comp_y_cnt = comp_metric.affected_samples - comp_x_cnt
+
+    ctable = [[slice_x_cnt, slice_y_cnt], [comp_x_cnt, comp_y_cnt]]
+
+    # if the slice size is too small, use Fisher's exact test, otherwise use a chi-square test
+    if min(min(row) for row in ctable) <= max_size_fisher:
+        logger.info("PerformanceBiasDetector: Fisher's exact test")
```
mattbit, line 317: logger.debug

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -303,36 +303,88 @@ def _is_unbalanced_target(classes: pd.Series):
...
+def _calculate_pvalue_from_contingency_table(slice_metric, comp_metric, max_size_fisher=30):
+    slice_x_cnt = round(slice_metric.value * slice_metric.affected_samples)
+    slice_y_cnt = slice_metric.affected_samples - slice_x_cnt
+
+    comp_x_cnt = round(comp_metric.value * comp_metric.affected_samples)
+    comp_y_cnt = comp_metric.affected_samples - comp_x_cnt
+
+    ctable = [[slice_x_cnt, slice_y_cnt], [comp_x_cnt, comp_y_cnt]]
+
+    # if the slice size is too small, use Fisher's exact test, otherwise use a chi-square test
```
mattbit, line 315: It‘s a G-test

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -303,36 +303,88 @@ def _is_unbalanced_target(classes: pd.Series):
...
     try:
         # If we have raw values for the metric, we perform a standard t-test
+        logger.info(f"PerformanceBiasDetector: metric name = {slice_metric.name}")
         if slice_metric.raw_values is not None:
+            logger.info("PerformanceBiasDetector: t-test")
             alternative = "less" if metric.greater_is_better else "greater"
             _, pvalue = scipy.stats.ttest_ind(
                 slice_metric.raw_values, comp_metric.raw_values, equal_var=False, alternative=alternative
             )
+        elif metric.name.lower() in ["accuracy", "precision", "recall"]:
```
mattbit, line 377: We should do this check differently. Let’s use some attribute on the metric object, like is_binary_metric or similar.

mattbit: maybe it's even better to let the Metric class the responsibility of calculating the contingency table entries

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -303,36 +303,88 @@ def _is_unbalanced_target(classes: pd.Series):
...
-
-            pvalue = scipy.stats.chi2_contingency(ctable, lambda_="log-likelihood")[1]
-    except ValueError:
+            # if the the contingency table cannot be calculated, do a permutation test
+            pvalue = _calculate_pvalue_from_permutation_test(
+                slice_dataset, comp_dataset, dataset, model, metric, perm_test_resamples
+            )
+    except ValueError as err:
         pvalue = np.nan
+        logger.info(f"PerformanceBiasDetector: p-value could not be calculated: {err}")
```
mattbit, line 387: logger.debug

mattbit (commented):
Few minor comments but otherwise looks good to me.

giskard/scanner/performance/metrics.py
```diff
@@ -52,10 +56,16 @@ def _calculate_metric(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseM
     def _calculate_affected_samples(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseModel) -> int:
         return len(y_true)
 
+    def _calculate_ctable_values(self, value, num_affected) -> list[list[int]]:
```
mattbit, line 59: The return type does not match. Also, this is not a contingency table?

giskard/scanner/performance/metrics.py
```diff
@@ -15,6 +15,7 @@ class MetricResult:
     value: float
     affected_samples: int
     raw_values: Optional[np.ndarray] = None
+    ctable_values: Optional[list[list[int]]] = None
```
mattbit, line 18: We can't have a contingency table

mattbit: We can be more general and maybe just get a categorical or binary representation?
Something like `binary_counts` or `binary_representation`?

giskard/scanner/performance/metrics.py
```diff
@@ -52,10 +56,16 @@ def _calculate_metric(self, y_true: np.ndarray, y_pred: np.ndarray, model: BaseM
...
+    def _calculate_ctable_values(self, value, num_affected) -> list[list[int]]:
+        x = round(value * num_affected)
+        y = num_affected - x
+        return [x, y]
+
 
 class Accuracy(ClassificationPerformanceMetric):
     name = "Accuracy"
     greater_is_better = True
+    has_contingency_table = True
```
mattbit, line 68: Same as above, this could be more general. In any case, I’m not sure it's needed if we have it on the result object.

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -303,36 +303,82 @@ def _is_unbalanced_target(classes: pd.Series):
...
+    slice_dataset, comp_dataset, dataset, model, metric, perm_test_resamples=1000
+):
+    logger.debug("PerformanceBiasDetector: permutation test")
+
+    def statistic(slice_ids, comp_ids):
+        perm_slice_dataset = Dataset(
+            dataset.df.loc[slice_ids],
+            target=dataset.target,
+            # column_types=dataset.column_types.copy(),
+            # validation=False,
```
mattbit, line 327: why is this commented?

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -303,36 +303,82 @@ def _is_unbalanced_target(classes: pd.Series):
...
+        )
+        perm_comp_dataset = Dataset(
+            dataset.df.loc[comp_ids],
+            target=dataset.target,
+            # column_types=dataset.column_types.copy(),
+            # validation=False,
```
mattbit, line 333: same

tests/scan/test_performance_bias_detector.py
```diff
@@ -92,67 +92,102 @@ def test_selects_issues_with_benjamini_hochberg(titanic_model, titanic_dataset):
...
-    detector = PerformanceBiasDetector(alpha=1e-10)
+    detector = PerformanceBiasDetector(alpha=1e-15)
 
     issues = detector.run(titanic_model, titanic_dataset, features=["Name", "Sex", "Pclass"])
     assert len(issues) == 2
 
 
 def test_calculate_slice_metrics():
+    SLICE_SIZE = 500
+    np.random.seed(42)
```
mattbit, line 103: should have the seed somewhere in the detector, not setting the global seed here

mattbit (changes requested):
Sorry, some small changes

giskard/scanner/performance/metrics.py
luca-rossi: That's why I called ctable_values but I get that it could be confusing, I couldn't come up with a better name. Maybe something like ctable_row?

I forgot to update the return type.

giskard/scanner/performance/metrics.py
luca-rossi: Same as above

giskard/scanner/performance/metrics.py
luca-rossi: It's not needed but I added it for clarity and "just in case". Should I remove it?

giskard/scanner/performance/performance_bias_detector.py
luca-rossi: Forgot to remove it

tests/scan/test_performance_bias_detector.py
luca-rossi: Yes, not sure why I put it there

giskard/scanner/performance/metrics.py
luca-rossi: Or binary_counts like in your comment below

giskard/scanner/performance/metrics.py
luca-rossi: Actually, it's needed because it's used to decide whether to calculate the binary counts (to prevent from calculating them for classification metrics where it wouldn't make sense, e.g. the F1-Score)

mattbit (approved):
Looks good, I’ll do some local testing.

PR #1676, opened 2023-12-12
GSK-2368 Recreate docker container on start

andreybavt:


giskard/commands/cli_hub.py
```diff
@@ -144,19 +144,21 @@ def _start(attached=False, skip_version_check=False, version=None, environment=N
...
-            ports={7860: port},
-            volumes={home_volume.name: {"bind": "/home/giskard/datadir", "mode": "rw"}},
-            environment=environment,
-        )
-    container.start()
+    if container:
+        logger.info(f"Removing old container {container.name}")
+        container.remove()
+
+    logger.info(f"Running a container {container.name}")
```
kevinmessiaen, line 153: This break when no container was previously running

```suggestion
```

kevinmessiaen (changes requested):

giskard/commands/cli_hub.py
```diff
@@ -144,19 +144,21 @@ def _start(attached=False, skip_version_check=False, version=None, environment=N
...
-            detach=not attached,
-            name=get_container_name(version),
-            ports={7860: port},
-            volumes={home_volume.name: {"bind": "/home/giskard/datadir", "mode": "rw"}},
-            environment=environment,
-        )
-    container.start()
+    if container:
+        logger.info(f"Removing old container {container.name}")
+        container.remove()
```
kevinmessiaen, line 151: When this container is still running the following error appears: 

`docker.errors.APIError: 409 Client Error for http+docker://localhost/v1.42/containers/23834e52622066f0e2cfa373c98f6e5c89148e45f304d19bcc659ecd58731fd8?v=False&link=False&force=False: Conflict ("You cannot remove a running container 23834e52622066f0e2cfa373c98f6e5c89148e45f304d19bcc659ecd58731fd8. Stop the container before attempting removal or force remove")`

We should probably just show the following message instead:
`Giskard Hub is currently running, please run "giskard hub stop" to stop it`

kevinmessiaen (approved):

PR #1680, opened 2023-12-14
[GSK-2378] Only persist/read cache when explicitly asked

kevinmessiaen:
## Description

- Only persist/read cache when explicitly asked
- Only load artifacts onces per test suite execution (model therefore share cache for a suite)
- For scan since only one model is loaded cache is also shared

## Type of Change

- [ ] 🔐 Security fix

## Checklist

### DIFF
--- a/giskard/ml_worker/websocket/listener.py
+++ b/giskard/ml_worker/websocket/listener.py
@@ -520,17 +520,20 @@ def run_test_suite(
     client: Optional[GiskardClient], params: websocket.TestSuiteParam, *args, **kwargs
 ) -> websocket.TestSuite:
     log_listener = LogListener()
+
+    loaded_artifacts = dict()
+
     try:
         tests = [
             {
                 "test": GiskardTest.download(t.testUuid, client, None),
-                "arguments": parse_function_arguments(client, t.arguments),
+                "arguments": parse_function_arguments(client, t.arguments, loaded_artifacts),
                 "id": t.id,
             }
             for t in params.tests
         ]
 
-        global_arguments = parse_function_arguments(client, params.globalArguments)
+        global_arguments = parse_function_arguments(client, params.globalArguments, loaded_artifacts)
 
         datasets = {arg.original_id: arg for arg in global_arguments.values() if isinstance(arg, Dataset)}
         for test in tests:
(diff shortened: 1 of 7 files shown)


giskard/ml_worker/websocket/listener.py
```diff
@@ -520,17 +520,20 @@ def run_test_suite(
     client: Optional[GiskardClient], params: websocket.TestSuiteParam, *args, **kwargs
 ) -> websocket.TestSuite:
     log_listener = LogListener()
+
+    loaded_artifacts = dict()
```
Googleton, line 524: Why is it initialized here if it's also set in the function with the 'if none' condition ?

giskard/ml_worker/websocket/utils.py
```diff
@@ -158,7 +158,24 @@ def map_dataset_process_function_meta_ws(callable_type):
     }
 
 
-def parse_function_arguments(client: Optional[GiskardClient], request_arguments: List[websocket.FuncArgument]):
+def _get_or_load(loaded_artifacts: Dict[str, Dict[str, Any]], type: str, uuid: str, load_fn: Callable[[], Any]) -> Any:
+    if type not in loaded_artifacts:
+        loaded_artifacts[type] = dict()
```
Hartorn, line 163: This could be avoided by giving a [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) instead of a dict, with dict as default factory

giskard/ml_worker/websocket/listener.py
Hartorn: it' done to share the cache between the global arguments and test ones

giskard/models/cache/cache.py
```diff
@@ -26,14 +26,23 @@ def flatten(xs):
...
+        cache_dir: Optional[Path] = None,
+    ):
         self.id = id
         self.prediction_cache = dict()
 
-        if cache_dir is None and self.id:
-            cache_dir = self._default_cache_dir_prefix.joinpath(self.id)
+        if persist_cache:
+            if cache_dir is None and self.id:
+                cache_dir = self._default_cache_dir_prefix.joinpath(self.id)
```
Hartorn, line 41: Could be self._default_cache_dir_prefix / self.id

Hartorn (commented):
Looks good, some small comment. 
Did not test , and just looked quickly

Hartorn (approved):

PR #1694, opened 2023-12-21
Added tool calls support for LLM clients

kevinmessiaen:
## Description

Added tool calls support for LLM clients, migrate function call to tool calls

## Type of Change

- [ ] 🔐 Security fix


giskard/llm/evaluators/base.py
```diff
@@ -90,18 +93,18 @@ def evaluate(self, model: BaseModel, dataset: Dataset):
...
                     [{"role": "system", "content": prompt}],
-                    functions=funcs,
-                    function_call={"name": "evaluate_model"},
+                    tools=funcs,
+                    tool_choice={"type": "function", "function": {"name": "evaluate_model"}},
                     temperature=self.llm_temperature,
                     caller_id=self.__class__.__name__,
                 )
-                if out.function_call is None or "passed_test" not in out.function_call.args:
+                if len(out.tool_calls) != 1 or "passed_test" not in out.tool_calls[0].args:
```
AbSsEnT, line 101: In this specific case, we do not expect, that we can make multiple tool calls, right?

giskard/llm/client/base.py
```diff
@@ -16,6 +16,7 @@ class LLMFunctionCall:
 class LLMOutput:
     message: Optional[str] = None
     function_call: Optional[LLMFunctionCall] = None
+    tool_calls: Sequence[LLMFunctionCall] = field(default_factory=list)
```
AbSsEnT, line 19: For this moment it is ok for a backward compatibility, but I believe, that by the end of the day we need to keep only "tool" and not "function" naming convention and thus related variables like "function_call"

giskard/llm/evaluators/base.py
kevinmessiaen: Yes since we are evaluating a single case we expect it to be evaluated once

giskard/llm/client/base.py
kevinmessiaen: Yes I'm not sure if `function_call` are being used in other PR's and might be used by some users having developed there own test.

I guess it's okay to keep `LLMFunctionCall` for now since only tool supported for now is of type `function`. I guess we will need to update it once we have .

mattbit:
@kevinmessiaen does this work with openai<1? We might need to add some version constraint in `pyproject.toml`

kevinmessiaen:
> @kevinmessiaen does this work with openai<1? We might need to add some version constraint in `pyproject.toml`

Yes, I tried with `openai==0.28.1` and it works

AbSsEnT (approved):
ok

PR #1732, opened 2024-01-11
feat: set __str__ to use name and id for Dataset and Model

AdriMarteau:
## Description

Define `__str__` for the Dataset and Model classes for more user friendly prints and reports

## Related Issue

[Original discussion](https://github.com/orgs/Giskard-AI/discussions/1723) 

## Type of Change

- [ ] 🔐 Security fix

## Checklist

- [x] I've updated the code style using `make codestyle`.


kevinmessiaen (changes requested):
Thanks for the work, I only have one remark is that the type of object is not visible:
`test-b55675f8-2ffd-4d7e-92b8-945e40be5b68` can be either a Model or a Dataset.

I believe something like that would be better:
`<giskard.Model name="test" uuid=b55675f8-2ffd-4d7e-92b8-945e40be5b68>`

AdriMarteau:
> Thanks for the work, I only have one remark is that the type of object is not visible: `test-b55675f8-2ffd-4d7e-92b8-945e40be5b68` can be either a Model or a Dataset.
> 
> I believe something like that would be better: `<giskard.Model name="test" uuid=b55675f8-2ffd-4d7e-92b8-945e40be5b68>`

I overloaded the `__str__` method for the moment. I can also add a `__repr__` method that with the details you are requesting. I found example like [this](https://stackoverflow.com/a/58514577) telling but I am happy to change my implementation if it's more relevant

kevinmessiaen:
> @kevinmessiaen I have added utest for `__repr__` to show that we have richer string vs the naming with `__str__`. But if you thing that it is better to also include the class name in `__str__`, I can add it

I don't have a strong opinion on it so we can go without it. But I'm not convinced by having  `name-uuid`, something like `name (uuid)` might be easier to read IMO

AdriMarteau:
> > @kevinmessiaen I have added utest for `__repr__` to show that we have richer string vs the naming with `__str__`. But if you thing that it is better to also include the class name in `__str__`, I can add it
> 
> I don't have a strong opinion on it so we can go without it. But I'm not convinced by having `name-uuid`, something like `name (uuid)` might be easier to read IMO

Very good point let me update this request

kevinmessiaen (approved):

PR #1743, opened 2024-01-18
[GSK-2625] Fix deprecations in tests fixtures from sklearn 1.4.0 

andreybavt:


mattbit:
Little problems with our test fixtures:

- Compatibility problem with imbalanced learn which we use in one fixture (new release should be out soon, see https://github.com/scikit-learn-contrib/imbalanced-learn/issues/1057)
- Updated the `lbfgs` solver causes some very little modifications of the fitted coefficients, in one case this affected our tests (one more reason to save the trained models as fixtures instead of rerunning the full training for each build…)

Nothing affecting the library from what I see.

mattbit:
Updated version of imbalanced-learn is now released:
https://github.com/scikit-learn-contrib/imbalanced-learn/releases/tag/0.12.0

Will re-enable our previously broken fixture

Hartorn (approved):

PR #1738, opened 2024-01-17
[GSK-2569] Scanner plugin for custom detectors

bmalezieux:
## Description

Adding a new Scanner object to handle custom models, datasets and list of detectors. This makes the integration of plugin tools easier for users.

## Type of Change

- [ ] 📚 Examples / docs / tutorials / dependencies update
- [ ] 🔧 Bug fix (non-breaking change which fixes an issue)
- [ ] 🥂 Improvement (non-breaking change which improves an existing feature)
- [x] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🔐 Security fix

## Checklist

- [x] I've read the [`CODE_OF_CONDUCT.md`](https://github.com/Giskard-AI/ai-inspector/blob/master/CODE_OF_CONDUCT.md) document.
- [x] I've read the [`CONTRIBUTING.md`](https://github.com/Giskard-AI/ai-inspector/blob/master/CONTRIBUTING.md) guide.
- [x] I've updated the code style using `make codestyle`.
- [ ] I've written tests for all new methods and classes that I created.
- [x] I've written the docstring in Google format for all the methods and classes that I used.

PR #1746, opened 2024-01-18
[GSK-2569] scan refactoring to accept external detectors

rabah-khalek:
supersedes https://github.com/Giskard-AI/giskard/pull/1738


mattbit:
Uhm… not a fan of this. I feel it adds complexity to an already bloated code (which I wrote 😅):

- now having us to maintain two scanners basically
- `BaseScanner` defining an `analyze` method
- `Scanner` ignoring the `analyze` and reimplementing its own
- want to add "extensibility", but there are no clear interfaces defined → we will break it as soon as someone touches some internal method
- what should be a thin utility function (`scan`) has now some weird logic
- lots of inheritance, little composition

My advice:
- If you need something quick, let's just implement it on the vision side (after all it's just a loop over the detectors, calling `.run()`), and you can fully reuse the widget component (just return a `ScanWidget`). You can probably isolate some extra components from the original scan if needed to access from outside.
- If you want to work on the extensibility of the scanner, let's define the interfaces and "decompose" the current bloat, favouring "composition over inheritance". I can help with this if needed (after all most of the bloat is mine).

rabah-khalek:
We already discussed in person, but let me summarize:

> now having us to maintain two scanners basically

well, there's not two scanners, there's a base one, and one that extends it with no overlap

> Scanner ignoring the analyze and reimplementing its own

Not at all, the Scanner uses the parent's class analyze, and wraps around it additional methods.

> what should be a thin utility function (scan) has now some weird logic

Not really weird, it uses all the internal detectors when it detects a Giskard Model, it uses the base scanner when it doesn't. That's a bit how mlflow does it with their Evaluator 

> If you need something quick, let's just implement it on the vision side (after all it's just a loop over the detectors, calling .run()), and you can fully reuse the widget component (just return a ScanWidget). You can probably isolate some extra components from the original scan if needed to access from outside.

Yeah, I agree, the current widgets are too text oriented perhaps, we should be able to have control over this (like Show details, and eliminate the Giskard-hub related messages). @bmalezieux will look into that.

rabah-khalek:
it's a beauty, but I'm closing it as it'll be superseded by the restructuring

PR #1748, opened 2024-01-19
GSK-2617 Missing kwargs in a test make the whole suite fail

kevinmessiaen:
### Issue

When running a test suite, we do a validation that all the parameters have been set based on the typing annotation. (type is not `Optional` and no default value has been provided. 

The problem is that the whole test Suite is cancelled if only one test is "misconfigured", furthermore it might happen that a required parameter is actually optional but the typing is wrong.

### Solution

The validation now show a warning instead of an exception and run the Suite letting the test fail.

### How to reproduce

Here is a small snippet that run a Suite with missing param

```python
import giskard
from giskard.core.test_result import TestResult

@giskard.test()
def my_test(is_pass: bool, unused: int):
  return TestResult(passed=is_pass)

suite = giskard.Suite().add_test(my_test()).add_test(my_test(is_pass=True)).add_test(is_pass=True, unused=1)
result = suite.run()
```

## Type of Change

- [ ] 🔐 Security fix


giskard/core/suite.py
```diff
@@ -401,10 +401,6 @@ def run(self, verbose: bool = True, **suite_run_args):
         run_args.update(suite_run_args)
 
         results: List[(str, TestResult, Dict[str, Any])] = list()
-        required_params = self.find_required_params()
-        undefined_params = {k: v for k, v in required_params.items() if k not in run_args}
-        if len(undefined_params):
```
andreybavt, line 406: there's the same check in `to_unittest` that should be consistent with `run`

andreybavt: why not keep the same check but convert an exception to a warning?

kevinmessiaen: Yes that's right, a warning make a lot more sense especially with live logs now

kevinmessiaen:
@andreybavt I updated the `Suite.run` and `Suite.to_unittest` to now raise a warning in case of missing params.

I removed the exception altogether since the test might pass with missing params if the type annotation is wrong

vjern (approved):
I agree with the overall solution. +1 for early validation of required params, but it should not blockade the test suite being used, and it's worthwhile to still warn the user about it.

Feel free to commit or disregard my other nitpicks.

PR #1549, opened 2023-11-04
[GSK-1955] stdout of scan

rabah-khalek:

PR #1751, opened 2024-01-21
[GSK-1955] Scan custom stdout

rabah-khalek:
Supersedes: https://github.com/Giskard-AI/giskard/pull/1549

Since the scan is our main API, it is very important to make it's output as readable as possible.
For this, a set of templates and colors are already defined (see `xprint.py` and `test_xprint.py`).

Todo:
- Wrap `xprint` into a class similar to `logging.Logger` that takes in three levels (Note that an attempt of inheriting from `Logger` was done [here](https://github.com/Giskard-AI/giskard/blob/cdc693fe77647737a3fed2845ef0f15f1a69fff2/giskard/scanner/logger.py) but wasn't really useful, let's create our own class) :
  - `INFO`: default that prints out useful logs (which detector is running, summaries, etc.)
  - `DEBUG`: default that prints out additional logs (content of prompts, failed examples, etc.)
  - `CRITICAL`: that only prints out errors and warnings if they occur
- Removing the `Catalog` in favour of putting templates per module where they are needed
- To take from the old PR: halting the user with a decision to make if they're scanning an LLM model (based on cost estimate): https://github.com/Giskard-AI/giskard/blob/cdc693fe77647737a3fed2845ef0f15f1a69fff2/giskard/scanner/scanner.py#L44-L57
- Control verbosity level globally via the class, and do NOT propagate a `verbose` parameter everywhere.
- A lot of examples to take from https://github.com/Giskard-AI/giskard/pull/1549 (prints already implemented in the detectors)


giskard/scanner/scanlogger.py
```diff
@@ -0,0 +1,65 @@
...
+    CRITICAL = 10
+
+class ScanLogger:
+    level: ScanLoggerLevel = ScanLoggerLevel.INFO
+    
+    def setLevel(self, 
+        level: ScanLoggerLevel
+    ):
+        if not isinstance(level, ScanLoggerLevel):
+            raise ValueError("Invalid log level. Please provide a ScanLoggerLevel enum.")
```
rabah-khalek, line 31: you can simply do instead:
```py
@configured_validate_arguments
def setLevel(self, level: ScanLoggerLevel):
    self.level = level
```

see https://github.com/Giskard-AI/giskard/blob/e13588dedbee34e8a661ad1ce797a6e2d40de5b0/giskard/core/validation.py#L16

giskard/scanner/scanlogger.py
```diff
@@ -0,0 +1,65 @@
...
+        self.level = level
+        
+    def checkLevel(self):
+        return self.level.name
```
rabah-khalek, line 35: Is that needed? if you just used it to debug, let's remove it

rabah-khalek (commented):
let's move `xprint.py` from `giskard/scanner` to `giskard/utils` since now you got a `scanlogger` for the scan, and we're removing the catalog, so xprint can be seen as a general utility

giskard/scanner/scanlogger.py
Sakayatsp: done!

giskard/scanner/scanlogger.py
Sakayatsp: I changed the code so that it matches strings to ScanLoggerLevel objects. It looks like this:

```
def setLevel(self, level: str):
        level_map = {
            "DEBUG": ScanLoggerLevel.DEBUG,
            "INFO": ScanLoggerLevel.INFO,
            "CRITICAL": ScanLoggerLevel.CRITICAL
        }

        if level not in level_map:
            raise ValueError(f"Invalid log level: {level}. Please provide 'DEBUG', 'INFO', or 'CRITICAL'.")
        
        self.level = level_map[level]
```

.gitignore
```diff
@@ -174,4 +174,6 @@ catboost_info
 generated
 # npm
 node_modules/
-mlruns
\ No newline at end of file
+mlruns
+example.ipynb
+logged_demo.ipynb
```
rabah-khalek, line 179: I wouldn't mention personal notebooks to gitibnore, if it help, you can use `git add -u .` to add only "tracked" files.

giskard/scanner/__init__.py
```diff
@@ -29,7 +29,7 @@ def scan(
     features: Optional[Sequence[str]] = None,
     params=None,
     only=None,
-    verbose=True,
+    level="INFO",
```
rabah-khalek, line 32: I would rather call it `verbosity_level`.

giskard/scanner/calibration/overconfidence_detector.py
```diff
@@ -55,7 +57,7 @@ def _find_issues(
         dataset_with_meta.load_metadata_from_instance(dataset.column_meta)
 
         p_threshold = self.p_threshold or _default_overconfidence_threshold(model)
-        logger.info(f"{self.__class__.__name__}: Using overconfidence threshold = {p_threshold}")
+        logger.info(p_threshold, template=ConfidenceThreshold)
```
rabah-khalek, line 60: this can be debug instead

giskard/scanner/scanlogger.py
```diff
@@ -0,0 +1,56 @@
...
+
+class ScanLogger:
+    level: ScanLoggerLevel = ScanLoggerLevel.INFO
+    
+    def setLevel(self, level: str):
+        level_map = {
+            "DEBUG": ScanLoggerLevel.DEBUG,
+            "INFO": ScanLoggerLevel.INFO,
+            "CRITICAL": ScanLoggerLevel.CRITICAL
+        }
```
rabah-khalek, line 22: No need, `Enum` allows you to do `ScanLoggerLevel[level].value`

giskard/scanner/calibration/underconfidence_detector.py
```diff
@@ -72,9 +75,7 @@ def _find_issues(
 
             relative_delta = (slice_rate - reference_rate) / reference_rate
 
-            logger.info(
-                f"{self.__class__.__name__}: Testing slice {slice_fn}\tUnderconfidence rate (slice) = {slice_rate:.3f} (global {reference_rate:.3f}) Δm = {relative_delta:.3f}"
-            )
+            logger.info(slice_fn, slice_rate, reference_rate, relative_delta, template=UnderconfTest)
```
rabah-khalek, line 78: logger.debug instead

giskard/scanner/scanlogger.py
```diff
@@ -0,0 +1,56 @@
...
+
+from giskard.utils.xprint import (
+    Template,
+    xprint,
+)
+
+class ScanLoggerLevel(enum.Enum):
+    DEBUG = 0
+    INFO = 5
+    CRITICAL = 10
```
rabah-khalek, line 12: I think you got these backward.

for instance `if level < ScanLoggerLevel.DEBUG.value` we want to show everything: DEBUG + INFO + CRITICAL for `if level < ScanLoggerLevel.INFO.value` we want to show only:  INFO + CRITICAL 
`if level < ScanLoggerLevel.CRITICAL.value` we want to show only:  CRITICAL 

see also: https://docs.python.org/3/library/logging.html#logging-levels

giskard/scanner/scanlogger.py
```diff
@@ -0,0 +1,56 @@
...
+
+class ScanLogger:
+    level: ScanLoggerLevel = ScanLoggerLevel.INFO
+    
+    def setLevel(self, level: str):
```
rabah-khalek, line 17: let's create a type for level similar to what we did here for modeltypes:
```py
class SupportedModelTypes(Enum):
    CLASSIFICATION = "classification"
    REGRESSION = "regression"
    TEXT_GENERATION = "text_generation"


ModelType = Union[SupportedModelTypes, Literal["classification", "regression", "text_generation"]]
```
in your case:
```py
class ScanLoggerLevel(enum.Enum):
    DEBUG = 0
    INFO = 5
    CRITICAL = 10


LoggerLevel = Union[ScanLoggerLevel, Literal["DEBUG", "INFO", "CRITICAL"]]
```
and use `LoggerLevel` for the typing hints

rabah-khalek (commented):
Round 2 of reviews :)

giskard/scanner/scanlogger.py
Sakayatsp: Let's say that we are currently in `debug` mode, meaning that `self.level.value = 0`.
Then when we call `logger.info`, the condition `self.level.value <= ScanLoggerLevel.INFO.value` becomes `0 <= 5`, which is true so INFO logs are shown. Same for `critical` mode.
Conversely, if we choose `critical` mode, we get `self.level.value = 10` so no conditions are satisfied except for `logger.critical`.
I think it's working as intended.

giskard/scanner/__init__.py
```diff
@@ -29,7 +29,7 @@ def scan(
     features: Optional[Sequence[str]] = None,
     params=None,
     only=None,
-    verbose=True,
+    verbosity_level="INFO",
```
rabah-khalek, line 32: now should be:
```py
verbosity_level: LoggerLevel = ScanLoggerLevel.INFO.name
```

giskard/scanner/logger.py
```diff
@@ -1,4 +1,4 @@
 import logging
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
+logger.setLevel(logging.DEBUG)
```
rabah-khalek, line 4: let's keep this to `INFO` by default

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -143,9 +167,9 @@ def detect(self, model: BaseModel, dataset: Dataset, slices: Sequence[SlicingFun
             p_values = np.array(p_values)
             p_values_rank = p_values.argsort() + 1
             p_value_threshold = p_values[p_values <= p_values_rank / len(p_values) * self.alpha].max()
-            logger.info(f"PerformanceBiasDetector: Benjamini–Hocheberg p-value threshold is {p_value_threshold:.3e}")
+            logger.info(p_value_threshold, template=BHpThreshold)
```
rabah-khalek, line 170: I think to be consistent, we should put these into `debug` no?

giskard/scanner/performance/performance_bias_detector.py
```diff
@@ -143,9 +167,9 @@ def detect(self, model: BaseModel, dataset: Dataset, slices: Sequence[SlicingFun
...
             issues = [issue for issue in issues if issue.meta.get("p_value", np.nan) <= p_value_threshold]
-            logger.info(f"PerformanceBiasDetector: Kept {len(issues)} significant issues out of {len(p_values)}.")
+            logger.info(len(issues), len(p_values), template=BHpOut)
```
rabah-khalek, line 172: same, into `debug`?

giskard/scanner/scanlogger.py
```diff
@@ -0,0 +1,59 @@
...
+
+class ScanLogger:
+    level: ScanLoggerLevel = ScanLoggerLevel.INFO
+
+    @configured_validate_arguments
+    def setLevel(self, level: LoggerLevel):
+        if isinstance(level, ScanLoggerLevel):
+            self.level = level
+        else:
+            self.level = ScanLoggerLevel[level]
```
rabah-khalek, line 26: nitpick: `self.level = level if isinstance(level, ScanLoggerLevel) else ScanLoggerLevel[level]`

tests/scan/test_xprint.py
```diff
@@ -0,0 +1,124 @@
...
+    MAGENTA_STYLE,
+    PLACEHOLDER,
+    RED_COLOR,
+    RED_STYLE,
+    RESET,
+    WHITE_COLOR,
+    WHITE_STYLE,
+    YELLOW_COLOR,
+    YELLOW_STYLE,
+    Catalog,
```
rabah-khalek, line 22: this is causing the tests the CI to fail

rabah-khalek (commented):
Overall good, some minor changes. Please fix the tests (removing `Catalog`) and let's move to the rest of the detectors :) Ping me when you want me to review again.

giskard/utils/xprint.py
```diff
@@ -0,0 +1,115 @@
...
+    return _content_template.format(*get_design_templates(template)).split(PLACEHOLDER)
+
+
+def style(*args, template: Optional[Template] = None):
+    if not template:
+        template = Template()
+    wraps = process_template(template)
+    out = []
+    for i, _style in enumerate(template.pstyles):
+        if i == template.num_styles - 1 and len(args[i:]) > 1:
```
Hartorn, line 76: not sure I follow the logic
if there are too many args, we just add the last slice all together ?

Hartorn (changes requested):
Code in xprint should be documented, as it's quite obscure as it is.
Type hinting may help in addition to explanation

Also, tests should also test limit cases (ie missing arguments to format, bad number of styles, missing reset, and so on)

Also new class scanlogger does not have any tests

mattbit (changes requested):
Can we reduce the abstractions? >400 lines of code just to print a couple of colored lines seems a lot.

I couldn't run it locally, I didn't manage to change the level.

The levels are not used appropriately, but only for visualization? The logic is scattered a bit everywhere: utils, scanner, logger, detectors.

I would simplify this drastically before merging and go with a simpler solution.

rabah-khalek:
@Sakayatsp, reflecting on all this, I think @Hartorn and @mattbit raised good points: I might've overcomplicated this.
My aim was to automate the wrapping of an output element: `"{reset}{font}{color}{" + PLACEHOLDER + "}{reset}"` with an interface, but yeah likely it was an overkill.

Based on all the comments above, here's what I propose:
- let's refactor `ScanLogger` into a `ScanPrinter` that handles two level of verbosities: `LOW` and `HIGH`.
- let's re-assign all the critical stuff (warnings and errors), into the `logging.Logger`, that way we have a clear separation between what's printer and what's logger. 
- let's forget about the dumping into file
- To simplify things, let's abandon the current `Template` and `xprint` interface. Let's just create aliases for the colors and simply call them directly.

An example:

```py
class ScanPrinter:
    def high(
        self,
        *args,
        template: Optional[Template] = None,
        filename: Optional[str] = None,
        **kwargs,
    ):
        if self.level.value <= VerbosityLevel.HIGH.value:
            xprint(*args, template=template, filename=filename, **kwargs)

RED_COLOR = Fore.LIGHTRED_EX
BLUE_COLOR = Fore.LIGHTBLUE_EX
MAGENTA_COLOR = Fore.LIGHTMAGENTA_EX

RED_STYLE = Style(color=RED_COLOR)
BLUE_STYLE = Style(color=BLUE_COLOR)
MAGENTA_STYLE = Style(color=MAGENTA_COLOR)

DetectedIssues = Template(content="{}: {} issues detected. (Took {})", pstyles=[BLUE_STYLE, RED_STYLE, MAGENTA_STYLE])
printer.low(detector.__class__.__name__, num_issues, time, template=DetectedIssues)
```

we just do:
```py
class ScanPrinter:
    def high(
        self,
        *args,
        **kwargs,
    ):
        if self.level.value <= VerbosityLevel.HIGH.value:
            print(*args, **kwargs)

from colorama import Fore
from colorama import Style


RED = Fore.LIGHTRED_EX
BLUE = Fore.LIGHTBLUE_EX
MAGENTA = Fore.LIGHTMAGENTA_EX

R = Style.RESET_ALL
BOLD = Style.BRIGHT
BRED = BOLD + RED
BBLUE = BOLD + BLUE
BMAGENTA = BOLD + MAGENTA

printer.low(f"{BBLUE}{detector.__class__.__name__}{R}: {RRED}{num_issues}{R} issues detected. (Took {BMAGENTA}{time}{R})"
```
Although we'd lose the idea of one `Template` for similar outputs, I think it makes it simpler and more readable.

Before diving into it (you may want to do it in a new branch @Sakayatsp , up to you), @mattbit @Hartorn do you agree on this?

Hartorn:
@rabah-khalek should we close this PR ?

PR #1756, opened 2024-01-22
[GSK-2311] Fixed load method 

rabah-khalek:
load method now uses load_model from the custom model class

### DIFF
--- a/tests/models/automodel/test_loading.py
+++ b/tests/models/automodel/test_loading.py
@@ -1,0 +1,46 @@
+import tempfile
+from pathlib import Path
+
+import pandas as pd
+from langchain.chains import LLMChain, load_chain
+from langchain.chains.base import Chain
+from langchain.llms.fake import FakeListLLM
+
+import giskard
+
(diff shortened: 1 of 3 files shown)


tests/models/automodel/test_loading.py
```diff
@@ -0,0 +1,46 @@
...
+        chain,
+        "text_generation",
+        name="Foo",
+        description="Model that reply foo, bar, baz",
+        feature_names=[TEXT_COLUMN_NAME],
+    )
+
+    with tempfile.TemporaryDirectory() as f:
+        model.save(f)
+        giskard.Model.load(f)
```
kevinmessiaen, line 46: I would probably add this assertion in the test to ensure load properly loaded the class:

```suggestion
        loaded_model = giskard.Model.load(f)
        assert isinstance(loaded_model, MyModel)
```

kevinmessiaen (approved):
LGTM, just left a small comment about the tests to ensure the loaded class is the one expected

Issue #1759, opened 2024-01-23
Refactor the `TestSuiteResult` structure


kevinmessiaen:
## 🚀 Feature Request

Currently the `TestSuiteResult` class uses a `List[Tuple[str, TestResult, Dict[str, Any]]])` for the `TestResult` information.

We need to have a `List[SuiteTestResult]` (`SuiteTestResult` does not exist yet) in order to greatly improve the code structure.

## 🔈 Motivation

This data structure is not humain readable and makes it hard to maintain (ei. We want to add an execution time for each test [in this PR](https://github.com/Giskard-AI/giskard/pull/1703))

PR #1766, opened 2024-01-24
Add type for SuiteTest

Smixi:
## Description

Add typing for test arguments.

## Related Issue

[](https://github.com/Giskard-AI/giskard/issues/1759)

## Type of Change

- [ ] 🔐 Security fix

## Checklist

- [x] I've updated the code style using `make codestyle`.

There is no make file for the `make codestyle` command ?


Hartorn:
> Btw, just looking into the logs :

Hello and thanks for your contribution :) 
CI is sometimes temperamental on github for windows and mac, let me retry it first.

We are using cachine here and there in the CI, but the error while installing packages was probably something looking like an architecture issue and/or some system packages missing. 

So should not be linked in any way to your PR indeed !

giskard/core/suite.py
```diff
@@ -76,10 +76,15 @@ def parse_function_arguments(client, project_key, function_inputs):
     return arguments
 
 
+TestName = str
+TestParams = Dict[str, Any]
+SuiteResult = Tuple[TestName, TestResult, TestParams]
```
kevinmessiaen, line 81: IMO this should be converted into a dataclass:

```python
@dataclass
def SuiteResult:
  test_name: str
  result: TestResult
  params: Dict[str, Any]
```

kevinmessiaen (commented):
For me we should avoid keeping the `Tuple` structure since it's a pain to maintain

giskard/core/suite.py
Hartorn: I guess so, but the aliasing is making it clearer than it was still

kevinmessiaen: Good point, it might be used externally. However we can prevent backward compatibility issue with some kind of hack:

```python
def SuiteResult:
  test_name: str
  result: TestResult
  params: Dict[str, Any]
  
  @property
  def _tuple():
    # Method to ensure backward compatibility
    return (self.test_name, self.result, self.params)

  def __getitem__(self, i):
    warnings.warn("The test results has been changed to a class, getting attribute by index will be removed in the future.", category=DeprecationWarning)
    return self._tuple[i]

  def __iter__(self, i):
    warning.warn("The test results has been changed to a class, getting attribute by index will be removed in the future.", category=DeprecationWarning)
    return self._tuple.__iter__()
```

Smixi: That should do the trick. This behavior should work only for readers (as setter must now set the dataclass instead of the tuples).
I believe tests are extensively using tuple, maybe they should be some effort to clean those sometimes.

kevinmessiaen: Well tuple are immutable so we don't have to worry about writing

Smixi: Sorry, by writer I meant any code passing the test result as tuple must now pass the dataclass from now (again I don't know if it is somehow exposed externally, I modified it in testsuite and seems only internal)

tests/test_suite.py
```diff
@@ -2,7 +2,7 @@
 import pytest
 
 from giskard import Model, Suite
-from giskard.core.suite import single_binary_result
+from giskard.core.suite import single_binary_result, SuiteResult
```
kevinmessiaen, line 5: Small details but we have a hook to sort imports in order to be able to solve conflicts efficiently 

```suggestion
from giskard.core.suite import SuiteResult, single_binary_result
```

kevinmessiaen (approved):
LGTM, thanks for your contribution!

PR #1772, opened 2024-01-29
Simplify scan metadata

luca-rossi:


giskard/slicing/tree_slicer.py
mattbit, line 1: This doesn't belong here

mattbit (changes requested):
All good but please remove extraneous changes

giskard/slicing/tree_slicer.py
luca-rossi: Sorry, didn't branch out correctly

mattbit (approved):

PR #1777, opened 2024-02-01
Switching from mlWorkerId to kernel name

Hartorn:
## Type of Change

- [ ] 🔐 Security fix

## Checklist


giskard/client/giskard_client.py
```diff
@@ -179,20 +180,27 @@ def get_project(self, project_key: str) -> Project:
...
+        description : str, optional
+            Describe your project, by default None
+
+        Returns
+        -------
+        Project
+            The project created in giskard
+
         """
+        # TODO(Bazire): handle properly the "auto" detection/creation of kernel
```
kevinmessiaen, line 203: 👍 

I really think that we need an auto detection and creation of kernel for ease of use.

kevinmessiaen (changes requested):
Not sure if I'm doing something wrong but I got this error:

`GiskardError: Not Found: Python kernel not found by key: No Kernel found with name external_worker`

I ran this piece of code:
```python
from giskard import GiskardClient

url = "http://localhost:9000"
api_key = "gsk-..."

# Create a giskard client to communicate with Giskard
client = GiskardClient(url, api_key)

client.create_project('test', 'test', 'external_worker', 'Empty desc')
```

and the worker have `external_worker` as kernel name:

image

giskard/client/giskard_client.py
Inokinoki: Do you think that we can create/use the kernel with the project id as name?

Inokinoki (commented):
LGTM in general

giskard/client/giskard_client.py
kevinmessiaen: Yes either by a project id or some name generated like: `{os_host}-{python_version}`

If we find another kernel using the same name with different dependencies we just do `{os_host}-{python_version}-{random-value}`

Hartorn: https://linear.app/giskard/issue/GSK-2896/auto-mode-for-kernel-is-a-need

PR #1806, opened 2024-02-19
GSK-2822 Don't edit global logger level

kevinmessiaen:
## Description

Updated logger to only update giskard log level to INFO by default

## Related Issue

- [GSK-2822 (Available on Linear)](https://linear.app/giskard/issue/GSK-2822/dont-edit-global-logger-level)

## Type of Change

- [ ] 🔐 Security fix

### DIFF
--- a/giskard/commands/cli_worker.py
+++ b/giskard/commands/cli_worker.py
@@ -91,7 +91,21 @@ def start_stop_options(func):
     default=None,
     help="Number of processes to use for parallelism (None for number of cpu)",
 )
-def start_command(url: AnyHttpUrl, is_server, api_key, is_daemon, hf_token, nb_workers):
+@click.option(
+    "--log-level",
+    "log_level",
+    default=logging.getLevelName(logging.INFO),
+    type=click.Choice(
+        [
+            logging.getLevelName(logging.DEBUG),
+            logging.getLevelName(logging.INFO),
+            logging.getLevelName(logging.WARNING),
+            logging.getLevelName(logging.ERROR),
+        ]
+    ),
+    help="Global log level",
+)
+def start_command(url: AnyHttpUrl, is_server, api_key, is_daemon, hf_token, nb_workers, log_level):
     """\b
     Start ML Worker.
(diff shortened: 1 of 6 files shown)


tests/utils/test_logging_utils.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_other_package_log_level_unset():
+    assert (
+        logging.getLogger(openai.__name__).level == logging.NOTSET
+    ), "Non giskard package log level should't be touched by giskard (NOTSET)"
+
+
+def test_root_log_level_default_warning():
+    assert logging.root.level == logging.WARNING, 'Root package log level should be set to default "WARNING"'
```
Hartorn, line 20: I'm all for this modification, but for scan and worker, we should have some logs, no ?

kevinmessiaen: Yes you're right, it should be an arg for the CLI to allow flexibility

giskard/commands/cli_worker.py
```diff
@@ -104,8 +118,11 @@ def start_command(url: AnyHttpUrl, is_server, api_key, is_daemon, hf_token, nb_w
     """
     analytics.track(
         "giskard-worker:start",
-        {"is_server": is_server, "url": anonymize(url), "is_daemon": is_daemon},
+        {"is_server": is_server, "url": anonymize(url), "is_daemon": is_daemon, "log_level": log_level},
     )
+
+    logging.root.setLevel(logging.getLevelName(log_level))
```
Hartorn, line 124: Setting log level does not work for me : if I set it to error, I still have logs from giskard weirdly. Not sure why

Hartorn: Normally setting Error level to root level should suppress every log even from giskard, in my case worker is still starting with logs.

Hartorn: Nvm, it was only at start, seems ok after

Hartorn (approved):

PR #1809, opened 2024-02-20
[GSK-2782] Allow GiskardClient to read api key from env

vjern:
## Description

There exists an env var called `GSK_API_KEY` used to [auto-provide the API key](https://github.com/Giskard-AI/giskard/blob/6f9bc5f9008f9688a5f894ad498d4f1ec45bf265/giskard/commands/cli_worker.py#L71) in the giskard worker CLI. It makes sense to reuse the name here.

We may want to also delete it from the environment once it's loaded to prevent leaks [as done in the CLI](https://github.com/Giskard-AI/giskard/blob/6f9bc5f9008f9688a5f894ad498d4f1ec45bf265/giskard/commands/cli_worker.py#L109).

I'm worried though that it may destroy the env var in the kernel itself, which would need to be restarted to properly instantiate a new client (or if the user was to reload the same cell where the client gets created).

## Todo

- [ ] Update client instantiation documentation to include (or update) a section on setting environment variables so they get picked up by `GiskardClient`.

## Testing Checklist

- [ ] Spin up a notebook with previously set env vars among which `GSK_API_KEY`
- [ ] Spin up a client without `key` kwarg and connect to the hub
- [ ] Destroy env var (`del os.environ['GSK_API_KEY']`) and attempt to re-instantiate client
  - This operation should ideally be idempotent. We could store the env var value as a class field or another property that will not get uploaded with the test code.

## Type of Change

- [ ] 🔐 Security fix

## Checklist

### DIFF
--- a/giskard/client/giskard_client.py
+++ b/giskard/client/giskard_client.py
@@ -128,7 +128,7 @@ def _limit_str_size(json, field, limit=255):
 class GiskardClient:
     def __init__(self, url: str, key: str = None, hf_token: str = None):
         self.host_url = url
-        self.key = key or os.getenv('GSK_API_KEY')
+        self.key = key or os.getenv("GSK_API_KEY")
         self.hf_token = hf_token
         base_url = urljoin(url, "/api/v2/")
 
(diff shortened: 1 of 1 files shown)


giskard/client/giskard_client.py
```diff
@@ -138,6 +138,9 @@ def __init__(self, url: str, key: str, hf_token: str = None):
 
         self._session.mount(url, adapter)
 
+        if self.key is None:
```
Hartorn, line 141: Nitpicking here, but could you put this check earlier ? like line 133

It's good practice to validate as soon as you can, then do the logic

Hartorn (changes requested):

PR #1822, opened 2024-02-26
Add detector param for min slice size

luca-rossi:


giskard/slicing/slice_finder.py
```diff
@@ -10,7 +10,16 @@ class SliceFinder:
...
-    def run(self, dataset: Dataset, features: Sequence[str], target: Optional[str] = None):
+    def run(
+        self,
+        dataset: Dataset,
+        features: Sequence[str],
+        target: Optional[str] = None,
+        min_slice_size: Optional[float] = None,
+    ):
+        if min_slice_size and len(dataset) < min_slice_size:
+            return {}
```
mattbit, line 21: nitpick: I prefer a more explicit `dict()` to avoid confusion with `set`.

mattbit (approved):

PR #1829, opened 2024-03-04
GSK-2896 Automatically generate a Kernel based on dependencies

kevinmessiaen:
## Description

Automatically generate a Kernel based on dependencies. 
If any kernel match current dependencies, then use it instead

## Type of Change

- [ ] 🔐 Security fix


giskard/client/giskard_client.py
```diff
@@ -180,7 +184,40 @@ def get_project(self, project_key: str) -> Project:
...
+        kernel_name = f"{base_kernel_name}-{uuid.uuid4()}"
+
+        kernels = self._session.get("kernels").json()
+        frozen_dependencies = [f"{dist.name}=={dist.version}" for dist in importlib_metadata.distributions()]
+
+        matching_kernels = [
+            kernel
+            for kernel in kernels
+            if kernel["name"].startswith(base_kernel_name)
+            and kernel["pythonVersion"] == python_version
```
Inokinoki, line 199: I think we can use `startswith` to allow changing patch version

giskard/client/giskard_client.py
```diff
@@ -180,7 +184,40 @@ def get_project(self, project_key: str) -> Project:
...
+        ]
+
+        if len(matching_kernels) > 0:
+            return matching_kernels[0]["name"]
+
+        self._session.post(
+            "kernels",
+            json={
+                "name": kernel_name,
+                "pythonVersion": python_version,
```
Inokinoki, line 210: I prefer storing the "major.minor.patch" here

giskard/client/giskard_client.py
```diff
@@ -180,7 +184,40 @@ def get_project(self, project_key: str) -> Project:
...
+            for kernel in kernels
+            if kernel["name"].startswith(base_kernel_name)
+            and kernel["pythonVersion"] == python_version
+            and set(frozen_dependencies).issubset(set(kernel["frozenDependencies"].split("\n")))
+        ]
+
+        if len(matching_kernels) > 0:
+            return matching_kernels[0]["name"]
+
+        self._session.post(
```
Inokinoki, line 206: Should we check the result here? Cuz for "PROCESS" kernel, backend will create an env as well. This can fail sometimes

Inokinoki (dismissed):

giskard/client/giskard_client.py
```diff
@@ -209,6 +245,10 @@ def create_project(self, project_key: str, name: str, kernel_name: str, descript
                 "name": anonymize(name),
             },
         )
+
+        if kernel_name is None:
```
Hartorn, line 249: Only issue here for me is if you try to create a project already existing, you will re-create a kernel everytime.
That's why changing kernel_name to `<project_key>_<auto>` (or at least without random/uuid in name) will protect that, since kernel creation is checking that

giskard/client/giskard_client.py
Hartorn: I think this object is doing it automatically, contrary to request ?

Hartorn (changes requested):
Main issue is with project creation error, and kernel is still created

giskard/client/giskard_client.py
Hartorn: Maybe we should change backend, to make start install deps and freeze if needed ?
We could make it so it save in DB the kernel, and then immediatly launch an async task in job manager to do install + freeze (+start ?) if needed ?

Hartorn: (unrelated to this PR)

giskard/client/giskard_client.py
Inokinoki: My bad: we have a pattern validator for only `major`.`minor`

giskard/client/giskard_client.py
Inokinoki: > I think this object is doing it automatically, contrary to request ?

Yes, you are right. So it will raise the exception in case of failed creation, right?

giskard/client/giskard_client.py
Inokinoki: idem

giskard/client/giskard_client.py
kevinmessiaen: I updated the kernel creation to use the `project_key`. It should not create a new kernel if the project key already exists (granted it has same dependencies)

giskard/client/giskard_client.py
```diff
@@ -180,7 +183,41 @@ def get_project(self, project_key: str) -> Project:
...
+            and existing_kernel["pythonVersion"] == python_version
+            and set(frozen_dependencies).issubset(set(existing_kernel["frozenDependencies"].split("\n")))
+        ):
+            # A similar kernel already exists
+            return existing_kernel["name"]
+        elif existing_kernel:
+            # A different kernel exists that uses the same name
+            kernel_name = f"{kernel_name}_{uuid.uuid4()}"
+
+        self._session.post(
```
Inokinoki, line 206: Can we extract this into a method like `create_kernel` in Giskard client?

I am also trying to do something similar for list kernels. Otherwise, I can do it on my side after merging this branch.

Inokinoki (commented):
LGTM, let's just handle @Hartorn 's comments on the case where project fails to create

giskard/client/giskard_client.py
kevinmessiaen: Sure it's done

kevinmessiaen:
> LGTM, let's just handle @Hartorn 's comments on the case where project fails to create

It's done, when the key already exists it'll reuse kernel of same key and won't create a new one.

Hartorn:
Looks good! Just need to test it on my side

Hartorn (approved):

PR #1839, opened 2024-03-11
Support for custom LLM clients

mattbit:
Support for custom LLMs (Mistral, OpenAI, Bedrock, etc.) in Giskard:

- simplify LLMClient, remove tools and function calls
- simplified evaluator prompts using chat mode one-shot
- added MistralClient
- refactored OpenAI client can also support local models through Ollama, because it now doesn't need function tools


kevinmessiaen (commented):
The idea is to remove `function_call` and `tools_call` and replace them by asking the model to answer json formatted answers?

I guess this is necessary to allow any model to work but how does it impact the reliability of current scan? 

Also, I know it might increase complexity but couldn't we have something like this instead so we keep the feature (not necessary if the switch won't impact scan reliability enough): 

```python
class Tool:
  tool_spec: Dict
  examples: Optional[List[Dict]]

  @property
  def to_format_instruction() -> str:
    pass

  def use_tool(chat_response) -> bool
    pass

  def parse_response(chat_response) -> List[Dict] # List because tool might be called more than once
    pass

class LLMClient(ABC):
  @abstractmethod
  def complete(..., tools: List[Tool] = [], tool_choice: Optional[Tool] = None):
     ...

class MistralClient(LLMClient):
  @abstractmethod
  def complete(..., tools: List[Tool] = [], tool_choice: Optional[Tool] = None):
    if (len(tools) > 0):
      tools_instruction = '\n'.join([tool.to_format_instruction for tool in tools])
      if (tool_choice is not None):
        tools_instruction += f'\n You must call {tool_choice.tool_call['name']}'
      # TODO: replace '{tools_instruction}' in `messages` and throw error if not found
      # 
    ...

# OpenAIClient use native tool_calls (if used model support it) 
```

I agree that we should keep the LLMClient as simple as possible. But I do believe that the client should be the one responsible to serialising the message to the API format. Feel free to disregard this comment if deemed not worth it to implement.

mattbit:
> The idea is to remove `function_call` and `tools_call` and replace them by asking the model to answer json formatted answers?
> I guess this is necessary to allow any model to work but how does it impact the reliability of current scan?

Yes, basically to be as flexible as possible in the support. By now, most of the models support some kind JSON format output (through `response_format="json_object"`), and in our tests the few-shot prompting in chat mode was just as effective.

> Also, I know it might increase complexity but couldn't we have something like this instead so we keep the feature (not necessary if the switch won't impact scan reliability enough):
> [...]
> I agree that we should keep the LLMClient as simple as possible. But I do believe that the client should be the one responsible to serialising the message to the API format and not the feature that need to be adapted to work for all LLMs (Ideally we should have something easy to understand in the scan/evaluator/... and the Client should be the translation based on the LLM's feature/needs). Feel free to disregard this comment if deemed not worth it to implement.

Yeah, it would be nice to abstract tools but then we would end up reimplementing langchain. I think it's wise to keep our code to a minimum for what is not strictly required. I realized that my first implementation with function calls was over engineered. I feel it's worth waiting for for now, maybe at some point there will be some light open-source library replacing langchain ;)

mattbit:
Things missing:

- [ ] Update docs
- [ ] Embeddings support for RAG toolset generation

kevinmessiaen:
> @kevinmessiaen there is a few important things that I would like you to double check:
> 
> * I removed the `details` from the `EvaluationResult` object, I couldn't really understand this structure. I understand that it's something for the hub, but can we make it simpler and less linked to the hub?

It's a very important feature of the Hub, but we should be able to generate `details` from the current `EvaluationResult` at the serialization step (we migth need to add the dataset indice but that's about it) 

> * I changed the LLM-dependent tests: in particular for the ground truth test, I put the `CorrectnessEvaluator` instead of the `RequirementEvaluator`, as it seemed to make more sense

Make sense to me

mattbit:
> It's a very important feature of the Hub, but we should be able to generate `details` from the current `EvaluationResult` at the serialization step (we migth need to add the dataset indice but that's about it)

What I don't want is the "TestResult" data structures inside the `llm` evaluation module. What is the information required for the Hub? The `TestResultDetails` remains shady to me. In any case, the `EvaluationResult` should expose all the info needed to create this, but creation of the test result should be responsibility of the test (I think we already have something like `create_test_result_from_evaluator`).

kevinmessiaen:
> > It's a very important feature of the Hub, but we should be able to generate `details` from the current `EvaluationResult` at the serialization step (we migth need to add the dataset indice but that's about it)
> 
> What I don't want is the "TestResult" data structures inside the `llm` evaluation module. What is the information required for the Hub? The `TestResultDetails` remains shady to me. In any case, the `EvaluationResult` should expose all the info needed to create this, but creation of the test result should be responsibility of the test (I think we already have something like `create_test_result_from_evaluator`).

It's used to display a table of all evaluations, it's a bit shady because the data structure is organized to reduce data size, actually it's converted back to EvaluationResult in the Hub.

The evaluator generate a list of dictionaries for each entries (151B):
`[{'reason': 'reason 1', 'conversation': 'conv 1'}, {'reason': 'reason 2', 'conversation': 'conv 2'}, {'reason': 'reason 3', 'conversation': 'conv 3'}]`

While we convert to a dictionary of lists to remove redundancy of keys (96B):
`{'reason': ['reason 1', 'reason 2', 'reason 3'], 'conversation': ['conv 1', 'conv 2', 'conv 3']}`

But yes you're right this should be outside of the `llm` module.

kevinmessiaen:
@mattbit I extracted the serialisation at DTO level. 

Only modification that I left on the `EvaluationResult` is the single result list (which allows to keep the order of results that is needed to match them with dataset row) and extraction of dict to EvaluationResultExample (arguably not a necessity).

mattbit:
Yeah this is definitely a major version and breaking for people who had built custom test based on our internal evaluators.

giskard/llm/client/mistral.py
```diff
@@ -0,0 +1,66 @@
...
+        max_tokens: Optional[int] = None,
+        caller_id: Optional[str] = None,
+        seed: Optional[int] = None,
+        format: str = None,
+    ) -> ChatMessage:
+        extra_params = dict()
+        if seed is not None:
+            extra_params["random_seed"] = seed
+
+        if format not in (None, "json", "json_object") and "large" not in self.model:
```
andreybavt, line 38: format could be of `Literal` type

also pydantic's `validate_arguments` could do the validation job

giskard/llm/embeddings/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    -------
+    BaseEmbedding
+    """
+    global _default_embedding
+
+    if _default_embedding is not None:
+        return _default_embedding
+
+    # Try with OpenAI/AzureOpenAI if available
+    try:
```
andreybavt, line 22: nitpick: the two try blocks could be extracted into dedicated methods

giskard/llm/embeddings/base.py
```diff
@@ -0,0 +1,22 @@
...
+import numpy as np
+
+
+class BaseEmbedding(ABC):
+    @abstractmethod
+    def embed(self, texts: Sequence[str]) -> np.ndarray:
+        ...
+
+
+def batched(iterable, batch_size):
```
andreybavt, line 15: should be moved to utils.py

giskard/llm/evaluators/base.py
```diff
@@ -1,45 +1,54 @@
...
+        return {"reason": self.reason, "sample": self.sample, "status": self.status}
 
 
 @dataclass
 class EvaluationResult:
-    failure_examples: Sequence[dict]
-    success_examples: Sequence[dict]
-    errors: Sequence[dict]
-    details: Optional[TestResultDetails] = None
+    results: Sequence[EvaluationResultExample] = field(default_factory=list)
```
andreybavt, line 39: since we're calling `.append` below this should rather be a List, not a Sequence

giskard/llm/evaluators/base.py
```diff
@@ -57,92 +66,109 @@ def has_errors(self):
...
-                {"input_vars": input_vars, "model_output": model_output, "reason": args.get("reason")},
-            )
-
-        return evaluation_result
+            logger.debug(f"{self.__class__.__name__} evaluation result: eval_passed={eval_passed}, reason={reason}")
+            result.add_sample(eval_passed, reason, sample)
+
+        return result
+
+    def _evaluate_sample(self, model: BaseModel, sample: Dict) -> Tuple[bool, str, Dict]:
```
andreybavt, line 137: return type should be `Tuple[bool, str]`

giskard/llm/evaluators/coherency.py
```diff
@@ -62,50 +68,56 @@ def evaluate(self, model: BaseModel, dataset_1: Dataset, dataset_2: Optional[Dat
         inputs_1 = dataset_1.df.to_dict("records")
         inputs_2 = dataset_2.df.loc[dataset_1.df.index].to_dict("records")
```
andreybavt, line 69: dataset_2 can be None

andreybavt (commented):
Generally looks good to me except for minor comments and a bunch of typing issues (we should add mypy or something similar)

I tested the new test_llm_ground_truth with the Hub, it worked fine

kevinmessiaen:
> Generally looks good to me except for minor comments, failing tests and a bunch of typing issues (we should add mypy or something similar)
> 
> I tested the new test_llm_ground_truth with the Hub, it worked fine

Should be good now, I just not sure about the first feedback regarding `Literal`

andreybavt (approved):

PR #1840, opened 2024-03-12
GSK-2409 Add doc to how to update the lockfile

kevinmessiaen:
## Description

- Added doc to how to update the lockfile
- Removed deprecated formatting instruction (new one hasn't been added since they're part of the pre-commit hook)

## Related Issue

- [GSK-2409 (Available on Linear)](https://linear.app/giskard/issue/GSK-2409/add-doc-to-how-to-update-the-lockfile)

## Type of Change

- [x] 📚 Examples / docs / tutorials / dependencies update
- [ ] 🔐 Security fix


Hartorn:
@kevinmessiaen I think we have an 'update-lock' command in pdm, to do it inside docker (else it may causes issues on MacOs). Maybe the command needs to be updated ?

kevinmessiaen:
> @kevinmessiaen I think we have an 'update-lock' command in pdm, to do it inside docker (else it may causes issues on MacOs). Maybe the command needs to be updated ?

Ah yes you're right, I've updated to use the `pdm update-lock` command instead

Hartorn (approved):

PR #1843, opened 2024-03-13
Issue with evaluating openai-based LLMs with mlflow.evaluate() API

rabah-khalek:
The issue is described here: https://github.com/mlflow/mlflow/issues/11266, and links to his proposals here: https://github.com/langchain-ai/langchain/issues/18420#issue-2164769403.

In a nutshell, upon passing a model uri to `mlflow.evaluate()`, mlflow tries to load the model, but since langchain don't yet support loading ChatOpenAI based models (the case of `RetrievalQA`powered by `gpt-4` for instance), `mlflow.evaluate()` was failing.

The workaround, was to avoid serializing, deserializing the langchain model, and pass a function to `mlflow.evaluate()`


giskard/integrations/mlflow/giskard_evaluator.py
```diff
@@ -107,6 +107,7 @@ def _generate_test_suite(self, scan_results):
                 "Please submit the traceback as a GitHub issue in the following "
                 "repository for further assistance: https://github.com/Giskard-AI/giskard."
             )
+            return None, None
```
Hartorn: Shouldn't we re raise an error ? as in perform scan ?

Hartorn (changes requested):

Hartorn (approved):

PR #1857, opened 2024-03-21
[GSK-3361] Model predictions are casted to double by default in regression mode

bmalezieux:
That's problematic, for decathlon their model outputs 32-bit precision values, and we're casting them to 64-bit, which I think makes our scan detecting it as data-leakage (predictions are changing with the same input)


rabah-khalek:
@Hartorn, @kevinmessiaen it seems that this is not breaking anything on the side of the library. @bmalezieux is verifying it on the hub side.

bmalezieux:
> @Hartorn, @kevinmessiaen it seems that this is not breaking anything on the side of the library. @bmalezieux is verifying it on the hub side.

It works for our current project, but maybe this would break in other scenarios.

giskard/models/base/wrapper.py
```diff
@@ -134,8 +134,8 @@ def predict_df(self, df: pd.DataFrame, *_args, **_kwargs):
 
         raw_prediction = np.concatenate(outputs)
 
-        if self.is_regression:
-            return raw_prediction.astype(float)
+        # if self.is_regression:
```
Hartorn, line 137: Just remove the line ;)

Hartorn (changes requested):

Hartorn:
> > @Hartorn, @kevinmessiaen it seems that this is not breaking anything on the side of the library. @bmalezieux is verifying it on the hub side.
> 
> It works for our current project, but maybe this would break in other scenarios.

I will try to run the e2e tests to make sure of it
Did you try on the regression notebooks in docs/reference/notebooks ?

Hartorn:
Running in https://github.com/Giskard-AI/giskard-hub/actions/runs/8374478584/job/22929809288?pr=544

Hartorn (approved):

Hartorn:
Tested on insurance_prediction_lgbm and hotel_text_regression

PR #1888, opened 2024-04-12
Fix call to LLM logger breaking scan

mattbit:
For scans that do not require the LLM client (special char injections, etc.), the scanner must not break if the LLM client is not set.


kevinmessiaen (approved):

giskard/scanner/scanner.py
```diff
@@ -334,6 +335,9 @@ def _get_cost_estimate(self, model, dataset, detectors):
         }
 
     def _get_cost_measure(self):
+        llm_logger = _maybe_get_llm_logger()
+        if llm_logger is None:
+            return None
         llm_logger = get_default_client().logger
```
andreybavt, line 341: you can remove this line, the `llm_logger` will already contain this value

andreybavt (changes requested):

PR #1904, opened 2024-04-19
Fix raget docs

pierlj:
- Fix RAGAS import error during docs build
- Update docstrings and docs to match the latest RAGET modifications


rabah-khalek (approved):
lgtm

Issue #1924, opened 2024-05-03
Raget: Possible miscalculation of all  Ragas metrics, in particular Precision and Recall


Chabert-Liddell:
### Issue Type

Bug

### Source

source

### Giskard Library Version

2.11

### Giskard Hub Version

-

### Current Behaviour?

```shell
Giskard RAGet uses the reference context when calling Ragas. 

https://github.com/Giskard-AI/giskard/blob/main/giskard/rag/metrics/ragas_metrics.py

        ragas_sample = {
            "question": question_sample["question"],
            "answer": answer,
            "contexts": question_sample["reference_context"].split("\n\n"),
            "ground_truth": question_sample["reference_answer"],
        }

According to Ragas documentation the retrieved context should be used (the one used for the answer Generation).

As an example, when computing Precision or Recall which both uses {"question", "contexts", "ground_truth"}, if you are giving the reference context, then you are evaluating your test set generation pipeline  and not your RAG pipeline.
```

### Standalone code OR list down the steps to reproduce the issue

```shell
.
```

alexcombessie:
@pierlj what do you think?

pierlj:
Hi @Chabert-Liddell, you are right, thanks for pointing this out. A fix will be release soon!

PR #1925, opened 2024-05-10
[GSK-3513] Fix RAGAS metric computation

pierlj:
According to [this issue](https://github.com/Giskard-AI/giskard/issues/1924), some RAGAS metrics are not properly computed by RAGAS (this includes context recall, context precision and faithfulness). 

To fix it: 
- added a `retrieved_documents` argument to the evaluate method 
- allowed the `answer_fn` to return retrieved documents alongside the answer to a question

### DIFF
--- a/giskard/rag/metrics/__init__.py
+++ b/giskard/rag/metrics/__init__.py
@@ -1,4 +1,4 @@
-from .base import Metric
+from .base import Metric, ModelOutput
 from .correctness import CorrectnessMetric, correctness_metric
 
-__all__ = ["Metric", "correctness_metric", "CorrectnessMetric"]
+__all__ = ["Metric", "correctness_metric", "CorrectnessMetric", "ModelOutput"]
--- a/giskard/rag/metrics/ragas_metrics.py
+++ b/giskard/rag/metrics/ragas_metrics.py
@@ -5,7 +5,8 @@ import logging
 import nest_asyncio
 
 from ...llm.client import ChatMessage, LLMClient, get_default_client
-from .base import Metric
+from ...llm.embeddings import BaseEmbedding, get_default_embedding
+from .base import Metric, ModelOutput
 
 nest_asyncio.apply()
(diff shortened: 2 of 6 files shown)


mattbit:
@pierlj looks good, can you add a test on the ragas metrics to make sure they are calculated correctly?

mattbit (approved):
Good for me, @pierlj do you want to make a last check?

pierlj:
> Good for me, @pierlj do you want to make a last check?

Yep, I will have a look

giskard/rag/metrics/ragas_metrics.py
pierlj, line 1: Why an empty except block here?

giskard/rag/metrics/ragas_metrics.py
```diff
@@ -26,6 +25,15 @@
     ) from err
 
 
+try:
+    # RAGAS has async functions, we try to apply nest_asyncio to avoid issues in notebooks
+    import nest_asyncio
+
+    nest_asyncio.apply()
+except:  # noqa
```
pierlj, line 33: Why an empty except block here?

giskard/rag/metrics/__init__.py
pierlj, line 1: I don't think it's needed here

giskard/rag/metrics/__init__.py
```diff
@@ -1,4 +1,5 @@
+from ..base import AgentAnswer
```
pierlj, line 1: This import can be removed I think

giskard/rag/metrics/ragas_metrics.py
pierlj, line 33: Why an empty exception block here ?

mattbit: just in case `nest_asyncio` is not available or something breaks due to strange env config

PR #1929, opened 2024-05-14
JSON report generation from scan results

bmalezieux:
## Type of Change

- [ ] 🔐 Security fix

## Checklist

  modified)


giskard/core/suite.py
```diff
@@ -219,6 +220,16 @@ def _to_dto(self, label: Optional[str], client: GiskardClient, project_key: str)
             completionDate=self.completion_date.isoformat(),
         )
 
+    def to_json(self, filename=None):
+        results = {}
+        for suite_result in self.results:
+            results[suite_result.test_name] = "Passed" if suite_result.result.passed else "Failed"
```
rabah-khalek, line 226: can we add all the meta data that is available / could be useful? 
- metric value
- model name
- dataset(s) name(s)

giskard/core/suite.py
```diff
@@ -219,6 +220,16 @@ def _to_dto(self, label: Optional[str], client: GiskardClient, project_key: str)
...
+        if filename is not None:
+            with open(filename, "w") as json_file:
+                json.dump(results, json_file, indent=4)
+        else:
+            return results
```
rabah-khalek, line 231: I would perhaps return the json representation instead of the dict here
```py
return json.dumps(results, indent=4)
```

giskard/core/suite.py
```diff
@@ -574,6 +585,7 @@ def run(self, verbose: bool = True, **suite_run_args):
 
                 if isinstance(result, bool):
                     result = TestResult(passed=result)
+                print("TEMP: ", test_partial.test_id, result.passed)
```
rabah-khalek, line 588: to be removed

rabah-khalek (changes requested):
since `<detector>.run()` produces a list of issues, which then you're looping over to jsonify, I would suggest to implicitly infer the detector names inside the run, and add this info to the issue's meta, which then you can group based on in `to_json` of `ScanReport`. `detectors_names` shouldn't be an input to the report, as we already have this info somewhere.

giskard/core/suite.py
rabah-khalek: Or let's simply remove this else, and always return the `results`

giskard/scanner/report.py
```diff
@@ -18,7 +19,7 @@
 
 
 class ScanReport:
-    def __init__(self, issues, model=None, dataset=None, as_html: bool = True):
+    def __init__(self, issues, model=None, dataset=None, detectors_names=None, as_html: bool = True):
```
rabah-khalek, line 22: i thought the idea was to remove this argument. Do we really need to know which detector didn't produce issues?

tests/scan/test_scan_report.py
```diff
@@ -51,3 +51,25 @@ def test_scan_report_exports_to_markdown():
         assert dest.exists()
         assert dest.is_file()
         assert dest.read_text() == markdown
+
+
+def test_scan_report_to_json():
```
rabah-khalek, line 56: let's add a test where you pass the detector-names list to the scanreport, and check if the detector is in the json file

rabah-khalek (approved):

PR #1932, opened 2024-05-17
Disabled perturbation on date

kevinmessiaen:
## Description

Disabled perturbation on date

## Type of Change

- [ ] 🔐 Security fix


giskard/push/perturbation.py
```diff
@@ -121,7 +121,7 @@ def _apply_perturbation(
     ds_slice_copy = ds_slice.copy()
 
     # Apply the transformation
-    if col_type == SupportedPerturbationType.NUMERIC:
+    if col_type == SupportedPerturbationType.NUMERIC and ds.df.dtypes[feature] in ["int64", "float64"]:
```
rabah-khalek, line 124: what about `int32`, `float32`, etc.?

I think this https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_datetime64_any_dtype.html would be better

kevinmessiaen: I ended up using https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.api.types.is_any_real_numeric_dtype.html which is more robust.

As it will exclude all non-real numeric data type so we don't end up with the same issue with another pseudo-numeric data type.

rabah-khalek (approved):

PR #2029, opened 2024-09-19
Added workflow to automatically rerun Python build on failure up to times

kevinmessiaen:
Added workflow to automatically rerun Python build on failure up to times


alexcombessie:
@kevinmessiaen OK to merge, as tests are OK?

henchaves (approved):
LGTM, but we should have a second approval before merging 
cc @Hartorn

PR #2098, opened 2025-01-23
Added test data locally

kevinmessiaen:
## Description

Added test data locally instead of downloading them to a ftp server


mattbit (changes requested):
I think you also need to update the tutorials

mattbit:
Note: waiting for s3 bucket to move assets

kevinmessiaen:
@mattbit Created a new PR since it was faster to start from main #2109

PR #2231, opened 2026-01-21
feat(ci): add release workflows for giskard libraries

kevinmessiaen:
- Add reusable release template workflow for publishing to PyPI
- Add manual version bump workflow for patch/minor/major bumps
- Add release workflows for giskard-core, giskard-agents, and giskard-checks
- Release workflows trigger on version bumps to main branch


.github/workflows/manual-bump.yml
```diff
@@ -0,0 +1,66 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: read
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
+
+      - name: Install uv
+        uses: stral-sh/setup-uv@681c641aba71e4a1c380be3ab5e12ad51f415867 # v7
```
mattbit: Typo here!

.github/workflows/manual-bump.yml
```diff
@@ -0,0 +1,66 @@
...
+          OLD_V=$(grep '^version\s*=' libs/$PACKAGE/pyproject.toml | cut -d '"' -f2)
+
+          NEW_V=$(uv run python -c "
+          import semver
+          v = semver.VersionInfo.parse('$OLD_V')
+          print(v.bump_${{ env.BUMP_TYPE }}())
+          ")
+
+          # Update the file
+          sed -i "s/^version = \"$OLD_V\"/version = \"$NEW_V\"/" pyproject.toml
```
mattbit: File here should be the `libs/$PACKAGE/pyproject.toml`, but you are working in the root dir, so I think this will edit the main `pyproject.toml` – which is not what we want.

.github/workflows/manual-bump.yml
```diff
@@ -0,0 +1,66 @@
...
+          echo "new_version=$NEW_V" >> $GITHUB_OUTPUT
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v6
```
mattbit: Can we avoid this unknown third party action?
I would rather use Github official [github-script](https://github.com/actions/github-script) action or directly with the `gh` CLI (see [here](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli)).

.github/workflows/manual-bump.yml
mattbit: Don't we need to lock again after this update?

.github/workflows/release-giskard-agents.yml
```diff
@@ -0,0 +1,12 @@
+name: Release Giskard Agents
+on:
+  push:
+    branches: [main]
+    paths: ["libs/giskard-agents/pyproject.toml"] # Only triggers if pyproject.toml changes (potential version bump)
```
mattbit: it's a bit strange to do this, because any changes to the pyproject (e.g. added a new test dependency) would trigger a potential release workflow. The release will then check and interrupt the workflow if the tag is already existing. Personally, I would be ok keeping the release trigger manual. In practice it can be aggregated with the bump workflow:
- bump version and lock
- add tag
- publish package

What is the advantage of having the bump to open a PR and then having the release triggered on merge?

mattbit (changes requested):
Hey @kevinmessiaen, I think there's some mistakes in the bump workflow.

I also suggest to take a simpler approach where the release workflow is triggered manually (and includes the version bump). It would be similar to what we have today – I don't know if there's an advantage in opening the bump PR.

Note also that if keeping the bump PR, I think you'll need to use the releaser bot or the workflows will not run on the PR.

.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+        with:
+          tag_name: ${{ steps.bump.outputs.tag }}
+          name: "${{ env.PACKAGE }} v${{ steps.bump.outputs.new_version }}"
+          generate_release_notes: true
+          files: |
+            libs/${{ env.PACKAGE }}/dist/*.tar.gz
+            libs/${{ env.PACKAGE }}/dist/*.whl
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
mattbit, line 128: use exact sha here

.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+          if [ ${#CMD[@]} -gt 2 ]; then
+            "${CMD[@]}"
+          else
+            echo "No version bump"
+            exit 1
+          fi
+
+          NEW_VERSION=$(uv version --short)
+          echo "new_version=$NEW_VERSION" >> $GITHUB_OUTPUT
+          echo "tag=${{ env.PACKAGE }}/v$NEW_VERSION" >> $GITHUB_OUTPUT
```
mattbit, line 96: it's good practice to use env variables rather than direct input

.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: read
+      id-token: write
+    env:
+      PACKAGE: ${{ github.event.inputs.package }}
+      BUMP_TYPE: ${{ github.event.inputs.bump_type }}
+      RELEASE_TYPE: ${{ github.event.inputs.release_type }}
+      TEST_MODEL: "gemini/gemini-2.0-flash"
```
Hartorn, line 49: Move to needed step

.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+          python-version: "3.12"
+
+      - name: Setup development environment
+        run: make setup
+
+      - name: Run checks
+        run: make check
+
+      - name: Run tests
+        run: make test
```
Hartorn, line 70: Should we ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+          name: "${{ env.PACKAGE }} v${{ steps.bump.outputs.new_version }}"
+          generate_release_notes: true
+          files: |
+            libs/${{ env.PACKAGE }}/dist/*.tar.gz
+            libs/${{ env.PACKAGE }}/dist/*.whl
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: libs/${{ env.PACKAGE }}/dist/
```
Hartorn, line 130: Missing token ? check how this work instead of uv publish ?

henchaves (approved):

PR #2233, opened 2026-01-22
feature(checks): Improve string matching checks

kevinmessiaen:


libs/giskard-checks/src/giskard/checks/builtin/string_matching.py
```diff
@@ -1,64 +1,229 @@
+"""String matching check implementation.
+
+This module provides a check that validates whether a keyword appears within
+a text string. It supports Unicode normalization, case sensitivity control,
+and flexible text/keyword extraction from traces.
+"""
```
pierlj, line 6: Is this required? It's not present in other checks, I would remove it for homogeneity

libs/giskard-checks/src/giskard/checks/builtin/string_matching.py
```diff
@@ -1,64 +1,229 @@
...
+        default="NFKC",
+        description="Unicode normalization form to apply (NFC, NFD, NFKC, NFKD). Defaults to NFKC.",
+    )
+    case_sensitive: bool = Field(
+        default=True,
+        description="If True, matching is case-sensitive. If False, both strings are lowercased before comparison.",
+    )
+
+    @model_validator(mode="after")
+    def validate_keyword_or_keyword_key(self) -> Self:
```
pierlj, line 119: Same should be done for the `text` and `text_key`

kevinmessiaen: It's not nullable since we default to  `trace.last.outputs` (the last result) if nothing is set

PR #2236, opened 2026-01-27
refactor(checks): replace Equality with ComparisonCheck system

kevinmessiaen:
Replace the Equality check with a more general ComparisonCheck base class that supports multiple comparison operators (equals, greater than, less than, greater than or equal, less than or equal).

BREAKING CHANGE: The Equality check has been replaced with Equals and other comparison checks. Import paths have changed from Equality to Equals, and new comparison operators are now available.

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Any, override
+from typing import Any, Self, override
 
-from pydantic import Field
+from giskard.core import NOT_PROVIDED, NotProvided, provide_not_none
+from pydantic import Field, model_validator
 
 from ..core.check import Check
-from ..core.extraction import NoMatch, resolve
+from ..core.extraction import NoMatch, provided_or_resolve, resolve
 from ..core.result import CheckResult
 from ..core.trace import Trace
 from ..utils.normalization import NormalizationForm, normalize_data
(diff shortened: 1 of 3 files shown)


libs/giskard-checks/src/giskard/checks/builtin/comparison.py
```diff
@@ -0,0 +1,360 @@
...
+        return "less than or equal to"
+
+    @property
+    @override
+    def _operator_symbol(self) -> str:
+        """Get the operator symbol for error messages."""
+        return "<="
+
+
+@Check.register("greater_equals")
```
pierlj: For consistency
```suggestion
@Check.register("greater_than_equals")
```

libs/giskard-checks/src/giskard/checks/builtin/comparison.py
```diff
@@ -0,0 +1,360 @@
...
+        return "greater than or equal to"
+
+    @property
+    @override
+    def _operator_symbol(self) -> str:
+        """Get the operator symbol for error messages."""
+        return ">="
+
+
+@Check.register("equality")
```
pierlj: For consistency
```suggestion
@Check.register("equals")
```

libs/giskard-checks/src/giskard/checks/builtin/comparison.py
```diff
@@ -0,0 +1,360 @@
...
+    - Value extraction from traces
+    - NoMatch handling
+    - Error handling for unsupported comparisons
+    - Result formatting
+    """
+
+    actual_value_key: str = Field(
+        ..., description="The key to extract the actual value from the trace"
+    )
+    expected_value: ExpectedType = Field(
```
pierlj: Any reason not having a parameters for `expected_value_key`?

PR #2239, opened 2026-02-06
feat(agents,core): [ENG-1272]/[ENG-1271] refactor rate limiters and add generator retry/timeout policies

kevinmessiaen:
- Moved rate limiters to core library
- Added timeout support for generator requests
- Implemented max delay for retry backoff
- Enhanced retry policy configuration (+ new tests)

### DIFF
--- a/libs/giskard-core/src/giskard/core/rate_limiter/base.py
+++ b/libs/giskard-core/src/giskard/core/rate_limiter/base.py
@@ -139,6 +139,7 @@ class BaseRateLimiter(Discriminated, ABC):
             or 0.0 if no wait was required.
         """
         raise NotImplementedError
+        yield 0.0  # unreachable; makes this an async generator for @asynccontextmanager typing
 
     def initialize_state(self, existing: Self | None = None) -> None:
         """Initialize internal state, optionally sharing from an existing matching instance.


mattbit (commented):
Thanks @kevinmessiaen! I only checked the core part for now: the implementation looks solid.

I have some doubts on the general modularity of this. For example, it's quite verbose:
```py
rate_limiter = RateLimiter.from_rpm(100, max_concurrent=10)

generator.with_rate_limiter(rate_limiter).run(...)
```
after:
```py
# we could gain a little with the max_concurrent helper, but doesn't change much
rate_limiter_rpm = MaxRequestsPerMinute(rpm=60)
rate_limiter_concurrency = MaxConcurrentRequests(max_concurrent=10)

generator.with_rate_limiters([
    rate_limiter_rpm,
    rate_limiter_concurrency
]).run(...)
```

Is it actually worth the tradeoff? I don't really expect to see a lot of extension beyond this, the basic controls on min interval and concurrency seem enough for most applications.

What worries me about this verbosity is not only that it makes it more complicated to read the code, but also that it will make it easier to introduce bugs because we forget one of the conditions.
The thing that I like the least is that we'd need to separately register each rule. Before I could do something like
```py
rate_limiter = RateLimiter.from_rpm(100, max_concurrent=10, rate_limiter_id="global")
...
# elsewhere in codebase
generator.with_rate_limiter("global")
```
and that would enforce all my limits. Now I would need to specify all IDs since each rule is registered separately.
```py
rate_limiter_1 = limiter.max_concurrent(10, "global_max_concurrent")
rate_limiter_2 = limiter.rpm(100, "global_rpm")
...
# elsewhere, I need to remember that both rules need to be passed
generator.with_rate_limiters("global_rpm", "global_max_concurrent")
```

The verbosity here can quickly become a source of bugs, when people will forget that actually that global rate limiter is 2 rules and just put "global_rpm".
It will also make it maintainability worse, let's say I had this:
```py
RateLimiter.from_rpm(100, rate_limiter_id="global")
...
# elsewhere
generator.with_rate_limiter("global")
```
If now I need to add max concurrency: I will just update my global definition, no need to change other code. In the current implementation instead, I would need to update all usages to add a "global_max_concurrent" everywhere.

Side remark: The current `MaxRequestPerMinute` is in fact a 'max time interval' in disguise: it's not only about RPM, but also spacing requests uniformly. I think it could be misunderstood, because 100 RPM could mean doing 100 reqs in the first second, and then idle for 59 seconds. Something like `MinInterval.from_rpm/request_per_hour` would be more explicit and more general.

---

My suggestion: let's keep the RateLimiter instance as the actual actor. If we want to keep the modularity, let's move it inside the RateLimiter.

I think that when people think about a rate limiter, they think about a single instance that enforces certain conditions (possibly multiple tules, e.g. both min interval and max concurrency).

We can always structure the RateLimiter as something containing a list of rules if needed, e.g.
```
class RateLimiter(...):
    id: str
    rules: list[RateLimitRule]
```

libs/giskard-core/src/giskard/core/limiter/basic.py
```diff
@@ -0,0 +1,68 @@
...
+
+@BaseRateLimiter.register("basic_rate_limiter")
+class BasicRateLimiter(BaseRateLimiter):
+    min_interval: float = Field(..., ge=0)
+    max_concurrent: int | None = Field(default=None, ge=1)
+
+    @override
+    def model_post_init(self, context: Any, /) -> None:
+        super().model_post_init(context)
+        pass
```
mattbit: uhm, what is this doing?

libs/giskard-core/src/giskard/core/limiter/base.py
```diff
@@ -0,0 +1,85 @@
...
+    @asynccontextmanager
+    def throttle(self) -> AsyncGenerator[float]:
+        raise NotImplementedError
+
+    def create_initial_state(self) -> Any:
+        return None
+
+    @property
+    def state(self) -> Any:
+        return self._registry.get_state(self)
```
mattbit: Why don't you bind the state to the instance? Once the reference is created by the registry, you could let it inject and avoid the lookup.

libs/giskard-core/src/giskard/core/limiter/base.py
```diff
@@ -0,0 +1,85 @@
...
+    Field,
+)
+
+from ..discriminated import Discriminated, discriminated_base
+
+
+class RateLimiterRegistry:
+    """Share limiter state across instances despite serialization round-trips."""
+
+    _lock: threading.Lock = threading.Lock()
```
mattbit: you are initializing the lock at class level, but immediately shadowing it in the init

libs/giskard-core/src/giskard/core/limiter/base.py
```diff
@@ -0,0 +1,85 @@
...
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._instances = {}
+        self._states = {}
+
+    def register_instance(self, rate_limiter: "BaseRateLimiter"):
+        with self._lock:
+            instances = self._instances.get(rate_limiter.id)
+            if instances is None:
+                instances = WeakSet()
```
mattbit: type annotation could help here `WeakSet["BaseRateLimiter"]()`

libs/giskard-core/src/giskard/core/limiter/base.py
```diff
@@ -0,0 +1,85 @@
...
+                self._instances[rate_limiter.id] = instances
+
+            if instances:
+                existing_instance = next(iter(instances))
+                if existing_instance != rate_limiter:
+                    raise ValueError(
+                        f"Rate limiter with id '{rate_limiter.id}' already registered"
+                    )
```
mattbit: I feel
```py
try:
    ...
except StopIteration:
    ...
```
would be more idiomatic

It would be nice to add some comments explaining exactly what we are doing here.

libs/giskard-core/src/giskard/core/limiter/base.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def get_state(self, rate_limiter: "BaseRateLimiter") -> Any:
+        return self._states.get(rate_limiter.id, None)
+
+    def get_instance(self, id: str) -> "BaseRateLimiter":
+        instances = self._instances.get(id)
+        if instances is None:
+            raise ValueError(f"Rate limiter with id '{id}' not found")
+
+        return next(iter(instances))
```
mattbit: this could still raise a `StopIteration` if instance was deleted meanwhile? I suggest using a single try except `next(iter(self._instances.get(id))` and handle the exceptions (should be `StopIteration` and `TypeError` if the instances are not iterable).

libs/giskard-core/src/giskard/core/limiter/base.py
mattbit: instances contents are `WeakSet`, but how about states? They are never cleaned up from my understanding.

libs/giskard-core/src/giskard/core/limiter/base.py
```diff
@@ -0,0 +1,85 @@
...
+            instances = self._instances.get(rate_limiter.id)
+            if instances is None:
+                instances = WeakSet()
+                self._instances[rate_limiter.id] = instances
+
+            if instances:
+                existing_instance = next(iter(instances))
+                if existing_instance != rate_limiter:
+                    raise ValueError(
+                        f"Rate limiter with id '{rate_limiter.id}' already registered"
```
mattbit: We could be more explicit with the error and say that we are trying to register the same rate limiter with different configuration…
By the way, shouldn't we try to handle this gracefully instead of failing? Because this would make certain objects impossible to unserialize.

libs/giskard-core/tests/test_rate_limiter.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def test_rate_limiter_should_cleanup_state_when_last_instance_is_deleted(self):
+        _rate_limiter_a = RateLimiter.from_rpm(60, max_concurrent=1, id="test")
+        assert _rate_limiter_a.state is not None
+        assert _rate_limiter_a.state.semaphore is not None
+        del _rate_limiter_a
+
+        _rate_limiter_b = RateLimiter.from_rpm(
+            120, id="test"
+        )  # This should not raise an error since the state is cleaned up
```
mattbit: I don't think this is testing what it says. I don't see what cleans up the state, in my opinion it's still there in the registry. Only the `instances` (`WeakSet`) are cleaned up. When this happens, the new rate limiter is overwriting the state.

I think that after `del _rate_limiter_a` if you check the registry you'll still find its state there.

libs/giskard-core/src/giskard/core/limiter/base.py
kevinmessiaen: yes it could be deleted in between

libs/giskard-core/src/giskard/core/limiter/basic.py
kevinmessiaen: Nothing, forgot to clean up

PR #2249, opened 2026-02-16
feat(generators): introduce personas and extended context for UserSimulator

davidberenstein1957:
- Added PersonaSimulator class to simulate user interactions with predefined or custom personas.
- Implemented output structure for client descriptions and conversation management.
- Created a Jinja2 template for persona instructions and context handling.
- Added tests to validate functionality and ensure consistent client descriptions across conversation turns.

## Description

closes #ENG-1388

## Type of Change

- [ ] 🔐 Security fix

## Checklist

  modified)

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/generators/user.py
+++ b/libs/giskard-checks/src/giskard/checks/generators/user.py
@@ -38,6 +38,12 @@ class UserSimulator[TraceType: Trace](  # pyright: ignore[reportMissingTypeArgum
 
     Examples
     --------
+    Predefined persona:
+
+    >>> simulator = UserSimulator(persona="frustrated_customer")
+
+    Custom persona with optional context:
+
     >>> simulator = UserSimulator(
     ...     persona="A polite elderly user who needs step-by-step guidance",
     ...     context="Ask about using the mobile app"
--- a/libs/giskard-checks/README.md
+++ b/libs/giskard-checks/README.md
@@ -166,7 +166,8 @@ API Overview
 
 **Interaction specs**
 - `giskard.checks.InteractionSpec`: discriminated base for describing inputs/outputs. Subclasses implement `generate()` to yield interactions.
-- `giskard.checks.Interact`: batteries-included spec that supports static values, callables, or generators for both inputs and outputs. Supports multi-turn interactions via generators.
+- `giskard.checks.Interact`: batteries-included spec that supports static values, callables, generators, or `InputGenerator` instances for both inputs and outputs. Supports multi-turn interactions via generators.
+- `giskard.checks.UserSimulator`: LLM-powered input generator that simulates user personas (predefined or custom) for multi-turn scenarios.
 
 **Scenarios and runners**
 - `giskard.checks.Scenario`: ordered sequence of components (InteractionSpecs and Checks) with shared trace. Components execute sequentially, stopping at first failure.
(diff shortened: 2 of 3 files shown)


kevinmessiaen:
> @kevinmessiaen @henchaves, I feel this also introduces some discussion on how to pass along and re-use information from `Traces` in preceding interactions or checks when using the `n_runs` setting. If we used the current persona setup, the persona defined during the first interaction would not be reused in subsequent interactions, because we would need to re-initialize the class. I'd expect to have the options to do something like `trace.last.interactionspec` to re-use the same class instance.
> 
> We could pass some class information along by default as metadata, or we can configure something more robust. Happy to hear your thoughts.

@davidberenstein1957 I not sure to understand the issue with `n_runs`. To me if we have multiple runs, they should all be self-contained and not sharing information. 

Regarding reusing the persona, I feel like the simpliest fix would be to allow describe a reusable persona and create simulators from persona.

```py
helpfull_persona = Persona("helpful_user", ...) # Persona only hold information on how to generate message, not the what to generate

simulator_1 = helpfull_persona.simulator("Describe your problem installing giskard v3", max_turn=3)
simulator_2 = helpfull_persona.simulator("Say that you tried the solution but you get resolution error, you are using python 3.10", max_turn=3)

# Change persona dynamically (assume inside a scenario
simulator_3 = Persona("thankfull_user", ...) .simulator("State that it's working now with python 3.13", max_turn=1)
```

libs/giskard-checks/src/giskard/checks/generators/user.py
```diff
@@ -22,13 +22,40 @@ class UserSimulatorOutput(BaseModel):
...
-    max_steps: int = Field(default=3)
+    """User simulation with predefined or custom personas.
+
+    Accepts either a predefined persona name (e.g., "frustrated_customer") or a custom
+    persona description.
+
+    Parameters
+    ----------
+    persona : str
+        Predefined persona name or custom persona description
```
kevinmessiaen, line 33: It's not documented what are the predefined persona. It would be nice to provide the list in the docstring

libs/giskard-checks/src/giskard/checks/generators/user.py
```diff
@@ -22,13 +22,40 @@ class UserSimulatorOutput(BaseModel):
...
+    --------
+    >>> simulator = UserSimulator(
+    ...     persona="A polite elderly user who needs step-by-step guidance",
+    ...     context="Ask about using the mobile app"
+    ... )
+    """
+
+    persona: str = Field(
+        ..., description="Predefined persona name or custom description", min_length=1
+    )
```
kevinmessiaen, line 55: While it's concise to have a single str field I'm fearing of potential issues:

- Typos won't be noticed: `persona=helpfull_user` or `persona=helpful-user` will work without error, but the instruction in the prompt won't be what we expect them to be
- No error if a persona has been removed/deleted (even if we should not make breaking changes, if this happens it will be easy to miss this)

Shall we have an enum of persona?

```py
class Persona(StrEnum):
    HelpfulUser = "helful_user"
    ...

# In the UserSimulator class
persona: Persona | str = Field(...)

# Then we can instaciate persona safely
UserSimulator(persona=Persona.HelpfulUser)
UserSimulator(persona="helful_user") # Will act as Persona.HelpfulUser but without type safety
UserSimulator(persona="You are an angry customer") # Custom prompt
```

This would also help fixing the documentation issue, as it'll be possible to explore personas with autocompletion/autodoc

PR #2264, opened 2026-02-18
refactor: introduce JsonPathStr type for JSONPath validation

davidberenstein1957:
- Updated comparison, groundedness, semantic similarity, and string matching checks to use the new JsonPathStr type for key fields.
- Added validation for JSONPath syntax in the extraction module.
- Introduced unit tests for JSONPath validation and enforcement tests to ensure all relevant fields use JsonPathStr.

## Type of Change

- [X] 🥂 Improvement (non-breaking change which improves an existing feature)
- [ ] 🔐 Security fix

## Checklist

  modified)

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,8 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(git log:*)",
+      "Bash(python -m pytest:*)"
+    ]
+  }
+}
--- a/libs/giskard-checks/src/giskard/checks/core/trace.py
+++ b/libs/giskard-checks/src/giskard/checks/core/trace.py
@@ -133,8 +133,10 @@ class Trace[InputType, OutputType](BaseModel, frozen=True):
 
     async def with_interaction(
         self,
-        interaction: Interaction[InputType, OutputType]
-        | InteractionGenerator[Interaction[InputType, OutputType], Self],
+        interaction: (
+            Interaction[InputType, OutputType]
+            | InteractionGenerator[Interaction[InputType, OutputType], Self]
+        ),
     ) -> Self:
         if isinstance(interaction, Interaction):
             return self.model_copy(
--- a/libs/giskard-checks/src/giskard/checks/scenarios/runner.py
+++ b/libs/giskard-checks/src/giskard/checks/scenarios/runner.py
@@ -58,8 +58,10 @@ class _ScenarioStepsBuilder[InputType, OutputType, TraceType: Trace]:  # pyright
 
     def add_interaction(
         self,
-        interaction: Interaction[InputType, OutputType]
-        | InteractionGenerator[Interaction[InputType, OutputType], TraceType],
+        interaction: (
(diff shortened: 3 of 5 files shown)


libs/giskard-checks/src/giskard/checks/core/extraction.py
```diff
@@ -15,11 +15,33 @@
...
+from pydantic import AfterValidator, BaseModel, Field
 
 from .trace import Trace
 
 
+class _JsonPathStrMarker:
+    """Marker placed in JsonPathStr metadata. Used by the enforcement test."""
+
+
+def _validate_jsonpath_syntax(v: str) -> str:
```
henchaves: the name of this method is not very correct, we are mixing the syntax check with the trace convention check. So, it would be better to rename it 

for example:
```suggestion
def _validate_trace_jsonpath(v: str) -> str:
```

libs/giskard-checks/src/giskard/checks/core/extraction.py
```diff
@@ -15,11 +15,33 @@
...
+    if not v.startswith("trace."):
```
henchaves: extract "trace." to a constant outside the method

`_REQUIRED_JSONPATH_PREFIX = "trace."`

henchaves (changes requested):

libs/giskard-checks/src/giskard/checks/builtin/string_matching.py
```diff
@@ -91,15 +91,15 @@ class StringMatching[InputType, OutputType, TraceType: Trace](  # pyright: ignor
         default=None,
         description="The text string to search within. If None, extracted from trace using text_key.",
     )
-    text_key: str = Field(
+    text_key: JsonPathStr = Field(
```
mattbit: Minor thing but for consistency, `JSONPathStr` would be the recommended [PEP8](https://peps.python.org/pep-0008/) style:

> When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

libs/giskard-checks/src/giskard/checks/core/extraction.py
```diff
@@ -15,11 +15,36 @@
...
+from pydantic import AfterValidator, BaseModel, Field
 
 from .trace import Trace
 
 
+class _JsonPathStrMarker:
+    """Marker placed in JsonPathStr metadata. Used by the enforcement test."""
+
+
+_REQUIRED_JSONPATH_PREFIX = "trace."
```
mattbit, line 28: I'm wondering: what's the use of forcing each key to start with `trace`? If this is the only thing allowed, shouldn't we extract the trace attributes instead?
E.g. `trace.last` -> `last`. Not needed in this PR, but maybe worth a quick discussion on the desired interface here.

henchaves: to make clear that is coming from the trace

Issue #2267, opened 2026-02-24
Support `annotations` in Scenario/Trace


henchaves:
**Summary**

Introduce **Scenario/Trace-level annotations** as a first-class concept in `giskard-checks`, allowing callers to attach shared, read-only context to a scenario that is exposed on the `Trace` object (e.g. `trace.annotations.my_key`) and usable from checks. This is complementary to, not a replacement for, per-interaction `metadata`.

**Problem Statement**

Today, `giskard-checks` supports:

* **Per-interaction metadata** via `Interaction.metadata`, attached as interactions are generated.
* **Trace-level inspection** via `Trace` and JSONPath selectors such as `trace.last.outputs`.

However, there is **no place to attach scenario-wide context** that:

* Is **not** tied to a specific interaction.
* Should be **stable across the whole scenario** (e.g. identifiers, configuration, reference values).
* Can be **referenced in checks and templates** alongside trace data.

This leads to the following issue:

* Checks that want to compare outputs against a stable reference must either:
  * Hard-code the reference in the check configuration, or
  * Encode it into `metadata` of one of the interactions and then fish it out again.

We need a **simple, explicit way to attach shared, scenario-level context** that participates in the trace object model and can be addressed via JSONPath.

**Goals**

* Add an `annotations` mapping at the **Scenario** level and surface it on the **Trace**:
  * `Scenario.annotations: dict[str, Any]`
  * `Trace.annotations: dict[str, Any]`
* Ensure annotations are **read-only at runtime** from the point of view of user code:
  * They are set when the scenario starts.
  * Every `Trace` instance produced during execution carries the same annotations value.
* Make annotations available to:
  * Built-in checks using JSONPath keys like `trace.annotations.my_key`.
  * Custom checks and templates via `Trace.annotations`.
* Support both:
  * Direct `Scenario(...)` construction.
  * The fluent builder (`scenario().with_annotations(...).interact().check()`).

**Non-goals**

* No mutation APIs for annotations during scenario execution (e.g. no `trace.set_annotation(...)`).
* No merging of annotations from interactions or checks (annotations are scenario-owned, not interaction-owned).
* No additional namespacing or hierarchical annotation system beyond a flat `dict[str, Any]`.

**Examples**

1. "As a user, I want to define a scenario with a reference answer and check that the model output matches that reference without hard-coding it twice."

```python
from giskard.checks import Equals, Scenario, Interaction

scenario = Scenario(
    name="reference_match",
[...]

PR #2268, opened 2026-02-24
feat: [OSS-20] Add support to `annotations` in scenario/trace

henchaves:
## Description

- Added `annotations: dict[str, Any]` to `Scenario` and `Trace`
- Added `with_annotations` method to `scenario` builder API

## Related Issue
#2267 

## Type of Change

- [ ] 🔐 Security fix

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/core/trace.py
+++ b/libs/giskard-checks/src/giskard/checks/core/trace.py
@@ -137,8 +137,10 @@ class Trace[InputType, OutputType](BaseModel, frozen=True):
 
     async def with_interaction(
         self,
-        interaction: Interaction[InputType, OutputType]
-        | InteractionGenerator[Interaction[InputType, OutputType], Self],
+        interaction: (
+            Interaction[InputType, OutputType]
+            | InteractionGenerator[Interaction[InputType, OutputType], Self]
+        ),
     ) -> Self:
         if isinstance(interaction, Interaction):
             return self.model_copy(
--- a/libs/giskard-checks/src/giskard/checks/scenarios/runner.py
+++ b/libs/giskard-checks/src/giskard/checks/scenarios/runner.py
@@ -58,8 +58,10 @@ class _ScenarioStepsBuilder[InputType, OutputType, TraceType: Trace]:  # pyright
 
     def add_interaction(
         self,
-        interaction: Interaction[InputType, OutputType]
-        | InteractionGenerator[Interaction[InputType, OutputType], TraceType],
+        interaction: (
+            Interaction[InputType, OutputType]
+            | InteractionGenerator[Interaction[InputType, OutputType], TraceType]
+        ),
     ):
         if len(self.current_step.checks) > 0:
             self.add_step()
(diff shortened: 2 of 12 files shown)


libs/giskard-checks/src/giskard/checks/core/scenario.py
```diff
@@ -77,6 +78,10 @@ class Scenario[InputType, OutputType, TraceType: Trace](BaseModel, frozen=True):
         default=None,
         description="Type of trace to use for the scenario. If not provided, the trace type will be inferred from the sequence of components.",
     )
+    annotations: dict[str, Any] = Field(
+        default_factory=dict,
+        description="Scenario-level annotations shared across the trace.",
```
mattbit:
```suggestion
        description="Scenario-level annotations that will be injected in the trace.",
```

libs/giskard-checks/src/giskard/checks/core/trace.py
```diff
@@ -154,6 +158,14 @@ async def with_interaction(
             if generator is not None:
                 await generator.aclose()
 
+    def with_annotations(self, annotations: dict[str, Any]) -> Self:
+        """Return a new trace instance with updated annotations.
+
+        This keeps the trace model immutable while allowing scenario-level
+        annotations to be projected onto all derived traces.
+        """
+        return self.model_copy(update={"annotations": annotations})
```
mattbit: You may want to merge with the existing annotations instead of replacing the whole object.

mattbit: I notice that this is returning a copy, but on the builder the same method is modifying in place. I think we should have a convention, or clarify the difference?

mattbit: E.g. maybe we don't need the "with_annotations" on the trace itself. We could set by assignment, that would remove any ambiguity, it would be managed by the runner when it creates the trace.

libs/giskard-checks/src/giskard/checks/scenarios/runner.py
```diff
@@ -131,6 +131,7 @@ async def run[InputType, OutputType, TraceType: Trace[Any, Any]](
             if scenario.trace_type is not None
             else cast(TraceType, Trace[InputType, OutputType]())
         )
+        trace = trace.with_annotations(scenario.annotations)
```
mattbit: I would simply do `trace_type(annotations=scenario.annotations)`. I see that we have the same somewhat confusing pattern with `with_interactions`. The code here is a but messy, maybe we can leave as-is and do a little cleanup later.

mattbit (commented):
Would do a little clean up of the `with_annotation` setter. I would rather keep these things explicit on the trace object since it's a frozen object.

libs/giskard-checks/src/giskard/checks/core/trace.py
mattbit: Thinking again, in practice merging annotations does not make sense for now. We are always creating the trace from scratch. Best would be an explicity `Trace(annotations=...)`, dropping the syntax sugar here for this internal usage.

PR #2279, opened 2026-02-26
feat(agents): generator-as-protocol-adapter and tool coercion

Hartorn:
## Known limitation

The `serialize_tools`, `serialize_messages`, and `deserialize_response` methods on `BaseGenerator` are currently "advisory" — they exist as override points but are called from inside `LiteLLMGenerator._complete_once`, not orchestrated by the base class. A new provider subclass could skip them entirely with no enforcement.

**Next step:** Restructure so `BaseGenerator` owns the serialize → call → deserialize pipeline, with the subclass implementing only a `_send(wire_messages, wire_tools, params)` method. This also ties into replacing the mixin-based retry/rate-limiting with a composable middleware pipeline. Tracked as a follow-up.

Add serialize_tools, serialize_messages, deserialize_response to BaseGenerator so provider subclasses own all wire-format translation. LiteLLMGenerator now delegates through these methods instead of calling to_litellm/from_litellm directly.

Tool.run() gains input coercion (TypeAdapter + _params_model) and output serialization (_return_adapter), returning str to eliminate double-quoting in workflow. Workflow _run_tools simplified accordingly.

## Type of Change

- [ ] 🔐 Security fix

## Checklist

  modified)

### DIFF
--- a/libs/giskard-agents/tests/test_generator.py
+++ b/libs/giskard-agents/tests/test_generator.py
@@ -1,6 +1,6 @@
 import json
 import time
-from typing import Any
+from typing import Any, override
 from unittest.mock import patch
 
 import pytest
@@ -13,7 +13,7 @@ from giskard.agents.tools import Function, Tool, ToolCall, tool
 from giskard.agents.workflow import ChatWorkflow
 from giskard.core import MinIntervalRateLimiter
 from litellm import ModelResponse
-from pydantic import Field, PrivateAttr
+from pydantic import Field
 
 
 @pytest.fixture
@@ -29,7 +29,7 @@ def mock_response():
 
 
 async def test_litellm_generator_completion_with_mock(
-    generator: LiteLLMGenerator, mock_response
+    generator: LiteLLMGenerator, mock_response: ModelResponse
 ):
     with patch(
         "giskard.agents.generators.litellm_generator.acompletion",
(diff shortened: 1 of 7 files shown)


libs/giskard-agents/src/giskard/agents/generators/base.py
```diff
@@ -22,17 +23,150 @@
...
+    chat code must never call provider APIs directly; they work exclusively
+    with ``Message`` objects and delegate wire translation to the generator.
+    """
 
     params: GenerationParams = Field(default_factory=GenerationParams)
     retry_policy: RetryPolicy | None = Field(default=None)
     rate_limiter: BaseRateLimiter | None = Field(default=None)
     middlewares: list[CompletionMiddleware] = Field(default_factory=list)
 
+    # -- Protocol adapter methods ------------------------------------------
```
mattbit: We should not expose these "adapter" methods or impose a structure for the `BaseGenerator`:
- These methods constraint the structure to have a clear mapping to a "messages" and "tools" list. This may not be a valid format for all APIs.
- In practice these are not protocol adapters, they are only used internally.

The actual protocol for the `Generator` is represented by the `complete` & derived methods. Serializazion and deserialization should be handled internally. Up to the `Generator` implementer to decide how to do that.

The wrong pattern we had was exposing the `to_litellm_function` on the base `Tool`. This has been correctly fixed in the PR.

I propose (if useful) to move these as private methods on the `LiteLLMGenerator`, removing the abstractions that are not needed.

mattbit (approved):
All good!

PR #2281, opened 2026-02-27
refactor(agents): replace mixin-based retry/rate-limiting with middle…

Hartorn:
…ware pipeline

Replace WithRetryPolicy and WithRateLimiter mixins with a composable CompletionMiddleware pipeline on BaseGenerator. This decouples cross-cutting concerns (retry, rate-limiting) from the generator inheritance hierarchy, making them stackable and independently testable.

## Type of Change

- [ ] 🔐 Security fix

## Checklist

  modified)

### DIFF
--- a/libs/giskard-agents/src/giskard/agents/tools/tool.py
+++ b/libs/giskard-agents/src/giskard/agents/tools/tool.py
@@ -4,7 +4,7 @@ import inspect
 from typing import Any, Callable, Literal, TypeVar
 
 import logfire_api as logfire
-from pydantic import BaseModel, Field, create_model
+from pydantic import BaseModel, Field, PrivateAttr, TypeAdapter, create_model
 
 from ..context import RunContext
 from ..errors.serializable import Error
@@ -38,6 +38,8 @@ class Tool(BaseModel):
     catch: Callable[[Exception], Any] | None = Field(default=None)
 
     run_context_param: str | None = Field(default=None)
+    _params_model: type[BaseModel] | None = PrivateAttr(default=None)
+    _return_adapter: TypeAdapter[Any] | None = PrivateAttr(default=None)
 
     @classmethod
     def from_callable(
@@ -96,7 +98,9 @@ class Tool(BaseModel):
             **fields,
         )
 
-        return cls(
+        return_annotation = sig.return_annotation
+
+        tool_instance = cls(
             name=fn.__name__,
             description=description,
             parameters_schema=model.model_json_schema(),
@@ -104,6 +108,10 @@ class Tool(BaseModel):
             run_context_param=run_context_param,
             catch=catch,
         )
+        tool_instance._params_model = model
+        if return_annotation is not inspect.Parameter.empty:
+            tool_instance._return_adapter = TypeAdapter(return_annotation)
+        return tool_instance
 
     def __call__(self, *args: Any, **kwargs: Any) -> Any:
         """Call the underlying function without modification.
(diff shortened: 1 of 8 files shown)


libs/giskard-agents/src/giskard/agents/generators/middleware.py
```diff
@@ -0,0 +1,94 @@
...
+@CompletionMiddleware.register("retry")
+class RetryMiddleware(CompletionMiddleware):
+    """Retries failed completions with exponential back-off (tenacity).
+
+    Override ``_should_retry`` in subclasses for provider-specific logic.
+    """
+
+    max_attempts: int = Field(default=3)
+    base_delay: float = Field(default=1.0)
+    max_delay: float | None = Field(default=None)
```
mattbit: We should retain the `RetryPolicy` object (which is the configuration) and encapsulate in the middleware, as done for the `RateLimiter`.

libs/giskard-agents/src/giskard/agents/generators/litellm_generator.py
```diff
@@ -7,51 +7,50 @@
...
+
+
+@BaseGenerator.register("litellm")
+class LiteLLMGenerator(BaseGenerator):
+    """A generator for creating chat completion pipelines using LiteLLM."""
 
     model: str = Field(
         description="The model identifier to use (e.g. 'gemini/gemini-2.0-flash')"
     )
+    middleware: list[CompletionMiddleware] = Field(
```
mattbit: `middlewares` (plural)

mattbit (changes requested):
The architectural change is solid and definitely an improvement, however the usability is impacted as we are making harder to configure retries and rate limiting on the fly.

E.g. say I have a `Generator` instance and I want to update the number of attempts, the current API is simple and predictable
```py
generator.with_retries(3)
```

While with the proposed middleware:
```py
# do we already have a retry middleware somewhere?
found = False
for i, mw in enumerate(generator.middleware):
    if isinstance(mw, RetryMiddleware):
        found = True
        # if we found it let's clone the generator and then replace the middleware in place
        new_generator = generator.model_copy()
        new_generator.middleware[i] = mw.model_copy(update={"max_attempts":5})

# otherwise, we'll add a new one
if not found:
    # where do I need to put the retry, at the beginning or at the end?
    # well it needs to be at the beginning, because otherwise it would break how the rate limiter middleware works
    new_generator = generator.model_copy()
    new_generator.middleware = [RetryMiddleware(max_attempts=5)] + generator.middleware
```

Overall, my feeling is that the middleware is flexible, but `RateLimiter` and `Retry` mechanisms are in practice very specific, for example they need to be added in precise order. It's easy to forget and add Retry after RateLimit.

I propose a compromise:
- Keep the middleware chain as the internal architecture. This is solid and very effective for advanced extension (hooks, etc.)
- Modify the `_build_chain` to add these two _fundamental_ mechanisms. Some kind of built-in middleware, rough example:
```py
def _build_chain(self, core: NextFn) -> NextFn:
    all_mw: list[CompletionMiddleware] = []
    if self._retry_mw:
        all_mw.append(self._retry_mw)
    if self._rate_limiter_mw:
        all_mw.append(self._rate_limiter_mw)
    all_mw.extend(self.middleware)
    return reduce(_wrap, reversed(all_mw), core)
```
(could probably be written better).
We can also keep exposing the internal configs via properties:
```py
@property
def rate_limiter(self):
    return self._rate_limiter_mw.rate_limiter if self._rate_limiter_mw else None
```
And we can also implement the `with_retries` and `with_rate_limiter`.

In this way, we would have:
- the convenient methods and easy access to a generator configuration (e.g. `generator.retry_policy`)
- strong guarantees that these middlewares are executed in the right order
- composable architecture with the middleware pattern
- extensibility with the exposed middlewares

Let me know what you think or if you have better ideas.

PR #2283, opened 2026-02-27
feat: [OSS-22] `Suite`: run batches of scenarios with dynamic binding

henchaves:
## Description

- refactored `Interact` class to set `outputs` as `NOT_PROVIDED` by default, raising error if still `NotProvided` during interaction generation
- refactored `Scenario` and `Scenario.run` to include a `target` attribute, which will be used to replace on run time the non provided `outputs`
- created a new class `Suite` (and `SuiteResult`) to run a batch of scenarios

## Related Issue

[OSS-22 (available on Linear)](https://linear.app/giskard/issue/OSS-22/suite-run-suites-of-scenarios-with-dynamic-binding)

## Type of Change

- [ ] 🔐 Security fix

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/core/result.py
+++ b/libs/giskard-checks/src/giskard/checks/core/result.py
@@ -75,15 +75,23 @@ class CheckResult(BaseModel):
 
     Attributes
     ----------
-    status:
+    status : CheckStatus
         Outcome status of the check.
-    message:
+    message : str or None
         Optional short message to surface to users (e.g., success/failure reason).
-    metrics:
-        List of auxiliary metrics captured by the check.
-    details:
+    metrics : list[Metric]
+        Auxiliary metrics captured by the check.
+    details : dict[str, Any]
         Arbitrary structured payload with additional context (e.g., failure reasons,
         timings, and any metadata the check wishes to include).
+    passed : bool
+        True if ``status`` is ``PASS``.
+    failed : bool
+        True if ``status`` is ``FAIL``.
+    errored : bool
+        True if ``status`` is ``ERROR``.
+    skipped : bool
+        True if ``status`` is ``SKIP``.
     """
 
     model_config: ClassVar[ConfigDict] = ConfigDict(frozen=True)
@@ -201,16 +209,24 @@ class ScenarioResult[InputType, OutputType](BaseModel):
 
     Attributes
     ----------
-    scenario_name:
+    scenario_name : str
         Name of the scenario that was executed.
-    check_results:
-        List of all check results from executed checks.
-    passed:
-        Whether all executed checks passed.
-    duration_ms:
+    steps : list[TestCaseResult]
+        Ordered list of test case results produced during execution.
+    duration_ms : int
         Total execution time in milliseconds.
-    final_trace:
-        The trace state after execution, containing all interactions that occurred.
+    final_trace : Trace[InputType, OutputType]
+        Trace state after execution, containing all interactions that occurred.
+    status : ScenarioStatus
+        Aggregated outcome of the scenario derived from its steps.
+    passed : bool
+        True when all steps passed.
+    failed : bool
+        True when at least one step failed and none errored.
+    errored : bool
+        True when at least one step errored.
+    skipped : bool
+        True when all steps were skipped.
     """
 
     scenario_name: str = Field(..., description="Scenario name")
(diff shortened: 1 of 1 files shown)


libs/giskard-checks/src/giskard/checks/scenarios/suite.py
```diff
@@ -0,0 +1,117 @@
...
+    ) -> None:
+        """Add a scenario to the suite.
+
+        Parameters
+        ----------
+        scenario : Scenario
+            The scenario to add to the suite. For scenario builders, use `scenario.build()` first.
+        """
+
+        self.scenarios.append(scenario)
```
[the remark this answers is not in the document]
henchaves: Let's not automatically build then

libs/giskard-checks/src/giskard/checks/core/result.py
```diff
@@ -418,3 +418,52 @@ def __rich_console__(
...
+    def skipped_count(self) -> int:
+        """Number of skipped scenarios."""
+        return sum(1 for r in self.results if r.skipped)
+
+    @computed_field
+    @property
+    def pass_rate(self) -> float:
+        """The pass rate of the suite (passed scenarios / total scenarios)."""
+        if not self.results:
+            return 1.0
```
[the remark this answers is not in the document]
mattbit: Well I think Gemini is not wrong here, the usual convention is that empty = failure. It's the same for pytest for example, if no test ran it will exit with non zero status code.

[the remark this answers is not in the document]
henchaves: it's a bit contradictory with the scenario logic

```
    @computed_field
    @property
    def status(self) -> ScenarioStatus:
        """The status of the scenario."""
        if not self.steps:
            return ScenarioStatus.PASS
 ```

File: libs/giskard-checks/src/giskard/checks/core/result.py [hunk 3]
```diff
@@ -418,3 +422,110 @@ def __rich_console__(
...
+        for r in self.results:
+            if r.status == ScenarioStatus.PASS:
+                char, color = ".", STATUS_MAPPING["pass"]["color"]
+            elif r.status == ScenarioStatus.FAIL:
+                char, color = "F", STATUS_MAPPING["fail"]["color"]
+            elif r.status == ScenarioStatus.ERROR:
+                char, color = "E", STATUS_MAPPING["error"]["color"]
+            else:  # SKIP
+                char, color = "s", STATUS_MAPPING["skip"]["color"]
+            dots += f"[{color}]{char}[/{color}]"
```
davidberenstein1957:
```suggestion
            if r.passed:
                char, color = ".", STATUS_MAPPING["pass"]["color"]
            elif r.failed:
                char, color = "F", STATUS_MAPPING["fail"]["color"]
            elif r.errored:
                char, color = "E", STATUS_MAPPING["error"]["color"]
            elif r.skipped:
                char, color = "s", STATUS_MAPPING["skip"]["color"]
            else:
                raise NotImplementedError(f"Status {r.status} handling not implemented")
            dots += f"[{color}]{char}[/{color}]"
```

libs/giskard-checks/src/giskard/checks/core/result.py
```diff
@@ -418,3 +422,110 @@ def __rich_console__(
...
+        # Failure/Error summary
+        failures_and_errors = [r for r in self.results if r.failed or r.errored]
+        if failures_and_errors:
+            yield ""
+            yield "[bold red]Failures/Errors Summary:[/bold red]"
+            for f in failures_and_errors[:20]:
+                status = STATUS_MAPPING[f.status]
+                yield f"  [{status['color']}]{f.status.value.upper()}[/{status['color']}] {f.scenario_name}"
+            if len(failures_and_errors) > 20:
+                yield f"  ... and {len(failures_and_errors) - 20} more"
```
davidberenstein1957:
```suggestion
        if failures_and_errors:
            n_loggable_failures = 20
            yield ""
            yield "[bold red]Failures/Errors Summary:[/bold red]"
            for f in failures_and_errors[:n_loggable_failures]:
                status = STATUS_MAPPING[f.status]
                yield f"  [{status['color']}]{f.status.value.upper()}[/{status['color']}] {f.scenario_name}"
            if len(failures_and_errors) > n_loggable_failures:
                yield f"  ... and {len(failures_and_errors) - n_loggable_failures} more"
```

libs/giskard-checks/src/giskard/checks/scenarios/suite.py
```diff
@@ -0,0 +1,130 @@
...
```
davidberenstein1957: should we add a check to see if the scenario has be build?

libs/giskard-checks/src/giskard/checks/scenarios/suite.py
```diff
@@ -0,0 +1,130 @@
...
+        Examples
+        --------
+        ```python
+        from giskard.checks import Suite
+
+        suite = Suite(name="my_suite", target=my_sut_v1)
+        suite.append(scenario_1)
+        suite.append(scenario_2)
+        result_v1 = await suite.run()
+        result_v2 = await suite.run(target=my_sut_v2)
```
davidberenstein1957, line 117: so this means there are 3 different ways to define the SUT? 
- during `Suite.__init__`
- for individual scenarios
- durign `Suite.run`?

libs/giskard-checks/src/giskard/checks/scenarios/suite.py
henchaves: we can also implement auto-build, might be easier for users

libs/giskard-checks/src/giskard/checks/scenarios/suite.py
henchaves: for when running Suite, yes. For `Scenario` there are 2 different ways: `Scenario.__init__` and `Scenario.run()`

PR #2293, opened 2026-03-06
fix(checks): preserve generator and embedding model when running scenarios

### LINKED ISSUE #2292: Bug: input generator instance is reset by runner, losing state
I have written a custom input generator, similar to a `PersonaSimulator`. I used the `WithGeneratorMixin` to provide the LLM `generator` instance.

The issue is that this generator is reset to default value at runtime:

```python
my_simulator = MySimulator(generator=Generator(model="my-model"))
```

At runtime in the scenario, the simulator generator is reset to the default value from `get_default_generator`.

This happens because the instance is serialized and then deserialized at runtime, but the `generator` field is excluded from the serialized object. So it's lost.

Some questions:

* why is the input generator serialized and deserialized (can't we instead keep the same instance?)
* instead of serializing, shouldn't we use `model_copy()`
* do we really need to exclude the `generator` field from the serialized version? It should be serializable

In my case, the only solution was to set the default generator so that it would be set to the right value every time, but this is not practicel.

kevinmessiaen:
## Description

- Stop excluding generator and embedding_model from serialization in WithGeneratorMixin and WithEmbeddingMixin so custom instances are preserved at runtime instead of being reset to defaults.
- Refactor scenario runner target binding into _bind_target helper and use model_copy + model_validate for Interact components so provider state is correctly initialized.
- Register mock generator and embedding model in tests for serialization.

## Related Issue

Fixes #2292

## Type of Change

- [ ] 🔐 Security fix

## Checklist

  modified)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,9 +79,7 @@ typecheck: ## Run type checking with basedpyright
 	uv tool run basedpyright --level error .
 
 security: ## Check for security vulnerabilities
-	# TODO: Remove --ignore-vuln CVE-2026-4539 flag when patch exists for pygments
-	# This is a low severity redos vulnerability: https://www.cve.org/CVERecord?id=CVE-2026-4539
-	uv run pip-audit --skip-editable --ignore-vuln CVE-2026-4539
+	uv run pip-audit --skip-editable
 
 generate-licenses: ## Generate licenses
 	uv tool run licensecheck --license MIT \
--- a/uv.lock
+++ b/uv.lock
@@ -719,7 +719,7 @@ dev = [
 
 [[package]]
 name = "giskard-agents"
-version = "1.0.2a1"
+version = "1.0.2b1"
 source = { editable = "libs/giskard-agents" }
 dependencies = [
     { name = "giskard-core" },
@@ -748,7 +748,7 @@ requires-dist = [
 
 [[package]]
 name = "giskard-checks"
-version = "1.0.1a2"
+version = "1.0.1b1"
 source = { editable = "libs/giskard-checks" }
 dependencies = [
     { name = "giskard-agents" },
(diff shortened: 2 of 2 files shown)


libs/giskard-checks/tests/builtin/test_judge.py
```diff
@@ -39,6 +42,24 @@ def serialization_roundtrip[InputType, OutputType, TraceType: Trace](  # pyright
     return cast(LLMJudge[InputType, OutputType, TraceType], check)
 
 
+async def test_custom_generator_preserved_after_serialization_roundtrip() -> None:
```
davidberenstein1957, line 45: do we need this test, if we are already testing the roundtrip behavior below ?

libs/giskard-checks/tests/builtin/test_semantic_similarity.py
```diff
@@ -333,6 +334,33 @@ async def test_empty_trace() -> None:
     assert isinstance(result.details["actual_answer"], NoMatch)
 
 
+async def test_custom_embedding_model_preserved_after_serialization_roundtrip() -> None:
```
davidberenstein1957, line 337: I believe this is already tested in the test below too, by not manually resetting the embedding model.

libs/giskard-checks/tests/builtin/test_judge.py
kevinmessiaen: Yes other tests also validate the roundtrips. But I think it's better to have a test specifically for this feature, making it obvious of what break if it fails and making it clear that we want the generator to be preserved.

libs/giskard-checks/tests/builtin/test_semantic_similarity.py
kevinmessiaen: Same as above

PR #2323, opened 2026-03-23
feat(checks): show scenario and check error details in suite report

kevinmessiaen:
- Add failures_and_errors on ScenarioResult and TestCaseResult
- Refactor suite dots to use STATUS_MAPPING symbols
- Render failures in Panels with FAILURES/SUMMARY sections and nested check output
- Align CheckResult rich output with tab-separated details for non-pass states

New report: 

CleanShot 2026-03-23 at 10 44 14@2x

Old report:

image

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/core/result.py
+++ b/libs/giskard-checks/src/giskard/checks/core/result.py
@@ -531,6 +531,11 @@ class SuiteResult(BaseResult, frozen=True):
             return 1.0
         return self.passed_count / denominator
 
+    @property
+    def failures_and_errors(self) -> list[ScenarioResult[Any, Any]]:
+        """Return a list of scenario results that failed or errored."""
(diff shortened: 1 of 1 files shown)


libs/giskard-checks/src/giskard/checks/core/result.py
```diff
@@ -517,30 +537,38 @@ def __rich_console__(
...
-        yield dots
-
-        # Failure/Error summary
+        yield "".join(
+            f"[{STATUS_MAPPING[r.status]['color']}]{STATUS_MAPPING[r.status]['symbol']}[/{STATUS_MAPPING[r.status]['color']}]"
+            for r in self.results
+        )
+        yield ""
+
         failures_and_errors = [r for r in self.results if r.failed or r.errored]
```
henchaves: could move it outside the method, making possible to call `suite_result.failures_and_errors`

PR #2329, opened 2026-03-24
feat: add giskard-llm as lean litellm replacement

Hartorn:
## Summary

- Introduces a new `giskard-llm` library that replaces `litellm` as the default routing layer for `giskard-agents`, wrapping native provider SDKs (OpenAI, Google GenAI, Anthropic, Azure OpenAI, Azure AI Foundry) and eliminating 200+ transitive dependencies.
- Updates `giskard-agents` to import from `giskard.llm` via `GiskardLLMGenerator`, with pass-through extras so users install only the provider SDKs they need (`pip install giskard-agents[openai]`).
- Keeps `LiteLLMGenerator` available as a first-class alternative backend behind an optional `litellm` extra, so both backends can coexist in the same process.
- Adds provider-specific pytest marks (`@pytest.mark.google`, `@pytest.mark.litellm`, etc.) with auto-skip logic and a CI matrix that installs/tests each backend + provider independently.

### New: `libs/giskard-llm/`
- **Protocol-based provider system** (`providers/base.py`) — `CompletionProvider`, `EmbeddingProvider`, `ResponseProvider` protocols. Providers implement only the protocols they support (e.g. Anthropic implements only `CompletionProvider`).
- **Provider adapters** (`providers/openai.py`, `providers/google.py`, `providers/anthropic.py`, `providers/azure_openai.py`, `providers/azure_ai.py`) — thin wrappers over native SDKs with unified error mapping via `_map_error()`, structured output via native SDK features, and class-level `_PROVIDER` attribute for correct error attribution in Azure subclasses.
- **Response/Interactions API** — `ResponseProvider` protocol with `respond()` method, supporting OpenAI Responses API and Gemini Interactions API with unified `ResponseResult`, `ResponseOutputText`, `ResponseOutputFunctionCall` types. Tool definitions are converted from nested Chat Completions format to the flat format expected by each API.
- **Canonical tool result format** — `FunctionCallOutput` TypedDict as the unified input format for feeding back tool results. GoogleProvider normalizes this to `function_result` format internally.
- **Unified error hierarchy** (`errors.py`) — `LLMError` base with `AuthenticationError`, `RateLimitError`, `ServerError`, `LLMTimeoutError`, `BadRequestError`, `UnsupportedOperationError`, `ProviderNotAvailableError`, each carrying a `status_code` for retry decisions. `ProviderNotAvailableError` accepts an optional `extra` hint (e.g. `extra="azure"` → `pip install giskard-llm[azure]`).
- **Routing** (`routing.py`) — parses `"provider/model"` strings and dispatches to the correct provider adapter. `LLMClient` supports named configurations via `configure()`. Module-level `configure()`, `reset()`, `acompletion()`, `aembedding()`, `aresponse()` convenience functions.
- **Response types** (`types.py`) — Pydantic v2 models inheriting from `_BaseModel` (which defaults `model_dump` to `exclude_none=True`). `CompletionResponse`, `EmbeddingResponse`, `ResponseResult`. `ToolCallFunction.arguments` is `str` (JSON) for wire-compatible round-trips. TypedDicts (`ToolDef`, `ChatMessage`, `FunctionCallOutput`) for user-constructed inputs.
- **Retry helper** (`retry.py`) — class-based `RETRYABLE_ERRORS = frozenset({LLMTimeoutError, RateLimitError, ServerError})` and a `should_retry(exc)` helper — no HTTP-status branching at the retry layer.
- **Design documentation** (`docs/design.md`) — documents type conventions (TypedDict vs Pydantic), tool format differences, canonical tool result format, and key design decisions.
- **Unit tests** — cover routing, retry, types, errors, error mapping completeness, message conversion, input normalization, and mocked provider integration including protocol conformance checks.
- **Functional tests** — per-provider tests for completion, embedding, response API, tool roundtrips (hardcoded + parallel), multi-system messages, and behavior config, auto-skipped if the SDK extra isn't installed.

### Changed: `libs/giskard-agents/`
- `GiskardLLMGenerator` (renamed from `LiteLLMGenerator`) and `litellm_embedding_model.py` now import `acompletion`, `aembedding`, `should_retry` from `giskard.llm`.
- `pyproject.toml` replaces the `litellm` hard dep with `giskard-llm>=1.0.0a1` and re-exports its provider extras (`openai`, `google`, `anthropic`, `all`). A new `litellm` extra re-enables the optional `LiteLLMGenerator`.
- **`LiteLLMGenerator` restored** as a first-class alternative backend using the `litellm` SDK. Importing it without the extra raises `ImportError` with a clear install hint; it is lazily re-exported from `generators/__init__.py` via `__getattr__` so plain imports of `giskard.agents` never require `litellm`. `GiskardLLMGenerator` no longer claims the `"litellm"` discriminator kind — each class owns its own `kind`.
- Test fixtures updated to use `CompletionResponse` / `EmbeddingResponse` from `giskard.llm`.
- `conftest.py` adds a `pytest_collection_modifyitems` hook to auto-skip provider-marked tests when the SDK isn't installed, plus an autouse fixture to clear the provider cache between tests.
- New `test_litellm_generator.py` — unit tests for the LiteLLM adapter (mocked `acompletion`, retry middleware status-code mapping, discriminator round-trip).
- New `test_generator_backends.py` — functional tests exercising both backends through direct `complete()` and `ChatWorkflow`, plus a `test_backends_coexist_in_same_process` test that only runs when both SDKs are installed.

### Changed: CI
- `ci.yml` adds `giskard-llm` to the unit test matrix (Python 3.12–3.14) and a `test-no-providers` job for SDK-less error handling.
- `integration-tests.yml`:
  - `test-agents-functional` uses a **3-layout × 3-python matrix** (`giskard-llm` / `litellm` / `both`) installed via top-level `giskard-agents[...]` extras so the pass-through declarations are exercised.
  - Adds `test-llm-functional`: matrix over 5 providers (openai, google, anthropic, azure, azure_ai), secrets scoped to `ci` environment.
  - Adds `test-checks-functional`: matrix over google provider, secrets scoped to `ci` environment.
  - Adds `workflow_dispatch` trigger for manual runs with `head.sha || github.ref` fallback.
  - All checkout steps use `persist-credentials: false`.
- `labeler.yml` adds a `Scope: LLM` label for `libs/giskard-llm/**`.
- `Makefile` gains `test-no-providers` and `install-no-providers` targets.
- Default models: `gpt-4.1-nano` (OpenAI/Azure), `gemini-2.5-flash` (Google), `claude-haiku-4-5-20251001` (Anthropic). Azure OpenAI default `api_version` updated to `2024-10-21`.

### Bug fixes (from review follow-up)
- **Google Interactions API input normalization** — rewrote `_normalize_input_items` to convert all items to Google's preferred flat `ContentParam` format (Pattern A). `function_call` uses `id` (not `call_id`) with `arguments` as dict; `function_call_output` → `function_result` with resolved name; role-tagged turns → `TextContentParam`. Docstring links to SDK type definitions, official example, and known issue [#1906](https://github.com/googleapis/python-genai/issues/1906).
- **Google `_to_response_result`** — prioritize `output.id` over `output.call_id` for function call IDs (matching what Google actually returns).
- **Google `_convert_messages`** — resolve `tool_call_id` to actual function name from preceding assistant `tool_calls`; set `role="user"` for `role="tool"` messages.
- **Google `_map_error`** — handle `API_KEY_INVALID` string check for `AuthenticationError`; fall back to status 500 on SDK `APIError`.
- **Anthropic `_validate_messages`** — skip alternation check when either role is `"tool"` (consecutive tool messages are valid since they merge).
- **Anthropic `_convert_messages`** — merge consecutive `role="tool"` messages into a single user message with multiple `tool_result` blocks.
- **OpenAI `_normalize_input_item`** — serialize `function_call.arguments` from dict to JSON string (OpenAI Responses API rejects dicts).
- **OpenAI `_map_error`** — tolerate `APIConnectionError` missing a `status_code` attribute.
- Moved pyright ignores to the `from` line so they apply to optional SDK imports without leaking to the rest of the module.

### Misc
- `pyrightconfig.json` adds `libs/giskard-llm/src` to execution environments.
- Workspace `pyproject.toml` adds the new package.
- `uv.lock` regenerated after the `litellm` removal and follow-up dependency upgrades (anthropic, cryptography, google-genai, logfire-api, numpy, and other transitive bumps).
- `.secrets.baseline` refreshed to track new test-file line numbers already flagged as false positives.

## Test plan

- [x] All unit tests pass across `giskard-llm`, `giskard-agents`, `giskard-checks`, `giskard-core` (103 giskard-llm tests)
- [x] All pre-commit hooks pass (ruff check, ruff format, trailing-whitespace, end-of-file-fixer, detect-secrets)
- [x] `basedpyright` passes with 0 errors
- [x] CI green on this branch
- [x] Integration tests (last dispatch — squashed commit): https://github.com/Giskard-AI/giskard-oss/actions/runs/24555428273
- [x] Integration tests (previous run, all 18 jobs green): https://github.com/Giskard-AI/giskard-oss/actions/runs/24555298358

To re-run integration tests manually: [Dispatch workflow](https://github.com/Giskard-AI/giskard-oss/actions/workflows/integration-tests.yml) with ref `feat/giskard-llm`.

Made with [Cursor](https://cursor.com)


libs/giskard-llm/src/giskard/llm/errors.py
```diff
@@ -0,0 +1,48 @@
...
+Every error carries a ``status_code`` attribute so the retry middleware in
+giskard-agents (which does ``getattr(err, "status_code", 0)``) works
+unchanged regardless of the underlying provider.
+"""
+
+
+class LLMError(Exception):
+    """Base for all giskard-llm errors."""
+
+    def __init__(self, status_code: int, message: str, provider: str):
```
kevinmessiaen: Shall we have `status_code` is root `LLMError`?

This is tightly coupled to having HTTP-based provider. I know it's due to the `should_retry` logic which is status code driven but this create gap for some edge case:

[ConnectionResetError](https://docs.python.org/3/library/exceptions.html#ConnectionResetError), [TimoutError](https://docs.python.org/3/library/exceptions.html#TimeoutError) (Not the HTTP one)

Or for example local embeddings (numpy)?

It's fine if we don't plan, want to support those.

libs/giskard-llm/src/giskard/llm/retry.py
```diff
@@ -0,0 +1,8 @@
+"""Retry eligibility check for LLM provider errors."""
+
+RETRYABLE_STATUS_CODES = frozenset({408, 429, 500, 502, 503, 504, 520, 524, 529})
```
kevinmessiaen: I would have error based policy (see comment on errors). Which would allow more flexibility for future evolutions

RETRYABLE_ERROS = frozenset({TimeoutError, RateLimitError, ServerError})

libs/giskard-llm/src/giskard/llm/providers/anthropic.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+class AnthropicProvider(BaseProvider):
+    def __init__(
+        self,
+        api_key: str | None = None,
+        base_url: str | None = None,
+        timeout: float | None = None,
+        merge_system: bool = False,
+        **_kwargs: Any,
```
kevinmessiaen, line 102: Shall we log ignored kwargs? Might be typos like `timout` or stuff not supported for this provider

libs/giskard-llm/src/giskard/llm/providers/anthropic.py
```diff
@@ -0,0 +1,368 @@
...
+                    400, "System messages must have non-empty content.", PROVIDER
+                )
+
+    # -- helpers ---------------------------------------------------------------
+
+    def _build_completion_kwargs(
+        self,
+        model: str,
+        messages: Sequence[ChatMessage],
+        params: dict[str, Any],
```
kevinmessiaen: Again logging ignored params should improve debugging/spotting issues

libs/giskard-llm/src/giskard/llm/providers/azure_ai.py
```diff
@@ -0,0 +1,66 @@
...
+        resolved_key = api_key or os.environ.get("AZURE_AI_API_KEY")
+        resolved_base = base_url or os.environ.get("AZURE_AI_ENDPOINT")
+
+        client_kwargs: dict[str, Any] = {}
+        if resolved_key is not None:
+            client_kwargs["api_key"] = resolved_key
+        if resolved_base is not None:
+            client_kwargs["base_url"] = resolved_base
+        if timeout is not None:
+            client_kwargs["timeout"] = timeout
```
kevinmessiaen: Seems like we have this kind of compacting not `None` to dict a lot, we could probably use an utils:

```py
def compact(**kwargs: dict[str, Any]) -> dict[str, Any]:
    return {k: v for k, v in kwargs.items() if v is not None}
```

libs/giskard-llm/src/giskard/llm/providers/azure_ai.py
```diff
@@ -0,0 +1,70 @@
...
+        **_kwargs: Any,
+    ) -> None:
+        if _kwargs:
+            logger.warning(
+                "%s provider: ignoring unknown kwargs: %s", PROVIDER, sorted(_kwargs)
+            )
+        try:
+            import openai
+        except ImportError as exc:
+            raise ProviderNotAvailableError(PROVIDER, "openai") from exc
```
kevinmessiaen: This would ask to use `pip install giskard-llm[azure_ai]` to fix the problem while the flavor is defined as `azure`

libs/giskard-llm/tests/functional/test_response.py
```diff
@@ -0,0 +1,256 @@
...
+    resp = await client.aresponse(
+        model,
+        "What is 2+2? Use the add tool.",
+        tools=[_ADD_TOOL],
+    )
+    fc_outputs = [o for o in resp.outputs if isinstance(o, ResponseOutputFunctionCall)]
+    assert len(fc_outputs) > 0
+    fc = fc_outputs[0]
+    assert fc.name == "add"
+    assert "a" in fc.arguments and "b" in fc.arguments
```
kevinmessiaen, line 108: I would also add test for “tool roundtrip” where:

- the model returns tool calls
- the workflow adds tool result messages (role="tool" with tool_call_id)
- those messages are then fed back into the provider conversion/validation

kevinmessiaen (changes requested):

kevinmessiaen:
@Hartorn following code should leads to 4xx

[repro_google_tool_name.py](https://github.com/user-attachments/files/26740627/repro_google_tool_name.py)
[repro_google_tool_role.py](https://github.com/user-attachments/files/26740628/repro_google_tool_role.py)
[repro_anthropic_multiple_tool_results.py](https://github.com/user-attachments/files/26740794/repro_anthropic_multiple_tool_results.py)

kevinmessiaen (approved):

PR #2331, opened 2026-03-25
chore: Remove hard dependency on rich

kevinmessiaen:
Remove `rich` from mandatory dependencies


pyproject.toml
```diff
@@ -54,6 +54,7 @@ markers = [
 
 [dependency-groups]
 dev = [
```
davidberenstein1957, line 56: Why do we add it as a dependency? I do feel it is a feature people might want to use outside of the notebooks, etc., too?

Also, what do visualisation look like without rich? In my experience, they are really minimal.

PR #2346, opened 2026-03-27
security(giskard-agents)!: opt-in Jinja parsing for Workflow.chat

kevinmessiaen:
Plain strings passed to chat() are appended as literal Message content by default, avoiding server-side template injection when user input is forwarded to the model. Use as_template=True or MessageTemplate for Jinja-rendered strings.

BREAKING CHANGE: str arguments to ChatWorkflow.chat() and BaseGenerator.chat() are no longer parsed as Jinja2 templates unless as_template=True.

Refs: ENG-1488

Also sync uv.lock with current workspace package versions.


libs/giskard-agents/src/giskard/agents/workflow.py
```diff
@@ -255,11 +255,31 @@ class ChatWorkflow(BaseModel, Generic[OutputType]):
...
+        """Add a chat message to the workflow.
+
+        Parameters
+        ----------
+        message : str | Message | MessageTemplate
+            The message to add. Plain strings are appended as literal text by default.
+        role : Role, default "user"
+            The role for the message, used when ``message`` is a string.
+        as_template : bool, default False
+            When True, parse a string ``message`` as a Jinja2 template.
```
mattbit, line 273: Let's warn about the security implications of this in the docstring.

PR #2378, opened 2026-04-09
feat(telemetry): add optional PostHog analytics to giskard-core and checks

kevinmessiaen:
## Description

- Send aggregate, non-content metrics from scenario/suite/test-case runs
- Tag context with package versions and coarse environment (ci/colab/kaggle/local)
- Opt out via DO_NOT_TRACK, GISKARD_TELEMETRY_DISABLED, or disable_telemetry()
- Document collection and privacy limits in READMEs; disable telemetry in tests
- Add posthog dependency; EU-hosted PostHog endpoint

### DIFF
--- a/libs/giskard-core/src/giskard/core/telemetry/telemetry.py
+++ b/libs/giskard-core/src/giskard/core/telemetry/telemetry.py
@@ -81,13 +81,17 @@ def _set_tags() -> None:
         tag(key, value)
 
 
-def _get_or_create_anonymous_id() -> str:
+def _get_or_create_anonymous_id() -> str | None:
     if _should_disable():
-        return "disabled"
+        return None
 
     config_path = Path.home() / ".giskard" / "id"
     if config_path.exists():
-        return config_path.read_text(encoding="utf-8").strip()
+        try:
+            return config_path.read_text(encoding="utf-8").strip()
+        except OSError:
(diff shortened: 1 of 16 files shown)


libs/giskard-core/src/giskard/core/telemetry/telemetry.py
```diff
@@ -0,0 +1,171 @@
...
+
+def _set_tags() -> None:
+    env_information = _get_env_information()
+    for key, value in env_information.items():
+        tag(key, value)
+
+
+def _get_or_create_anonymous_id() -> str:
+    if _should_disable():
+        return "disabled"
```
henchaves:
```suggestion
        return None
```

libs/giskard-core/src/giskard/core/telemetry/telemetry.py
```diff
@@ -0,0 +1,171 @@
...
+
+    config_path = Path.home() / ".giskard" / "id"
+    if config_path.exists():
+        return config_path.read_text(encoding="utf-8").strip()
```
henchaves: should also handle read access errors

henchaves (approved):

henchaves:
LGTM, only mentioned a small issue - optional to tackle

PR #2396, opened 2026-04-13
feat(checks): add operator composition and utility functions

### LINKED ISSUE #2395: Improve ergonomics for composing checks in scenarios
## Problem

Building composed checks today is verbose and reads heavier than the logical structure of the assertion. Authors must always pass explicit keyword arguments (`checks=[...]`, `check=...`) and nest constructors, which obscures intent when several primitives are combined.

**Current shape** (see `test.py`):

```python
.check(
    AllOf(
        checks=[
            Equals(expected_value="...", key="trace.last.outputs"),
            Not(
                check=Equals(
                    expected_value="...",
                    key="trace.last.outputs",
                )
            ),
        ]
    )
)
```

This works but does not scale nicely for quick scenario authoring, and it is easy to mis-order or mis-nest `checks=` vs `check=`.

## Goal

Make common compositions (all / any / not) feel natural at the call site while preserving the existing discriminated `Check` model, serialization, and runtime behavior.

**Desired direction** (illustrative — exact API to be decided in implementation):

- utility methods, e.g. `all_of(eq_a, expect_fails(eq_b))` instead of only `AllOf(checks=[...])`.
- operator composition (e.g. `check_a & check_b` for all, `check_a | check_b` for any), document semantics (short-circuit, precedence) to match `AllOf` / `AnyOf` today.

> Note: The following sketch is **not** valid Python as written; it only conveys the *readability* target. Implementation should use real check combinators (`AllOf`, `AnyOf`, `Not`), not `and` / `not` on check instances.

```python
from giskard.checks import Scenario, expect_fails, Equals, expect_all
import asyncio

test_scenario = (
    Scenario("test_france_capital")
    .interact(
        inputs="What is the capital of France?",
        outputs="The capital of France is Paris.",
    )
    .check(
        all_of(
            Equals(
                expected_value="The capital of France is Paris.",
                key="trace.last.outputs",
            ),
            expect_fails(
                Equals(
                    expected_value="The capital of France is Paris.",
                    key="trace.last.outputs",
                )
            ),
        )
    )
)

result = asyncio.run(test_scenario.run())
result.print_report()
```

### Lazy import / forward references on `Check`

Any ergonomics that tie the base `Check` type to concrete combinators (**especially anything resembling “or” / `AnyOf`**, or operators implemented on `Check`) must avoid import cycles and import-time failures.

- The generic base lives at `libs/giskard-checks/src/giskard/checks/core/check.py` (the `Check[InputType, OutputType, TraceType]` class).
- **Requirement:** use **lazy imports** (e.g. import inside `__or__` / factory methods), `TYPE_CHECKING` guards, string annotations, or another agreed pattern so that `check.py` stays safe to import before compositional subclasses are loaded.
- Do not regress Pydantic discriminated serialization for composed checks.

## Acceptance criteria

- [ ] Composing multiple checks for `.check(...)` is measurably less noisy than the current `checks=[...]` / `check=...` pattern for the common case, without losing type clarity where we already have it.
- [ ] Behavior of short-circuiting and messages matches existing `AllOf`, `AnyOf`, and `Not` semantics.
- [ ] Adding “or” / `AnyOf`-related convenience on or near `Check` does **not** introduce import-time failures; lazy (deferred) import strategy is documented in code or module docstring where non-obvious.

## References

- Combinators: `libs/giskard-checks/src/giskard/checks/builtin/composition.py`
- Base type: `libs/giskard-checks/src/giskard/checks/core/check.py` (`Check` class, ~L9–10)

mvanhorn:
## Summary

Adds ergonomic composition for checks via Python operators and utility functions, as requested in #2395.

### Operator overloads on `Check` base class (`core/check.py`)
- `check_a & check_b` creates `AllOf(checks=[check_a, check_b])` (flattens nested AllOf)
- `check_a | check_b` creates `AnyOf(checks=[check_a, check_b])` (flattens nested AnyOf)
- `~check_a` creates `Not(check=check_a)`

### Utility functions (`builtin/composition.py`)
- `all_of(*checks)` - variadic AllOf constructor
- `any_of(*checks)` - variadic AnyOf constructor
- `expect_fails(check)` - readable alias for Not

### Exports
- All three utilities exported from `giskard.checks`

## Before/After

```python
# Before
AllOf(checks=[Equals(expected_value="x", key="k"), Not(check=Equals(expected_value="y", key="k"))])

# After (utilities)
all_of(Equals(expected_value="x", key="k"), expect_fails(Equals(expected_value="y", key="k")))

# After (operators)
Equals(expected_value="x", key="k") & ~Equals(expected_value="y", key="k")
```

The existing `AllOf`, `AnyOf`, `Not` classes and their serialization are unchanged. Operators and utilities produce the same model instances.

Closes #2395

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/__init__.py
+++ b/libs/giskard-checks/src/giskard/checks/__init__.py
@@ -20,6 +20,9 @@ from .builtin import (
     RegexMatching,
     SemanticSimilarity,
     StringMatching,
+    all_of,
+    any_of,
+    expect_fails,
     from_fn,
 )
 from .core import (
@@ -107,6 +110,10 @@ __all__ = [
     "SemanticSimilarity",
     "StringMatching",
     "RegexMatching",
+    # Composition utilities
+    "all_of",
+    "any_of",
+    "expect_fails",
     # Generators
     "UserSimulator",
     # Testing
--- a/libs/giskard-checks/src/giskard/checks/builtin/__init__.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/__init__.py
@@ -14,7 +14,7 @@ from .comparison import (
 )
 
 # Import other builtin checks (staying in builtin)
-from .composition import AllOf, AnyOf, Not
+from .composition import AllOf, AnyOf, Not, all_of, any_of, expect_fails
 from .fn import FnCheck, from_fn
 from .semantic_similarity import SemanticSimilarity
 from .text_matching import RegexMatching, StringMatching
@@ -23,6 +23,9 @@ __all__ = [
     "AllOf",
     "AnyOf",
     "Not",
+    "all_of",
+    "any_of",
+    "expect_fails",
     "from_fn",
     "FnCheck",
     "StringMatching",
--- a/libs/giskard-checks/src/giskard/checks/builtin/composition.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/composition.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import override
 
 from pydantic import Field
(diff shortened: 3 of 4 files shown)


mvanhorn:
Applied in 730102b - swapped both `__class__.__name__` checks for `isinstance()`. AllOf/AnyOf are already imported inside the methods, so no scope change needed. The redundant `hasattr(self, "checks")` guard is also gone since `isinstance(x, AllOf)` already implies `.checks` exists.

davidberenstein1957:
Thank you for the contribution but  it seems this has already been included in the library.

PR #2401, opened 2026-04-13
Allow scenarios to run multiple times

### LINKED ISSUE #2398: Allow running a Scenario multiple times (`multiple_runs`)
## Problem

LLM-driven scenarios can be flaky: the same scenario may pass/fail depending on generation randomness or external variance. We need a built-in way to **repeat the same scenario run (n) times and stop on the first failure**, to increase confidence and to let generator-based scenarios explore different approaches across attempts.

## Proposal

Introduce `multiple_runs: int = 1` that makes a scenario run execute **up to (n) attempts**, **stopping early on first failure**.

Support configuration at **two levels**:

* **Scenario level**: default repeat count for that scenario.
* **ScenarioRun level**: per-run override (takes precedence over scenario default).

### Precedence

* **ScenarioRunner.run(multiple_runs)** (if set) **overrides**
* **Scenario.multiple_runs** (if set) **defaults to**
* **1**

## Expected behavior

* If `multiple_runs = n`:
  * Execute attempt 1..n sequentially.
  * If an attempt **fails**: stop immediately and return that failure result.
  * If all attempts **pass**: return a passing result.
* Each attempt should be treated as an independent execution of the scenario run (so generator-based flows can vary).
* `multiple_runs` must be validated:
  * `n >= 1` required (reject 0/negative with a clear error).

## API / Data model changes

* Add optional `multiple_runs` to `Scenario` (default 1).
* Add optional `multiple_runs` to `ScenarioRunner.run/Scenario.run` (default: `None`, meaning “inherit from Scenario”).
* Ensure serialization/printing (if any exists for these objects) includes the configured repeat count.

## Result reporting (minimum requirement)

* Keep backward-compatible overall pass/fail semantics.
* On failure: surface the failing attempt’s details as the primary result.
* Include metadata on the result (or in logs) indicating:
  * configured `multiple_runs`
  * how many attempts actually executed
  * which attempt failed (if any)

## Acceptance criteria

* Can set `multiple_runs` on a `Scenario` and it repeats runs accordingly.
* Can set `multiple_runs` on a `ScenarioRunner.run/Scenario.run` and it overrides scenario-level setting.
* Stops on first failure.
* Defaults to a single run when unset everywhere.
* Invalid values (0, negative, non-int) raise a clear error.
* Unit tests cover:
  * scenario-level repetition
  * run-level override
  * stop-on-first-fail short-circuit
  * “all pass” across multiple attempts
  * validation errors

## Example (intended usage)

* Scenario-level:

```python
scenario = Scenario(..., multiple_runs=5)
```

* ScenarioRunner.run/Scenario.run override:

```python
run = Scenario().run(..., multiple_runs=10)  # overrides scenario default
```

raghavtries:
## Description

Adds `multiple_runs` support for scenarios so flaky or generator-driven scenarios can be executed repeatedly before reporting success.

The repeat count can be configured at the scenario level with `Scenario(..., multiple_runs=n)` or overridden per execution with `Scenario.run(..., multiple_runs=n)` / `ScenarioRunner.run(..., multiple_runs=n)`. Each attempt runs independently with a fresh trace, and execution stops early on the first non-passing attempt.

Scenario results now include repeat metadata:

- configured `multiple_runs`
- `attempts_executed`
- `failed_attempt` when execution stops early

## Related Issue

Closes #2398

## Type of Change

- [ ] 🔐 Security fix

## Checklist

- [x] I've updated the `uv.lock` running `uv lock` (only applicable when `pyproject.toml` has been
  modified)

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/core/result.py
+++ b/libs/giskard-checks/src/giskard/checks/core/result.py
@@ -234,7 +234,7 @@ class ScenarioResult[TraceType: Trace](BaseResult, frozen=True):  # pyright: ign
     final_trace : TraceType
         Trace state after execution, containing all interactions that occurred.
     multiple_runs : int
-        Configured number of attempts for this scenario execution.
+        Configured number of runs for this scenario execution.
     runs_executed : int
         Number of runs that actually executed.
     status : ScenarioStatus
@@ -255,7 +255,7 @@ class ScenarioResult[TraceType: Trace](BaseResult, frozen=True):  # pyright: ign
     final_trace: TraceType = Field(..., description="Final trace state after execution")
     multiple_runs: int = Field(
         default=1,
-        description="Configured number of scenario execution attempts.",
+        description="Configured number of runs for this scenario execution.",
     )
     runs_executed: int = Field(
         default=1,
--- a/libs/giskard-checks/src/giskard/checks/core/result.py
+++ b/libs/giskard-checks/src/giskard/checks/core/result.py
@@ -235,10 +235,8 @@ class ScenarioResult[TraceType: Trace](BaseResult, frozen=True):  # pyright: ign
         Trace state after execution, containing all interactions that occurred.
     multiple_runs : int
         Configured number of attempts for this scenario execution.
-    attempts_executed : int
-        Number of attempts that actually ran.
-    failed_attempt : int or None
-        One-based attempt number that failed, if execution stopped early.
+    runs_executed : int
+        Number of runs that actually executed.
     status : ScenarioStatus
         Aggregated outcome of the scenario derived from its steps.
     passed : bool
@@ -259,13 +257,9 @@ class ScenarioResult[TraceType: Trace](BaseResult, frozen=True):  # pyright: ign
         default=1,
         description="Configured number of scenario execution attempts.",
     )
-    attempts_executed: int = Field(
+    runs_executed: int = Field(
         default=1,
-        description="Number of scenario execution attempts that actually ran.",
-    )
-    failed_attempt: int | None = Field(
-        default=None,
-        description="One-based attempt number that failed, if any.",
+        description="Number of scenario runs that actually executed.",
     )
 
     @computed_field
@@ -330,7 +324,7 @@ class ScenarioResult[TraceType: Trace](BaseResult, frozen=True):  # pyright: ign
 
         yield Rule(
             f"{_pluralize(len(self.steps), 'step')} in {self.duration_ms}ms"
-            f" | attempts: {self.attempts_executed}/{self.multiple_runs}",
+            f" | runs: {self.runs_executed}/{self.multiple_runs}",
             style=f"{status['color']} bold",
         )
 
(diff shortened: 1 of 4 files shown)


libs/giskard-checks/src/giskard/checks/core/scenario.py
```diff
@@ -98,6 +100,19 @@ class Scenario[InputType, OutputType, TraceType: Trace](BaseModel):  # pyright:
...
+    )
+
+    @field_validator("multiple_runs", mode="before")
+    @classmethod
+    def _validate_multiple_runs(cls, value: object) -> int:
+        if not isinstance(value, int) or isinstance(value, bool):
+            raise ValueError("multiple_runs must be an integer greater than or equal to 1")
+        if value < 1:
+            raise ValueError("multiple_runs must be greater than or equal to 1")
+        return value
```
kevinmessiaen: We can just use builtin `ge` constraint

```suggestion
    multiple_runs: int = Field(
        default=1,
        description="Default number of times to run this scenario before reporting success.",
        ge=1,
    )
```

libs/giskard-checks/src/giskard/checks/core/result.py
```diff
@@ -233,6 +233,12 @@ class ScenarioResult[TraceType: Trace](BaseResult, frozen=True):  # pyright: ign
         Total execution time in milliseconds.
     final_trace : TraceType
         Trace state after execution, containing all interactions that occurred.
+    multiple_runs : int
+        Configured number of attempts for this scenario execution.
+    attempts_executed : int
+        Number of attempts that actually ran.
+    failed_attempt : int or None
+        One-based attempt number that failed, if execution stopped early.
```
kevinmessiaen: The fields names seems a bit confusing, we mix up with `runs` and `attempts`. I would use run everywhere. Furthermore `multiple_runs` here might not be the best. Maybe `scheduled_runs` or `target_runs` might work better.

IMO attempts is better suited when we re-execute until it passes (flaky tests). In our case each scenario execution are excepted to pass, so I would use run.

Regarding `failed_attempt` it's misleading. It's the failed index, which is always equals to `attempts_executed` if failed or None. I would probably drop this variable which can be inferred.

libs/giskard-checks/src/giskard/checks/core/result.py
```diff
@@ -311,7 +329,8 @@ def __rich_console__(
             yield repr(self.final_trace)
 
         yield Rule(
-            f"{_pluralize(len(self.steps), 'step')} in {self.duration_ms}ms",
+            f"{_pluralize(len(self.steps), 'step')} in {self.duration_ms}ms"
+            f" | attempts: {self.attempts_executed}/{self.multiple_runs}",
```
kevinmessiaen:
```suggestion
            f" | runs: {self.attempts_executed}/{self.multiple_runs}",
```

kevinmessiaen (changes requested):
Thanks for you contribution! Looking good overall.

We should address the naming discrepancy issue as well as the docstring

raghavtries:
Thanks for the feedback! I changed it to have:

- including runs_executed instead of attempts_executed
- there is no separate failed_attempt field, since the failed run is inferable from runs_executed when the result is non-passing
- multiple_runs validation through the Pydantic field constraint
- including run terminology in the output and docstrings
- resolved another conflict while keeping the new telemetry instrumentation

Please let me know if there should be any other changes!

kevinmessiaen (approved):
Thanks for the changes, looking good!

PR #2406, opened 2026-04-16
feat: checks trace otel support

kevinmessiaen:
## Summary

Adds parsing of **OpenTelemetry GenAI semantic-convention** log events into a structured **`GenAiTrace`** (messages and choices per turn), optional **deduplication** of repeated conversation history emitted by some instrumentations, and **optional `giskard-llm` extras** that bundle each provider SDK with the matching OTEL instrumentation packages.

[Semantic Conventions for Generative AI Events](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/gen-ai/gen-ai-events.md)

## What changed

- **`giskard-checks`**: New module `giskard.checks.core.interaction.gen_ai` with `GenAiTrace.from_otel_logs()` to build a `Trace` of `Interaction`s from ordered `event_name` + `body` dicts (e.g. `gen_ai.user.message`, `gen_ai.assistant.message`, `gen_ai.tool.message`, `gen_ai.choice`). Rich rendering helpers for messages and choices. Normalization for Anthropic-style choice payloads missing `role` on the nested assistant message.
- **`giskard-llm`**: New optional extras `openai-otel`, `google-otel`, `anthropic-otel`, `azure-otel`, and `azure_ai-otel` (SDK + OTEL instrumentation + SDK + `wrapt<2` pin where needed). Lockfile updated accordingly.
- **Tests**: Unit tests for `GenAiTrace.from_otel_logs` (multiple choices per request, two-turn flow, redundant history stripping). Functional tests that run real completions per provider, capture logs with in-memory exporters, and assert trace shape against API responses.
- **CI**: `integration-tests` job for **giskard-checks** functional tests runs a **provider matrix** (openai, google, anthropic, azure, azure_ai) on Python **3.12**, installs `giskard-llm[$PROVIDER,$PROVIDER-otel]`, and wires env vars for each provider’s API keys and Azure endpoints.
- **Tooling**: `.secrets.baseline` updates for detect-secrets (baseline self-ignore and placeholder pattern).

## How to test locally

- Unit: `make test-unit PACKAGE=giskard-checks` (or full `make test-unit`).
- Functional (requires keys and OTEL extras): install e.g. `giskard-llm[google,google-otel]`, set `GOOGLE_API_KEY`, then `make test-functional PACKAGE=giskard-checks PROVIDER=google` (and similarly for other providers).

## Notes

- GenAI semconv and instrumentation payloads **may change** across OTEL versions; `raise_on_unknown_event` defaults to `True` to surface unsupported events early.
- Functional tests set `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` so message bodies appear in captured logs.


kevinmessiaen:
@gemini-code-assist can you check again it was targeting the wrong branch

kevinmessiaen:
@gemini-code-assist can you review

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+    type: str
+
+    def __rich_console__(
+        self, console: Console, options: ConsoleOptions
+    ) -> RenderResult:
+        yield "[dim]No details available[/dim]"
+
+
+class FunctionCallLike(ToolCallLike, frozen=True, extra="allow"):
+    type: Literal["function_call"] = "function_call"
```
Hartorn: **Critical — `type: Literal["function_call"]` doesn't match what any instrumentor actually emits.**

Every `-otel` extra we install emits `"type": "function"` inside tool-call objects on `gen_ai.choice` events — not `"function_call"`:
- `opentelemetry-instrumentation-openai-v2` forwards `ChatCompletionMessageToolCall.type` which is `Literal["function"]`.
- `opentelemetry-instrumentation-google-genai` emits `{"type": "function", "function": {...}}`.
- `opentelemetry-instrumentation-anthropic` (Traceloop) emits `"function"` after its own normalization.

Because `ToolCallLike` is the parent class with `extra="allow"`, the `FunctionCallLike | ToolCallLike` union silently degrades to `ToolCallLike` and the `function` payload is absorbed into the raw-dict bucket instead of being exposed as `FunctionLike`.

**CI reproduces this on the dispatched run 24558275847**: `test_completion_with_tools[openai]` and `[azure]` both fail on `libs/giskard-checks/tests/functional/test_genai.py:242` — `assert isinstance(tc0, FunctionCallLike)` is False.

**Fix** — change the literal to `"function"` (and the default value accordingly):

```python
class FunctionCallLike(ToolCallLike, frozen=True, extra="allow"):
    type: Literal["function"] = "function"
    function: FunctionLike
```

And add a unit test that feeds a realistic OpenAI-shaped `gen_ai.choice` body through `GenAiTrace.from_otel_logs` and asserts `isinstance(trace.interactions[0].outputs[0].message.tool_calls[0], FunctionCallLike)` — doesn't require network, catches regression.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+    return {"role": role} | body
+
+
+def _normalize_choice_body_for_semconv(body: dict[str, Any]) -> dict[str, Any]:
+    """Anthropic instrumentation omits assistant ``role`` in the nested message (semconv)."""
+    choice_body = dict(body)
+    nested = choice_body.get("message")
+    if isinstance(nested, dict) and "role" not in nested:
+        choice_body["message"] = {"role": "assistant", **nested}
+    return choice_body
```
Hartorn: **Critical — Anthropic `gen_ai.choice` places `tool_calls` at the choice body top level, not inside `message`.**

The Traceloop Anthropic instrumentor (`opentelemetry-instrumentation-anthropic`) emits:

```json
{
  "message": {"content": "..."},
  "tool_calls": [{"id": "...", "type": "function", "function": {...}}],
  "finish_reason": "tool_use"
}
```

Not the OTel GenAI semconv shape that OpenAI v2 and Google emit:

```json
{
  "message": {"role": "assistant", "content": "...", "tool_calls": [...]},
  "finish_reason": "..."
}
```

The current normalizer only patches the missing `message.role`; the top-level `tool_calls` is silently absorbed by `ChoiceLike.extra="allow"` and `ChoiceLike.message.tool_calls` stays `None`. Anthropic tool-call data is effectively lost from the parsed trace even though it is present in the event.

**Fix** — extend the normalizer to lift top-level `tool_calls` into `message`:

```python
def _normalize_choice_body_for_semconv(body: dict[str, Any]) -> dict[str, Any]:
    """Normalize vendor quirks in gen_ai.choice event body.

    - Anthropic omits role inside the nested message (semconv).
    - Anthropic emits tool_calls at the choice body top level; lift them into message.
    """
    choice_body = dict(body)
    nested = choice_body.get("message")
    if isinstance(nested, dict):
        if "role" not in nested:
            nested = {"role": "assistant", **nested}
        top_level_tool_calls = choice_body.pop("tool_calls", None)
        if top_level_tool_calls is not None and "tool_calls" not in nested:
            nested = {**nested, "tool_calls": top_level_tool_calls}
        choice_body["message"] = nested
    return choice_body
```

**Test** — feed a Traceloop-shaped Anthropic `gen_ai.choice` body through `from_otel_logs` and assert `trace.interactions[-1].outputs[-1].message.tool_calls` is non-empty (no network needed).

**Not yet reproduced in CI** — the `checks/anthropic` job short-circuits earlier on the `TextMessageLike.content: str` vs. list-of-blocks issue (line 24–26) before any `gen_ai.choice` is processed. Once that is fixed, this one will surface next.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+            elif event_name == "gen_ai.tool.message":
+                _flush_if_has_choices()
+                inputs.append(
+                    ToolMessageLike.model_validate(_body_with_role("tool", body))
+                )
+            elif event_name == "gen_ai.choice":
+                choice = ChoiceLike.model_validate(
+                    _normalize_choice_body_for_semconv(body)
+                )
+                choices.append(choice)
```
Hartorn: **Critical — Google GenAI default-mode event shape is incompatible with the parser, in three distinct ways.**

`opentelemetry-instrumentation-google-genai` (default / content-recording enabled) emits event bodies that don't match the OpenAI-v2-shaped semconv this parser assumes. Source: `generate_content.py::_maybe_log_single_user_prompt` in the instrumentor:

```python
body = {}
if self._content_recording_enabled:
    logged_contents = contents
    if isinstance(contents, list):
        logged_contents = Content(parts=contents)
    body["content"] = _to_dict(logged_contents)   # dict, not str
self._otel_wrapper.log_user_prompt(attributes=attributes, body=body)
```

And for choices, `_maybe_log_response_candidate` emits `body = {"index": ..., "content": <Content dict>, "finish_reason": "STOP"}` — **no `message` key at all**.

**Three independent failures:**

1. `TextMessageLike.content: str` rejects the `Content` dict → reproduced live in CI (dispatched run 24558275847): `test_user_only[google]` and `test_completion_with_tools[google]` fail with:
    ```
    ValidationError: 1 validation error for TextMessageLike
    content
      Input should be a valid string, input_value={'parts': [{'text': '...'}], 'role': 'user'}
    ```
2. `ChoiceLike.message` is required; Google's choice body has a flat `content` with no `message` key — fails even after (1).
3. Google never emits `gen_ai.assistant.message` or `gen_ai.tool.message` in default mode (grep `otel_wrapper.py`: only `log_system_prompt`, `log_user_prompt`, `log_response_content`). Multi-turn prior assistant / tool turns in a conversation are **silently dropped** from the trace, which also breaks `drop_redundant_input_history` accounting.

**Fix** — add a Google-specific normalization path in `from_otel_logs`:
- Stringify `Content` dicts by concatenating `parts[*].text` (drop function-call / function-response parts to `tool_calls` / `tool.message`).
- Rewrite choice body to the OpenAI-v2 shape: `{"index", "finish_reason", "message": {"role": "assistant", "content": <text>, "tool_calls": [...]}}`.
- Decide how to represent Google's merged user+tool_result inputs (they arrive in the same `user` message as function_response parts).

**Test** — fixture `gen_ai.user.message` and `gen_ai.choice` bodies built straight from the Google instrumentor source (one with plain text, one with function-call / function-response parts) and assert the parsed `GenAiTrace` has the right role/content/tool_calls.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+    "assistant": "bold yellow",
+    "tool": "bold purple",
+}
+
+_DEFAULT_BORDER_STYLE = "bold gray"
+
+
+class TextMessageLike(BaseModel, frozen=True, extra="allow"):
+    role: str
+    content: str
```
Hartorn: **Critical — `content: str` rejects Anthropic inputs whose content is a list of blocks.**

Traceloop `emit_input_events` forwards `message.get("content")` verbatim. In the Anthropic SDK, `messages[].content` is commonly a list of block dicts:

```json
[
  {"type": "text", "text": "..."},
  {"type": "tool_use", "id": "...", "name": "...", "input": {...}},
  {"type": "tool_result", "tool_use_id": "...", "content": "..."}
]
```

And `system=` on the Anthropic API can also be a list of text blocks.

`TextMessageLike.content: str` rejects anything non-string, so any multi-turn Anthropic trace whose history contains a previous assistant `tool_use` or a user `tool_result` fails validation on the input replay — independent of the top-level `tool_calls` normalization issue (line 156–162).

**CI reproduces this on dispatched run 24558275847**: `test_completion_with_tools[anthropic]` fails with:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for TextMessageLike
  at libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py:251
```

**Fix options** (pick one):

- *Normalize at parse time*: concatenate `text` block `.text` fields into a single string, synthesize `gen_ai.tool.message` events from `tool_result` blocks, lift `tool_use` blocks into the previous assistant message's `tool_calls`. Keeps the `TextMessageLike` shape clean for downstream checks.
- *Widen the model*: change `content: str` to `content: str | list[BlockLike]` with a block-aware renderer in `__rich_console__`. Simpler, but pushes block-awareness onto every consumer.

**Test** — feed a list-of-blocks `gen_ai.user.message` body and a `gen_ai.assistant.message` body containing a `tool_use` block through `from_otel_logs` and assert the shape of the resulting `inputs`.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+                        _body_with_role("assistant", body)
+                    )
+                )
+            elif event_name == "gen_ai.tool.message":
+                _flush_if_has_choices()
+                inputs.append(
+                    ToolMessageLike.model_validate(_body_with_role("tool", body))
+                )
+            elif event_name == "gen_ai.choice":
+                choice = ChoiceLike.model_validate(
```
Hartorn: **Critical — Traceloop Anthropic emits one `gen_ai.choice` per content _block_, not per _candidate_.**

This branch assumes consecutive `gen_ai.choice` events represent `n>1` candidates of the same request. That is true for OpenAI v2 and Google GenAI, but not for Anthropic.

Traceloop `emit_response_events` iterates `response.content` and emits one `ChoiceEvent` per block:

```python
elif response.get("content"):
    for i, completion in enumerate(response.get("content")):
        ...
        emit_event(ChoiceEvent(index=i, ...), event_logger)
```

A single Claude response whose `content` is `[text_block, tool_use_block]` produces two `gen_ai.choice` events with `index=0` and `index=1`. The parser accumulates them into `Interaction.outputs` as if the model returned two candidates — but semantically it is **one** assistant turn with both text and tool_calls.

Downstream checks that interpret `len(outputs) == n` read Anthropic traces wrong. The semconv is explicit: *"each message corresponds to exactly one generation (choice/candidate)"* — this is a provider quirk, not spec behavior.

**Fix** — Anthropic-specific merge (don't touch the OpenAI/Google path):
- If the instrumentor attaches `gen_ai.response.id` on the choice body, fold consecutive choices that share the same id into one `AssistantMessageLike`.
- Otherwise: merge whenever the previous choice's `finish_reason` is not a terminal stop reason.

**Test** — fixture with a two-block Anthropic response (text + tool_use) replayed as two `gen_ai.choice` events; assert `len(trace.interactions[-1].outputs) == 1` and the single `ChoiceLike.message` has both `content` and `tool_calls`.

**Not yet reproduced in CI** — the `checks/anthropic` job short-circuits earlier on the `TextMessageLike.content: str` vs. list-of-blocks issue (line 24–26) before any multi-block choice is processed. After the `tool_calls` normalization (line 156–162) and that `content` fix land, this becomes the next failure.

Related tracking: https://github.com/traceloop/openllmetry/issues/3515

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+                )
+
+
+# Backward-compatible name for assistant messages that include tool calls.
+ToolCallMessageLike = AssistantMessageLike
+
+
+class ToolMessageLike(TextMessageLike, frozen=True, extra="allow"):
+    role: str = "tool"
+    id: str
```
Hartorn: **High — `ToolMessageLike.id` is required, but the semconv and the instrumentor both allow it to be absent/null.**

OTel GenAI semconv [gen-ai-events.md](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-events.md) specifies the `gen_ai.tool.message` body as:

```
{ "content": "string", "id": "string | null" }
```

And `opentelemetry-instrumentation-openai-v2::message_to_event` only sets `body["id"]` if `tool_call_id` is truthy; otherwise the key is absent:

```python
if tool_call_id:
    body["id"] = tool_call_id
```

Any client whose tool message is missing or null `tool_call_id` (legal per OpenAI SDK — you can call a "tool" without an id in some flows) will crash trace parsing with `ValidationError` at `ToolMessageLike.model_validate(...)`.

**Fix:**

```python
class ToolMessageLike(TextMessageLike, frozen=True, extra="allow"):
    role: str = "tool"
    id: str | None = None
```

**Test** — feed a `gen_ai.tool.message` body without `id` (and one with `id: null`) and assert the parsed `ToolMessageLike.id` is `None`, not a raise.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+            repeated assistant turn.
+
+        Trailing ``gen_ai.*.message`` events without a following ``gen_ai.choice``
+        (e.g. incomplete log streams or interrupted requests) are ignored after
+        logging a warning; successfully parsed interactions are still returned.
+
+        Returns
+        -------
+        GenAiTrace
+            Parsed trace of interactions (inputs and choice outputs per turn).
```
Hartorn: **High — parser targets the legacy per-role event model; OTel GenAI semconv is migrating to a single-event schema, and both OpenAI v2 and Google instrumentors already emit it under the standard opt-in flag.**

The canonical [gen-ai-events.md](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-events.md) on `main` now defines only two events:

- `gen_ai.client.inference.operation.details`
- `gen_ai.evaluation.result`

The per-role `gen_ai.{system,user,assistant,tool}.message` + `gen_ai.choice` events this parser assumes are the **pre-1.37 pattern**, retained by instrumentors for the `DEFAULT` opt-in mode. When a user sets the standard, documented environment flag

```
OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental
```

…the OpenAI v2 and Google GenAI instrumentors switch to emitting a single `gen_ai.client.inference.operation.details` `LogRecord` carrying `gen_ai.input.messages` / `gen_ai.output.messages` / `gen_ai.system_instructions` attributes.

- OpenAI v2: `chat_completions_create_v_new` path uses `TelemetryHandler` which emits the new single event (see `OpenAIInstrumentor._instrument`, `is_experimental_mode()` branch).
- Google GenAI: `_maybe_log_completion_details` emits `event_name="gen_ai.client.inference.operation.details"` when `sem_conv_opt_in_mode == _StabilityMode.GEN_AI_LATEST_EXPERIMENTAL`.

**Impact:** the moment a user flips that flag — and they will, because it is the recommended path forward — `raise_on_unknown_event=True` raises, or (worse) `raise_on_unknown_event=False` silently returns an empty trace. Traceloop's Anthropic [PR #3515](https://github.com/traceloop/openllmetry/issues/3515) tracks the same migration.

**Options:**

- (a) **Hard-lock on legacy for now**: update the docstring to state explicitly that this parser targets the pre-1.37 per-role event model, and emit a clear error message if `event_name == "gen_ai.client.inference.operation.details"` is seen. Ship the parser-shape fixes flagged elsewhere in this review (literal mismatch, Anthropic `tool_calls` normalization, Anthropic block content, Google event shapes, multi-block accumulation) first; handle the new schema in a follow-up.
- (b) **Add a second entry point**: `GenAiTrace.from_otel_inference_details(record)` that reads `gen_ai.input.messages` / `gen_ai.output.messages` directly; keep `from_otel_logs` as the legacy path. Future-proof, but roughly doubles parser surface area.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+                )
+            elif event_name == "gen_ai.tool.message":
+                _flush_if_has_choices()
+                inputs.append(
+                    ToolMessageLike.model_validate(_body_with_role("tool", body))
+                )
+            elif event_name == "gen_ai.choice":
+                choice = ChoiceLike.model_validate(
+                    _normalize_choice_body_for_semconv(body)
+                )
```
Hartorn: **High — `gen_ai.system.message` isn't always emitted, and roles outside the hardcoded list (notably `developer`) make the parser raise.**

`opentelemetry-instrumentation-openai-v2::message_to_event` builds the event name as `f"gen_ai.{role}.message"`. If a caller passes `role="developer"` (the current OpenAI default for instructions with the reasoning / o-series models), the emitted event name is `gen_ai.developer.message`. The current dispatch treats it as unknown and either raises (`raise_on_unknown_event=True`, the default) or silently drops the instruction message (`False`).

Two real-world paths hit this:

1. Any OpenAI chat completion using `role="developer"` — immediately crashes trace parsing.
2. For Google, `_maybe_log_system_instruction` only logs `gen_ai.system.message` when content recording is on **and** `content.parts` has text; otherwise no system event is emitted at all. Traceloop Anthropic emits `gen_ai.system.message` from `kwargs.get("system")` which can be absent. Both leave the trace without a system message even when the user passed one — silent data loss.

**Fix options:**

- *Normalize*: map any `gen_ai.<role>.message` that isn't one of the four known roles to `gen_ai.system.message` when the role is `developer`, or to the closest semantic match. Simple, preserves the explicit dispatch chain.
- *Dispatch by prefix*: catch any `gen_ai.<role>.message` in a single branch, use the role suffix as the `role` field, and let body-shape validators raise with precise field errors. This matches the semconv's treatment of role as the event-name suffix, not an enum.

**Test** — feed a `gen_ai.developer.message` event and assert the parsed trace contains a message with `role="developer"` (or `"system"` if you normalize), no raise.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+
+def _event_name_and_body(event: dict[str, Any]) -> tuple[str, dict[str, Any]]:
+    event_name = event.get("event_name", None)
+    if not isinstance(event_name, str):
+        raise ValueError(f"Event name is not a string: event={event}")
+    body = event.get("body", None)
+    if body is None:
+        raise ValueError(
+            f"Event body is missing, ensure to enable OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: {event}"
+        )
```
Hartorn: **High — Rejecting `body is None` is overly strict and the hint misleads users.**

`opentelemetry-instrumentation-openai-v2`'s `message_to_event` intentionally returns `LogRecord(body=None)` when an assistant/tool/user message has no content, no `tool_calls`, and no `id` — e.g. a bare `{"role": "assistant"}` that an SDK emits during role announcement. That's a legitimate LogRecord per the OTel Logs SDK contract; it is *not* evidence that content capture is off.

The current behaviour:
- `raise_on_unknown_event=True` (default) → the whole trace fails to parse.
- Hint points at `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`, which only toggles the inclusion of `content` and tool-call arguments — it doesn't turn metadata events on/off.

**Suggested fix.** Treat `body is None` as `{}` and let the role-specific Pydantic validators decide whether the event is usable. That keeps the parser tolerant of empty-but-valid events, and pushes any real error to a precise field-level `ValidationError` (e.g. "user message missing `content`"), which is much more actionable than the current blanket hint.

A small test feeding a `gen_ai.assistant.message` event with `body=None` would pin this down.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
Hartorn: **High — `_normalize_choice_body_for_semconv` has no unit test.**

This helper exists specifically to paper over Anthropic's `gen_ai.choice` body shape (nested `message` without a `role` field, or `tool_calls` hoisted to the top level). It's only exercised indirectly via the Anthropic functional test, which requires live API credentials — meaning in practice the branch is unguarded on PRs.

**Suggested unit tests** (parametrized; no live calls needed):

| input `choice_body` | expected behaviour |
|---|---|
| `{"message": {"role": "assistant", "content": "hi"}}` | unchanged (role already present) |
| `{"message": {"content": "hi"}}` | `role` injected as `"assistant"` |
| `{"tool_calls": [...]}` (top-level, Anthropic shape) | normalized into `message={"role": "assistant", "tool_calls": [...]}` |
| `{}` | unchanged (no-op) |
| `{"message": "stringly"}` | unchanged (guard against non-dict `message`) |

Each row flows through `GenAiTrace.from_otel_logs` with a single `gen_ai.choice` event plus a minimal user message, and asserts the resulting `ChoiceLike.message.role` / `tool_calls` shape. This locks the Anthropic parity behaviour to a unit test rather than relying on the functional gate.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
Hartorn: **High — `raise_on_unknown_event` is part of the public API of `from_otel_logs` and neither branch is tested.**

- `raise_on_unknown_event=True` (the default) → `raise ValueError(f"Unknown event name: {event_name}")`.
- `raise_on_unknown_event=False` → the event is silently dropped.

Both paths are reachable by any caller, and the default is *strict*. Right now a refactor could invert the condition or swap the two branches without a single test complaining.

**Suggested tests** (small, self-contained, no live calls):

```python
@pytest.mark.parametrize(
    "raise_on_unknown",
    [pytest.param(True, id="strict"), pytest.param(False, id="lenient")],
)
def test_from_otel_logs_unknown_event(raise_on_unknown):
    events = [
        {"event_name": "gen_ai.user.message", "body": {"content": "hi"}},
        {"event_name": "gen_ai.something.new", "body": {"foo": "bar"}},
        {"event_name": "gen_ai.choice", "body": {"message": {"role": "assistant", "content": "ok"}}},
    ]
    if raise_on_unknown:
        with pytest.raises(ValueError, match="Unknown event name"):
            GenAiTrace.from_otel_logs(events, raise_on_unknown_event=True)
    else:
        trace = GenAiTrace.from_otel_logs(events, raise_on_unknown_event=False)
        assert len(trace.interactions) == 1
        assert trace.interactions[0].inputs[0].content == "hi"
```

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+    if not isinstance(body, dict):
+        raise ValueError(f"Event body is not a dict: {body}")
+    return event_name, body
```
Hartorn: **High — the three `ValueError` branches in `_event_name_and_body` are uncovered.**

```python
if not isinstance(event_name, str):
    raise ValueError(f"Event name is not a string: event={event}")
body = event.get("body", None)
if body is None:
    raise ValueError(
        f"Event body is missing, ensure to enable OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: {event}"
    )
if not isinstance(body, dict):
    raise ValueError(f"Event body is not a dict: {body}")
```

Each of these is a real failure mode — a malformed event in a log export would hit one — and each has bespoke phrasing (the middle one, in particular, nudges users toward an env var; see the body-is-None comment above). None of them are pinned, so a future refactor can drop or reword the hint without any signal.

**Suggested test** (parametrized):

```python
@pytest.mark.parametrize(
    ("event", "match"),
    [
        pytest.param({"body": {}}, "Event name is not a string", id="missing-name"),
        pytest.param({"event_name": 123, "body": {}}, "Event name is not a string", id="non-string-name"),
        pytest.param({"event_name": "gen_ai.user.message"}, "Event body is missing", id="missing-body"),
        pytest.param({"event_name": "gen_ai.user.message", "body": "str"}, "Event body is not a dict", id="non-dict-body"),
    ],
)
def test_from_otel_logs_invalid_event(event, match):
    with pytest.raises(ValueError, match=match):
        GenAiTrace.from_otel_logs([event])
```

If the body-is-None branch is relaxed per the earlier comment, drop the `missing-body` row.

libs/giskard-checks/tests/functional/test_genai.py
```diff
@@ -0,0 +1,255 @@
...
+        "base_url": "os.environ/AZURE_AI_ENDPOINT",  # pragma: allowlist secret
+    },
+}
+
+_PROVIDER_MARKS = {
+    "openai": pytest.mark.openai,
+    "google": pytest.mark.google,
+    "anthropic": pytest.mark.anthropic,
+    "azure": pytest.mark.azure,
+    "azure_ai": pytest.mark.azure_ai,
```
Hartorn, line 76: **High — `giskard-checks` conftest is missing the provider-mark auto-skip pattern used by sibling packages.**

The current `pytest_collection_modifyitems` only handles the `integration` mark. Tests in `libs/giskard-checks/tests/functional/test_genai.py` carry per-provider marks (`openai`, `google`, `anthropic`, `azure`, `azure_ai`) via `_PROVIDER_MARKS`, but nothing maps those marks to SDK availability. Running `pytest -m anthropic` in a minimal env (e.g. only `giskard-llm[openai-otel]` installed, which is what CI cells on other providers look like) will attempt `_make_client("anthropic")` and crash on `ImportError: No module named 'anthropic'` rather than cleanly skipping.

Both sibling packages already solve this:

- `libs/giskard-llm/tests/conftest.py:28–45` maps `openai`/`bare`/`google`/`gemini`/`anthropic`/`azure`/`azure_ai` → the relevant SDK module and `pytest.mark.skip`s when the import fails.
- `libs/giskard-agents/tests/conftest.py:16–45` has the same shape.

**Suggested fix.** Port that pattern into `libs/giskard-checks/tests/conftest.py`: a `_PROVIDER_PACKAGES` dict plus a second pass in `pytest_collection_modifyitems` that adds `pytest.mark.skip` when the package isn't importable. With that in place, a cell running `openai-otel` won't try to collect and fail the `anthropic`-marked tests.

Optionally pair it with an autouse teardown fixture that force-uninstruments any residual OTel instrumentor between tests (see `libs/giskard-llm/tests/functional/conftest.py:6–17` for the `_clear_llm_provider_cache` analogue) to prevent cross-test global-state leakage if an earlier test errors before its `finally: uninstrument()`.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+                _flush_if_has_choices()
+                inputs.append(
+                    ToolMessageLike.model_validate(_body_with_role("tool", body))
+                )
+            elif event_name == "gen_ai.choice":
+                choice = ChoiceLike.model_validate(
+                    _normalize_choice_body_for_semconv(body)
+                )
+                choices.append(choice)
+            elif raise_on_unknown_event:
```
Hartorn: **Medium — `from_otel_logs` has no event-to-span correlation; concurrent requests silently merge.**

The parser splits `Interaction`s purely on emission order: `gen_ai.*.message` events accumulate into a "current inputs" list, and a `gen_ai.choice` event flushes them into one `Interaction`. That works in the functional tests because each test issues exactly one `acompletion` — one LLM request, one span, one well-ordered sequence of log records (`user.message` → `choice`).

**Where it breaks.** As soon as a caller does:

```python
await asyncio.gather(
    client.acompletion(model, [{"role": "user", "content": "A"}]),
    client.acompletion(model, [{"role": "user", "content": "B"}]),
)
```

the provider instrumentor emits events for the two requests into the same `LoggerProvider`, each tagged with its own `span_id`. The SDK interleaves them in the exporter roughly in wall-clock order, e.g.:

```
gen_ai.user.message   (span=A, body={"content": "A"})
gen_ai.user.message   (span=B, body={"content": "B"})
gen_ai.choice         (span=A, body={"message": {"role": "assistant", ...A}})
gen_ai.choice         (span=B, body={"message": {"role": "assistant", ...B}})
```

The parser sees that as:
1. accumulate `user(A)`, accumulate `user(B)`
2. hit `choice(A)` → flush → one `Interaction` with **both** user messages paired with A's choice
3. hit `choice(B)` → flush → one `Interaction` with **zero** inputs paired with B's choice

No exception; just wrong `trace.interactions`. This is a normal pattern for evaluation harnesses and anything batching prompts, so it's not a hypothetical edge case — any concurrent evaluation using `acompletion` will produce corrupted traces.

The test helper drops the only signal that would let us split correctly:

```123:123:libs/giskard-checks/tests/functional/test_genai.py
events.append({"event_name": lr.event_name, "body": lr.body})
```

`lr.trace_id` / `lr.span_id` are populated by the SDK; they just aren't forwarded.

**Suggested fix.**

1. Preserve `trace_id` / `span_id` on each event dict (one-line change in the helper).
2. In `from_otel_logs`, group events by `span_id` before running the flush logic, then concatenate interactions across groups. When `span_id` is absent the current flat behaviour stays as-is.
3. Add a unit test with two interleaved spans (hand-built event list, no SDK needed) asserting two `Interaction`s with the right inputs paired to the right choice.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
Hartorn: **Medium — `gen_ai.evaluation.result` will trip `raise_on_unknown_event=True`.**

The semantic conventions reserve `gen_ai.evaluation.result` for evaluator emissions (still in-development in the spec, but already emitted by some evaluation harnesses). If any evaluator instrumentation sits alongside the provider instrumentor — a very plausible setup for a `giskard-checks` user, since `giskard-checks` *is* an evaluation library — the exporter stream will contain `gen_ai.evaluation.result` records and the parser will crash with `ValueError: Unknown event name: gen_ai.evaluation.result` (default `raise_on_unknown_event=True`).

**Suggested fix.** Add an explicit passthrough for `gen_ai.evaluation.result`: either silently skip it (parser stays focused on completion traces) or capture it onto the `GenAiTrace` via the existing `annotations: dict[str, Any]` hook on `Trace` so downstream checks can access evaluator scores. A one-line skip is the zero-risk option; the annotation path is the forward-looking one.

Worth a small unit test either way: a stream with one `gen_ai.evaluation.result` event plus a normal user/choice pair, asserting the trace parses cleanly.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+
+    @model_validator(mode="after")
+    def _has_text_or_tool_calls(self) -> Self:
+        has_text = self.content is not None and self.content != ""
+        has_tools = self.tool_calls is not None and len(self.tool_calls) > 0
+        if not has_text and not has_tools:
+            raise ValueError(
+                "Assistant message must include non-empty content and/or at least one tool call"
+            )
+        return self
```
Hartorn: **Medium — assistant turns that carry both text *and* tool_calls lose the text under the OpenAI v2 legacy-attributes path.**

This isn't a parser bug per se, but it's an invisible-to-Pydantic data loss that `AssistantMessageLike` happily validates. In `opentelemetry-instrumentation-openai-v2::patch.py::message_to_event`, the legacy path does:

```python
if role == "assistant":
    tool_calls = extract_tool_calls(message, capture_content)
    if tool_calls:
        body = {"tool_calls": tool_calls}   # overwrites `content`
```

So when the model emits *both* a text reply and a tool call (a real case — "Let me check…" + a `search` tool call — supported by OpenAI, Anthropic, Gemini), the emitted `gen_ai.assistant.message` body has only the tool_calls. `AssistantMessageLike.content` is optional and the validator only requires *either* text or tool_calls (`gen_ai.py:74–82`), so parsing succeeds and silently returns an assistant turn with no text. Downstream checks that look at the assistant's free-form reasoning will miss it.

**Suggested handling.**

- Document this explicitly in the `from_otel_logs` docstring so callers understand the limitation is at the instrumentor layer, not in our parser.
- Optional: when we migrate off the legacy per-role events onto the single-event `gen_ai.client.inference.operation.details` schema (see the earlier comment on the legacy → latest-experimental transition), this ceases to be a problem because the new event shape carries the full `messages` array verbatim. That's another reason to prioritise that migration.

libs/giskard-checks/tests/functional/test_genai.py
```diff
@@ -0,0 +1,255 @@
...
+    GenAiTrace,
+    TextMessageLike,
+    ToolMessageLike,
+)
+from giskard.llm import ChatMessage, LLMClient, ToolDef
+
+pytestmark = pytest.mark.functional
+
+# openai-v2 instrumentation only treats the literal "true" as enabling capture for logger.emit()
+os.environ["OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"] = "true"
```
Hartorn: **Medium — module-level `os.environ[...] = "true"` leaks across the test session.**

Setting `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` at import time mutates the process environment and is never reset. Anything pytest imports after this module — other functional tests, `conftest.py` helpers, subprocesses spawned from tests, `pytest-xdist` workers — runs with capture enabled whether they want it or not. That's not a purely theoretical concern: a future test that expects the default (content capture off) will silently get full payloads, and a failure will be hard to reproduce outside the full test run because the env state depends on collection order.

**Suggested fix.** Move the env var into a session-scoped autouse fixture using `monkeypatch.setenv`, which restores the prior value at teardown:

```python
@pytest.fixture(autouse=True, scope="session")
def _enable_otel_content_capture(monkeypatch_session):
    monkeypatch_session.setenv("OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", "true")
```

(pytest's built-in `monkeypatch` is function-scoped; use the `pytest-monkeypatch-session` pattern or a `try/finally` with `os.environ.get(...)` + restore.)

This keeps the capture toggle local to the module that needs it.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+        yield from self.message.__rich_console__(console, options)
+
+
+def _start_index_after_last_assistant(messages: list[MessageLike]) -> int:
+    """Index to slice ``messages`` so only entries after the last ``assistant`` remain."""
+    last_assistant = -1
+    for i, m in enumerate(messages):
+        if m.role == "assistant":
+            last_assistant = i
+    return last_assistant + 1
```
Hartorn: **Medium — the `no-assistant` path of `_start_index_after_last_assistant` is untested.**

`test_from_otel_logs_drop_redundant_full_history` only exercises the case where an assistant message exists in the inputs. The other branch — `last_assistant` stays `-1`, helper returns `0`, the full inputs list is kept even with `drop_redundant_input_history=True` — is silently uncovered. A refactor that changes the sentinel (e.g. flipping to `last_assistant: int | None = None` and returning `len(messages)` by mistake) would pass the current suite while completely breaking the contract.

**Suggested test.** One parametrized row per branch:

```python
@pytest.mark.parametrize(
    ("inputs", "expected_kept"),
    [
        pytest.param(
            [user("hi"), user("hi again")],
            [user("hi"), user("hi again")],
            id="no-assistant-keeps-all",
        ),
        pytest.param(
            [user("hi"), assistant("hello"), user("again")],
            [user("again")],
            id="strips-up-to-last-assistant",
        ),
    ],
)
def test_from_otel_logs_drop_redundant_input_history(inputs, expected_kept): ...
```

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+    ) -> RenderResult:
+        yield from self.function.__rich_console__(console, options)
+
+
+class AssistantMessageLike(BaseModel, frozen=True, extra="allow"):
+    """Assistant turn: optional text ``content`` and/or optional ``tool_calls``."""
+
+    role: str
+    content: str | None = None
+    tool_calls: list[FunctionCallLike | ToolCallLike] | None = None
```
Hartorn: **Medium — the `list[FunctionCallLike | ToolCallLike]` discriminator is only exercised by live-API functional tests.**

`AssistantMessageLike.tool_calls` is annotated as `list[FunctionCallLike | ToolCallLike] | None`. Pydantic v2 resolves this in *smart mode* rather than via a proper discriminator (since `ToolCallLike.type: str` isn't a `Literal`). The behaviour we ship depends on:

- a well-formed `type="function_call"` dict being routed to `FunctionCallLike`,
- a raw `{"type": "web_search", ...}` dict being routed to `ToolCallLike`,
- a malformed `{"type": "function_call", "id": "x"}` (missing `function`) quietly falling back to `ToolCallLike` because `extra="allow"`.

None of that is pinned at the unit-test level. The only way the code path is exercised today is via the Anthropic functional test, which requires live credentials and won't run in a default-install environment.

**Suggested tests** (parametrized, unit-level, no SDK):

```python
@pytest.mark.parametrize(
    ("tool_call_body", "expected_cls"),
    [
        pytest.param(
            {"type": "function_call", "id": "c1", "function": {"name": "search", "arguments": "{}"}},
            FunctionCallLike,
            id="function-call-valid",
        ),
        pytest.param(
            {"type": "web_search", "id": "c2"},
            ToolCallLike,
            id="generic-tool-call",
        ),
        pytest.param(
            {"type": "function_call", "id": "c3"},  # missing `function`
            ToolCallLike,  # smart-mode fallback, per current behaviour
            id="function-call-malformed-falls-back",
        ),
    ],
)
def test_from_otel_logs_assistant_tool_call_dispatch(tool_call_body, expected_cls): ...
```

If the `FunctionCallLike` / `ToolCallLike` union later uses a proper discriminator, the third row becomes a `pytest.raises(ValidationError)` case.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
Hartorn: **Medium — the `_has_text_or_tool_calls` validator has no direct test.**

`AssistantMessageLike` enforces "must have non-empty `content` or at least one `tool_calls` entry" as a post-init model validator. That's a real invariant — a `gen_ai.assistant.message` body with `{"role": "assistant"}` and nothing else should be rejected — but no unit test exercises it. A future refactor could silently weaken the rule (e.g. by dropping the empty-string check) and every existing test would still pass.

**Suggested test** (parametrized, small):

```python
@pytest.mark.parametrize(
    "body",
    [
        pytest.param({"role": "assistant"},                                  id="empty"),
        pytest.param({"role": "assistant", "content": ""},                    id="empty-string-content"),
        pytest.param({"role": "assistant", "content": None, "tool_calls": []}, id="empty-tool-calls"),
        pytest.param({"role": "assistant", "tool_calls": None},               id="null-tool-calls-only"),
    ],
)
def test_assistant_message_requires_content_or_tool_calls(body):
    with pytest.raises(ValidationError, match="non-empty content"):
        AssistantMessageLike.model_validate(body)
```

Pair with a passing case per shape (content-only, tool-calls-only, both) if you want positive coverage too.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
Hartorn: **Medium — Pydantic smart-mode union on `list[FunctionCallLike | ToolCallLike]` silently degrades on malformed tool calls.**

`ToolCallLike.type: str` is a catch-all, while `FunctionCallLike.type: Literal["function_call"]` is a fixed tag. Because one arm of the union is not a `Literal`, Pydantic v2 falls back to *smart mode* rather than tag-dispatch. For a well-formed `{"type": "function_call", "id": "x", "function": {...}}` that's fine — the `Literal` match wins over `str`. But for a **malformed** `{"type": "function_call", "id": "x"}` (missing the required `function`), `FunctionCallLike` validation fails and smart mode silently falls back to `ToolCallLike`. Because `ToolCallLike` has `extra="allow"`, the missing `function` field just lands in the allow-extras bag and the caller gets a seemingly-valid `ToolCallLike` instead of a `ValidationError`. Any assertion downstream that expects `function_call` payloads to be parseable will either crash late or silently lose fields.

`Field(discriminator="type")` can't be applied directly here because a Pydantic discriminator requires every member to declare a `Literal` for the discriminator field, and `ToolCallLike.type: str` is the catch-all.

**Two realistic fixes:**

1. Give `ToolCallLike` a distinct set of literals (e.g. `Literal["web_search", "computer_use", "code_interpreter", ...]` — union of the non-function tool types we care about) and wire an explicit discriminator. Strictest, but needs enumerating the tool types.
2. Keep smart mode but add a `model_validator` on `AssistantMessageLike` that rejects any raw dict in `tool_calls` whose `type == "function_call"` but whose `function` is missing. Localised, no enumeration needed.

Pair with a unit test feeding a malformed `{"type": "function_call", "id": "x"}` and asserting `ValidationError`.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+            for tool_call in self.tool_calls:
+                yield Panel(
+                    tool_call,
+                    title=f"{tool_call.type.capitalize()} call: {tool_call.id}",
+                    border_style=_DEFAULT_BORDER_STYLE,
+                )
+
+
+# Backward-compatible name for assistant messages that include tool calls.
+ToolCallMessageLike = AssistantMessageLike
```
Hartorn: **Medium — the `ToolCallMessageLike = AssistantMessageLike` alias is unneeded; drop it.**

Git history shows `ToolCallMessageLike` was renamed to `AssistantMessageLike` *within this PR* (the rename commit is the one that introduced `gen_ai.py`; the file itself is new on this branch). `giskard-checks` is at `1.0.2b1` (beta — `libs/giskard-checks/pyproject.toml:3`), and a full repo `rg ToolCallMessageLike` returns a single hit, this line. There is no external caller to preserve compatibility for — the symbol never shipped.

Keeping a back-compat alias for a name that was only ever live within the same PR just adds import noise and invites future "why are there two names for this?" questions.

**Suggested change.** Remove lines 103–104 outright. If anyone pulled a pre-merge branch and imported it, a one-line rename on their side is trivial and they'll discover it immediately via `ImportError`.

libs/giskard-checks/src/giskard/checks/core/interaction/gen_ai.py
```diff
@@ -0,0 +1,304 @@
...
+    inputs_redundant_prefix_stripped: bool = False
+
+    @classmethod
+    def from_otel_logs(
+        cls,
+        events: list[dict[str, Any]],
+        /,
+        *,
+        raise_on_unknown_event: bool = True,
+        drop_redundant_input_history: bool = False,
```
Hartorn: **Medium — the field name and the kwarg that sets it use different verbs and different nouns for the same thing.**

- Field (line 182): `inputs_redundant_prefix_stripped: bool = False` — verb *strip*, noun *prefix*.
- Kwarg (line 191): `drop_redundant_input_history: bool = False` — verb *drop*, noun *history*.

Same concept, two vocabularies. Readers have to pattern-match to realise the field is just the reified form of the kwarg, and anyone using the public API has to keep both names in their head.

**Suggested fix.** Pick one pair and apply it to both places. Two equally good options:

- Keep the field name, align the kwarg: `inputs_redundant_prefix_stripped: bool` as the field, `strip_redundant_input_prefix=False` as the kwarg.
- Keep the kwarg, align the field: `drop_redundant_input_history=False` as the kwarg, `redundant_input_history_dropped: bool` as the field.

Either is fine — the important property is that the field name is the past-participle form of the kwarg so the mapping is obvious.

libs/giskard-llm/pyproject.toml
```diff
@@ -16,6 +16,32 @@ google = ["google-genai>=1.0,<2"]
...
+    "opentelemetry-sdk>=1.41,<2",
+    "wrapt<2",
+]
+anthropic-otel = [
+    "giskard-llm[anthropic]",
+    "httpx>=0.28.1",
+    "opentelemetry-instrumentation-anthropic>=0.59b0",
+    "opentelemetry-sdk>=1.41,<2",
+    "wrapt<2",
+]
```
Hartorn, line 42: **Medium — `httpx>=0.28.1` is repeated in every base `-otel` extra; the ownership is unclear.**

`libs/giskard-llm/pyproject.toml` now has the pin in three places (`openai-otel`, `google-otel`, `anthropic-otel`). A quick `rg '^import httpx' libs/giskard-llm/src libs/giskard-checks/src` returns zero hits — nothing we own imports `httpx` directly. The provider SDKs (`openai`, `anthropic`, `google-genai`) already declare and constrain it, and none of `opentelemetry-sdk` / `opentelemetry-instrumentation-openai-v2` / `opentelemetry-instrumentation-google-genai` / `opentelemetry-instrumentation-anthropic` pulls `httpx` through their dep graphs.

So the pin has two problems:

1. **It's not OTel-related.** The repetition inside `*-otel` extras implies it is, but the underlying motivation — based on earlier commit messages — was an observed-in-practice fix, not an OTel requirement.
2. **Duplication invites drift.** If someone later bumps two of the three to `>=0.29` and forgets the third, users on that extra silently stay on the older floor, and the discrepancy is invisible without `pyproject.toml` diffing.

**Suggested fix.** Either:

- Drop the `httpx` lines outright — provider SDKs already constrain it — and rely on their floors.
- Or hoist it to `[project] dependencies` with a comment explaining why `0.28.1` specifically is required.

Either way, removes three duplicate lines and centralises ownership.

libs/giskard-checks/tests/core/test_gen_ai.py
```diff
@@ -0,0 +1,184 @@
...
+        {
+            "event_name": "gen_ai.user.message",
+            "body": {"content": "Second — no completion logged"},
+        },
+    ]
+    with caplog.at_level(logging.WARNING):
+        trace = GenAiTrace.from_otel_logs(events)
+    assert len(trace.interactions) == 1
+    assert trace.interactions[0].inputs[0].content == "First"
+    assert any("gen_ai.choice" in r.message for r in caplog.records)
```
Hartorn, line 159: **Low — `caplog.records[i].message` returns the format string, not the rendered message.**

```168:168:libs/giskard-checks/tests/core/test_gen_ai.py
assert any("gen_ai.choice" in r.message for r in caplog.records)
```

`LogRecord.message` is set by `LogRecord.getMessage()` *only* if the formatter has already been invoked on that record. For records captured by `caplog` (which uses its own handler), `.message` is the raw `msg` template — in our case `"Ignoring %d input message(s) without a final gen_ai.choice (incomplete log stream or interrupted request)"`. The substring `"gen_ai.choice"` is in the template itself, so the assertion passes *by accident* — it would pass even if the runtime args were completely wrong.

**Suggested fix.** Use `r.getMessage()`, and tighten the assertion to actually validate intent:

```python
messages = [r.getMessage() for r in caplog.records]
assert any("gen_ai.choice" in m and "1 input message(s)" in m for m in messages)
```

That way a regression that changes "1" to "0" or drops the count formatting will fail the test loudly. Same fix applies to the analogous assertion in `test_from_otel_logs_assistant_without_choice_ignored` (line 184).

libs/giskard-llm/src/giskard/llm/routing.py
```diff
@@ -103,6 +107,15 @@ def configure_from_dict(self, config: dict[str, dict[str, Any]]) -> None:
...
             self.configure(name, **kwargs)
 
+    @staticmethod
+    def instrumentor_for(provider: str) -> "BaseInstrumentor":
+        """Load the OpenTelemetry GenAI instrumentor for a provider kind.
+
+        Does not use client configuration; ``provider`` is the logical SDK kind
+        (same strings as in model prefixes: ``openai``, ``google``, etc.).
+        """
+        return instrumentor_for_provider(provider)
```
Hartorn, line 117: **Low — two public entry points for the same stateless helper; pick one.**

- `instrumentor_for_provider(provider)` is a free function in `libs/giskard-llm/src/giskard/llm/otel.py:50`, re-exported from `libs/giskard-llm/src/giskard/llm/__init__.py`.
- `LLMClient.instrumentor_for(provider)` is a `@staticmethod` here (`routing.py:110–117`) that takes no client state and just delegates to the same free function.

The docstring itself tells the story: *"Does not use client configuration"*. It doesn't belong on the client. Both paths are exercised (unit + functional), so there's no coverage argument for keeping both — they're the same code path.

Shipping two names for the same operation makes the public API look bigger than it is and forces future readers to figure out which one is "canonical".

**Suggested fix.** Pick one:

- **Keep the free function** (my preference — it's already in the top-level re-export and matches the lazy-import pattern for OTel bits). Remove `LLMClient.instrumentor_for`.
- Or keep the static method as the canonical entry point and drop the top-level re-export.

Either way resolves the duplication; the free function feels more natural because there's nothing client-specific about the operation.

libs/giskard-checks/tests/core/test_gen_ai.py
```diff
@@ -0,0 +1,184 @@
...
+
+import pytest
+from giskard.checks.core.interaction.gen_ai import (
+    AssistantMessageLike,
+    GenAiTrace,
+    TextMessageLike,
+)
+
+
+def test_from_otel_logs_multiple_choices_same_request():
```
Hartorn: **Low — several `test_gen_ai.py` tests are natural candidates for parametrization.**

Five independent test functions in this file each hand-build a separate events list. Per the repo test-style guidance, parametrized functions with `ids=[...]` are preferred when the scenarios differ only in input shape. Natural consolidations:

- **Event-to-message-type dispatch** — one parametrized test over `[("gen_ai.system.message", TextMessageLike, "system"), ("gen_ai.user.message", TextMessageLike, "user"), ("gen_ai.assistant.message", AssistantMessageLike, "assistant"), ("gen_ai.tool.message", ToolMessageLike, "tool")]` replacing separate per-role tests.
- **Invalid-event validation** — one parametrized test over the three `_event_name_and_body` error shapes (covered by the earlier H5 comment).
- **Trailing-inputs warning** — `test_from_otel_logs_trailing_inputs_without_choice_ignored` and `test_from_otel_logs_assistant_without_choice_ignored` have the same shape, different input — one parametrized function with two ids covers both.

None of these change coverage; they just compress the file and make adding a new role/shape row a one-line addition instead of a new function. Worth doing alongside the new tests proposed in earlier comments so the file stays consistent.

Hartorn (changes requested):
This review collects findings from a multi-reviewer pass (API semantics, test coverage, dependency hygiene, architecture). I've broken it into inline comments — each anchored at the relevant lines — so they can be triaged individually. A quick map of what's in the review:

**Critical (5).** `FunctionCallLike.type = Literal["function_call"]` mismatches real instrumentor emissions; Anthropic `gen_ai.choice` body shape not normalised; Google GenAI default-mode body shape entirely incompatible; `TextMessageLike.content: str` rejects Anthropic's list-of-blocks content; Traceloop Anthropic emits one `gen_ai.choice` per content block, not per candidate.

**High (8).** `ToolMessageLike.id` required but spec/instrumentor allow null; parser targets the pre-1.37 legacy event model while semconv is migrating to a single `gen_ai.client.inference.operation.details` event; `gen_ai.system.message` not always emitted and `role="developer"` crashes dispatch; `body=None` rejection is overly strict and the hint misleads; `_normalize_choice_body_for_semconv` has no unit test; `raise_on_unknown_event` has no test in either mode; `_event_name_and_body` validation errors uncovered; `giskard-checks` conftest missing the provider-mark auto-skip pattern used by the sibling packages.

**Medium (11).** No event-to-span correlation (concurrent `asyncio.gather` requests silently merge); `gen_ai.evaluation.result` not handled (default strict mode will crash); assistant-with-text-and-tool-calls loses text under OpenAI v2 legacy path; module-level `os.environ[...]` in `test_genai.py`; no-assistant branch of `_start_index_after_last_assistant` untested; `AssistantMessageLike` validator untested; tool-call polymorphism untested; smart-mode union on tool calls silently degrades; dead `ToolCallMessageLike` alias; `inputs_redundant_prefix_stripped` vs `drop_redundant_input_history` vocabulary mismatch; `httpx` duplicated across three `-otel` extras.

**Low (3).** `caplog.record.message` vs `getMessage()`; two public entry points for the same stateless helper (`instrumentor_for_provider` + `LLMClient.instrumentor_for`); parametrization opportunities in `test_gen_ai.py`.

**Addressed in a follow-up commit on the branch:** unit tests for the happy paths and missing-extra `ImportError` of `instrumentor_for_provider` (5 provider rows × 2 tests — `libs/giskard-llm/tests/test_routing.py`).

**Running CI against this branch directly.** The `Integration Tests` workflow supports `workflow_dispatch` in addition to `pull_request_target`. To test the actual branch state (without the merge into the target branch that `pull_request_target` applies):

```bash
gh workflow run integration-tests.yml --ref feat/giskard-llm-otel-support
gh run watch
```

Or via the Actions tab → *Integration Tests* → *Run workflow* → pick `feat/giskard-llm-otel-support`. This is useful for verifying matrix changes and branch-level fixes without waiting for the PR merge preview.

Issue #2410, opened 2026-04-17
Adopt `TypeVar(default=...)` for generic base classes so callers can omit bounded type parameters


kevinmessiaen:
## Problem

We use PEP 695 generics (`class Foo[InputType, OutputType, TraceType: Trace]:`) widely across `giskard-checks`. These declarations carry bounded parameters (e.g. `TraceType: Trace`) but do not offer defaults, because **PEP 695 native syntax does not support defaults for type parameters**. As a result:

- Callers are forced to specify every type argument, or we silence the missing-arg diagnostic with `# pyright: ignore[reportMissingTypeArgument]`.
- Almost every core/base and judge class currently carries that ignore comment — it has become noise and hides genuine missing-arg issues.
- Users doing `Check(...)` / `Scenario(...)` / `TestCase(...)` without explicit parameters get either degraded inference or an invalid-generic diagnostic.

### Where we've already solved this

in #2406 `Trace` in `libs/giskard-checks/src/giskard/checks/core/interaction/trace.py` switched from PEP 695 syntax to `typing_extensions.TypeVar(..., default=...)` to allow `Trace[str, str]` to transparently expand to `Trace[str, str, Interaction[str, str]]`:

```11:17:libs/giskard-checks/src/giskard/checks/core/interaction/trace.py
InputType = TypeVar("InputType")
OutputType = TypeVar("OutputType")
InteractionType = TypeVar(
    "InteractionType",
    bound=Interaction[Any, Any],
    default=Interaction[InputType, OutputType],
)
```

This pattern (PEP 696, `typing_extensions.TypeVar`) gives us the missing ergonomic piece for the rest of the codebase.

## Goal

Apply the same `TypeVar(..., default=...)` pattern to every generic class/function where a sensible default exists, and drop the `# pyright: ignore[reportMissingTypeArgument]` suppressions.

### Primary candidates

All carry `TraceType: Trace` today without a default — default should be `Trace[InputType, OutputType]`:

- `Check[InputType, OutputType, TraceType]` — `libs/giskard-checks/src/giskard/checks/core/check.py`
- `TestCase[InputType, OutputType, TraceType]` — `libs/giskard-checks/src/giskard/checks/core/testcase.py`
- `Step`, `Scenario` — `libs/giskard-checks/src/giskard/checks/core/scenario.py`
- `ScenarioResult[TraceType]` — `libs/giskard-checks/src/giskard/checks/core/result.py`
[...]

PR #2411, opened 2026-04-17
Adopt TypeVar defaults for giskard-checks generics

harsh21234i:
Hey @kevinmessiaen  #2410 
## What changed

  This updates `giskard-checks` generic base classes and related helpers to
  use module-local `typing_extensions.TypeVar(..., default=...)` defaults
  instead of relying on PEP 695 bounded parameters plus `# pyright:
  ignore[reportMissingTypeArgument]`.

  The change was applied across:
  - core check/scenario/testcase/result/input generator abstractions
  - interaction specs and helpers

- builtin checks such as composition, comparison, text matching, semantic
  similarity, and function-backed checks
  - LLM judges and user simulator helpers
  - test runner and spy utilities

  ## Why

  Issue #2410 identified that bounded generic parameters like `TraceType:
  Trace` forced callers to spell out all type arguments or rely on pyright
  suppressions. That created noise in the codebase and degraded ergonomics
  for common usages like:

  - `Check()`
  - `Scenario()`
  - `TestCase()`
  - `Interact()`
  - builtin combinators and judges without explicit generic arguments

  Using `TypeVar(default=...)` restores the intended typing behavior while
  keeping runtime behavior unchanged.

  ## Root cause

  PEP 695 native generic syntax supports bounds but does not support default
  type parameters. Because of that, classes using bounded `TraceType`
  parameters could not express the natural default of `Trace[InputType,
  OutputType]`, and the codebase had accumulated `reportMissingTypeArgument`
  suppressions as a workaround.

  ## Impact

  - callers can omit bounded trace type parameters in the common case
  - pyright suppressions for missing type arguments are no longer needed in
  the updated `giskard-checks` surface
  - public APIs remain the same from a runtime perspective
  - this is a typing-only ergonomics improvement; serialization and runtime
  behavior are unchanged

  ## Validation

  - `uv tool run basedpyright --level error libs/giskard-checks/src libs/
  giskard-checks/tests`
  - `uv run pytest libs/giskard-checks -m "not functional"`
  - `uv tool run ruff check libs/giskard-checks/src libs/giskard-checks/
  tests`
  - `uv tool run ruff format --check libs/giskard-checks/src libs/giskard-
  checks/tests`

  Results:
  - basedpyright: passed
  - pytest: `469 passed, 4 skipped`
  - ruff check: passed
  - ruff format check: passed

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
@@ -1,8 +1,9 @@
 from abc import ABC, abstractmethod
-from typing import Any, Self, override
+from typing import Any, Generic, Self, override
 
 from giskard.core import NOT_PROVIDED, NotProvided, provide_not_none
 from pydantic import Field, model_validator
+from typing_extensions import TypeVar
 
 from ..core import Trace
 from ..core.check import Check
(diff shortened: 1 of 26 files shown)


kevinmessiaen (changes requested):
Hey, thanks for your contribution. Let's move the `TypeVar` definition to a shared module so we don't have to redefine it. Let's also update usages (mainly in tests) to show how it improve dev-ex

harsh21234i:
I moved the shared InputType / OutputType / TraceType definitions into a shared core.typevars module and updated the
  affected giskard-checks components to import them instead of redefining them locally.

  I also updated test usages to make the typing ergonomics clearer, mainly in the scenario and testcase tests, so the
  improved generic usage is visible in practice.

  Validation:

  - pytest libs/giskard-checks/tests/core/test_scenario.py -q
  - pytest libs/giskard-checks/tests/scenarios/test_testcase.py -q

  Both are passing on the updated branch.

harsh21234i:
Hi @kevinmessiaen, I noticed PR #2411 was closed and
  I wanted to check whether that was intentional
  because the proposed direction is no longer desired,
  or if you’d prefer the changes to be reopened/
  reworked in a new PR.

PR #2442, opened 2026-05-10
fix(agents): surface unknown tool calls

### LINKED ISSUE #2413: Giskard-agents: Workflow silently ignores unknown tool   calls
### Issue Type

Bug

### Source

source

### Giskard Library Version

 1.0.2b1

### OS Platform and Distribution

_No response_

### Python version

_No response_

### Installed python packages

```shell

```

### Current Behaviour?

```shell
A bug happened! giskard-agents workflow execution silently ignores unknown tool calls. In
  libs/giskard-agents/src/giskard/agents/workflow.py,
  _StepRunner._run_tools() does:
  if tool_call.function.name not in self._workflow.tools: continue  # TODO:
  raise an error?
  This hides misconfigurations and makes failures hard to debug because the
  tool call is never handled and no error is surfaced.
```

```shell
1. Configure a ChatWorkflow with a generator that returns an assistant
     message containing a tool call for a tool name that is not registered
     in ChatWorkflow.tools.
  2. Run the workflow.
  3. Observe the tool call is skipped without an error (no exception, no
     tool error message), leading to confusing behavior.
 Expected behaviour
  Unknown tool calls should be surfaced clearly, e.g. raise a WorkflowError
  (or return a structured tool error message), ideally behind a configurable
  policy (raise/return/skip) with logging.
```

### Relevant log output

```shell

```

extrasmall0:
## Summary
- raise a WorkflowError when a generator requests an unregistered tool
- preserve explicit WorkflowError messages instead of rewrapping them as a generic step failure
- add coverage for the unknown-tool path

## Testing
- uv run pytest libs/giskard-agents/tests/test_workflow_error_handling.py -q

Closes #2413

### DIFF
--- a/libs/giskard-agents/src/giskard/agents/workflow.py
+++ b/libs/giskard-agents/src/giskard/agents/workflow.py
@@ -174,7 +174,9 @@ class _StepRunner:
 
         for tool_call in chat.last.tool_calls:
             if tool_call.function.name not in self._workflow.tools:
-                continue  # TODO: raise an error?
+                raise WorkflowError(
+                    f"Unknown tool call requested by generator: {tool_call.function.name}"
+                )
 
             tool = self._workflow.tools[tool_call.function.name]
             tool_content = await tool.run(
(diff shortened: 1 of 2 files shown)


extrasmall0:
Quick maintainer note: the patch is still small and ready on my side. The only red status I can see is the fork authorization gate (the `authorize` job), so if a maintainer is comfortable authorizing or re-running that workflow, I can refresh anything else right away.

davidberenstein1957:
already covered here https://github.com/Giskard-AI/giskard-oss/pull/2414/

PR #2450, opened 2026-05-14
checks: support pydantic-compatible input types

### LINKED ISSUE #2449: feat(giskard-checks): extend input_type support to all pydantic-compatible types
## Problem

`InputGenerator.__call__`, `BaseLLMGenerator`, and `_infer_input_type` all restrict `input_type` to `str | BaseModel`. This means users who annotate their target callable with a pydantic-compatible but non-`BaseModel` type get no structured output:

```python
# Today — only works for str or BaseModel subclasses
def my_target(inputs: list[str]) -> str: ...      # ❌ inferred as None → falls back to str
def my_target(inputs: int) -> str: ...             # ❌ same
def my_target(inputs: dict[str, Any]) -> str: ... # ❌ same
def my_target(inputs: MyModel) -> str: ...         # ✓ works
```

The limitation is spread across three layers:
- `_infer_input_type` (`libs/giskard-checks/src/giskard/checks/core/interaction/interact.py:17`) — only returns a type if it is a `BaseModel` subclass
- `InputGenerator.__call__` overloads (`libs/giskard-checks/src/giskard/checks/core/input_generator.py:14`) — typed as `type[str] | None` or `type[T: BaseModel]`
- `LLMGeneratorOutput[T: str | BaseModel]` (`libs/giskard-checks/src/giskard/checks/generators/base.py:16`) — generic bound restricts `T`

## Desired Behavior

Any type that pydantic can validate should be accepted as `input_type`:

```python
def my_target(inputs: list[str]) -> str: ...             # ✓ inferred, generator yields list[str]
def my_target(inputs: int) -> str: ...                   # ✓ inferred, generator yields int
def my_target(inputs: tuple[str, int]) -> str: ...       # ✓
def my_target(inputs: MyModel | None) -> str: ...        # ✓
def my_target(inputs: list[MyModel]) -> str: ...         # ✓
```

Pydantic's [`TypeAdapter`](https://docs.pydantic.dev/latest/concepts/type_adapter/) is the mechanism that enables validation and JSON schema generation for arbitrary types without requiring `BaseModel`.

## Constraints

- **Do not break the existing `BaseModel` path.** `BaseLLMGenerator` uses `with_output(LLMGeneratorOutput[T])`, which calls `model_validate_json` and `model_json_schema()` — both `BaseModel` methods. Non-`BaseModel` types must go through a `TypeAdapter`-based wrapper that exposes the same interface. `RootModel[T]` is a candidate.
- **`str` remains the default** when `input_type` is `None` or cannot be inferred.
- **The public API surface of `InputGenerator.__call__` must remain backward-compatible** — existing code passing a `BaseModel` subclass must continue to work unchanged.

## Acceptance Criteria

- [ ] `_infer_input_type` returns the annotated type for any pydantic-compatible type (not only `BaseModel` subclasses); returns `None` for `str` and unannotated callables (existing behavior preserved)
- [ ] `InputGenerator.__call__` accepts `type[Any]` as `input_type` (or a union of supported types), not just `type[str] | type[T: BaseModel]`
- [ ] `BaseLLMGenerator.__call__` correctly constructs a pydantic-parseable schema for non-`BaseModel` types (e.g. via `RootModel` wrapping or `TypeAdapter`) and unwraps the result before yielding
- [ ] `LLMGeneratorOutput[T]` generic bound is relaxed to accept non-`BaseModel` types
- [ ] Passing `list[str]` as `input_type` produces a `list[str]` value (not a `RootModel` wrapper) from the generator
- [ ] Passing `int` as `input_type` produces an `int` value
- [ ] Passing `MyModel` (a `BaseModel` subclass) as `input_type` continues to produce a `MyModel` instance (no regression)
- [ ] Passing `str` (or `None`) as `input_type` continues to produce a `str` value (no regression)
- [ ] Unit tests added for `_infer_input_type` with `list[str]`, `int`, `dict[str, Any]`, `MyModel | None`
- [ ] Integration test: `BaseLLMGenerator` yields a `list[str]` when `input_type=list[str]`

## Files to Touch

| File | Change |
|------|--------|
| `libs/giskard-checks/src/giskard/checks/core/interaction/interact.py` | Broaden `_infer_input_type` to return any pydantic-compatible type |
| `libs/giskard-checks/src/giskard/checks/core/input_generator.py` | Relax overload signatures |
| `libs/giskard-checks/src/giskard/checks/generators/base.py` | Relax `LLMGeneratorOutput` bound; add `RootModel`/`TypeAdapter` wrapping in `BaseLLMGenerator.__call__` |
| `libs/giskard-checks/tests/core/test_input_type_inference.py` | New test cases |
| `libs/giskard-checks/tests/generators/test_llm_generator.py` | Integration test for non-`BaseModel` types |

## Out of Scope

- Supporting types that pydantic cannot validate (e.g. arbitrary dataclasses without pydantic config)
- Changing the `with_output` API in `giskard-agents`
- Auto-detecting supported types at import time

harsh21234i:
Closes #2449

  ## Summary
  - Broaden `_infer_input_type` to return Pydantic-compatible annotations beyond `BaseModel`
  - Relax `InputGenerator.__call__` and `LLMGeneratorOutput` typing for non-BaseModel inputs
  - Validate generator input types with Pydantic `TypeAdapter`
  - Add tests for `list[str]`, `int`, `dict[str, Any]`, and `MyModel | None`

  ## Testing
  - `uv run -m pytest -q libs/giskard-checks/tests/core/test_input_type_inference.py libs/giskard-checks/tests/core/
  test_input_generator.py libs/giskard-checks/tests/generators/test_llm_generator.py`
  - `uv run ruff check ...`
  - `uv run basedpyright ...`

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/core/input_generator.py
+++ b/libs/giskard-checks/src/giskard/checks/core/input_generator.py
@@ -1,5 +1,5 @@
 from collections.abc import AsyncGenerator
-from typing import TYPE_CHECKING, overload
+from typing import TYPE_CHECKING, Any, overload
 
 from giskard.core import Discriminated, discriminated_base
 from pydantic import BaseModel
@@ -18,7 +18,11 @@ class InputGenerator[TraceType: "Trace"](Discriminated):  # pyright: ignore[repo
     def __call__[T: BaseModel](
         self, trace: TraceType, input_type: type[T]
     ) -> AsyncGenerator[T, TraceType]: ...
+    @overload
+    def __call__[T](
+        self, trace: TraceType, input_type: T
+    ) -> AsyncGenerator[Any, TraceType]: ...
     def __call__(
-        self, trace: TraceType, input_type: type[str | BaseModel] | None = None
-    ) -> AsyncGenerator[str | BaseModel, TraceType]:
+        self, trace: TraceType, input_type: Any | None = None
+    ) -> AsyncGenerator[Any, TraceType]:
         raise NotImplementedError
--- a/libs/giskard-checks/src/giskard/checks/core/input_generator.py
+++ b/libs/giskard-checks/src/giskard/checks/core/input_generator.py
@@ -20,9 +20,9 @@ class InputGenerator[TraceType: "Trace"](Discriminated):  # pyright: ignore[repo
     ) -> AsyncGenerator[T, TraceType]: ...
     @overload
     def __call__[T](
-        self, trace: TraceType, input_type: T
-    ) -> AsyncGenerator[Any, TraceType]: ...
+        self, trace: TraceType, input_type: type[T]
+    ) -> AsyncGenerator[T, TraceType]: ...
     def __call__(
-        self, trace: TraceType, input_type: Any | None = None
+        self, trace: TraceType, input_type: type[Any] | None = None
     ) -> AsyncGenerator[Any, TraceType]:
         raise NotImplementedError
--- a/libs/giskard-checks/src/giskard/checks/core/interaction/interact.py
+++ b/libs/giskard-checks/src/giskard/checks/core/interaction/interact.py
@@ -13,12 +13,13 @@ from .interaction import Interaction
 from .trace import Trace
 
 
-def _infer_input_type(outputs: object) -> Any | None:
+def _infer_input_type(outputs: object) -> type | None:
     """Infer the input type from the first parameter annotation of a callable.
 
-    Returns any pydantic-compatible type except ``str``, which is the default
-    generator type. Returns ``None`` for non-callables and callables whose hints
-    cannot be resolved (e.g. forward references to undefined names).
+    Returns any pydantic-compatible type, including ``str``. Returns ``None``
+    for non-callables, callables with no annotation, and callables whose hints
+    cannot be resolved (e.g. forward references to undefined names) or whose
+    type is not supported by Pydantic.
     """
     if not callable(outputs):
         return None
--- a/libs/giskard-checks/src/giskard/checks/core/interaction/interact.py
+++ b/libs/giskard-checks/src/giskard/checks/core/interaction/interact.py
@@ -4,8 +4,7 @@ from typing import Any, cast, get_type_hints, override
 
 from giskard.checks.utils.injectable import ValueGenerator, ValueProvider
 from giskard.core.utils import NOT_PROVIDED, NotProvided
-from pydantic import BaseModel as PydanticBaseModel
-from pydantic import Field, PrivateAttr, model_validator
+from pydantic import Field, PrivateAttr, TypeAdapter, model_validator
 
 from ..input_generator import InputGenerator
 from ..types import GeneratorType, ProviderType
@@ -14,11 +13,11 @@ from .interaction import Interaction
 from .trace import Trace
 
 
-def _infer_input_type(outputs: object) -> type | None:
+def _infer_input_type(outputs: object) -> Any | None:
     """Infer the input type from the first parameter annotation of a callable.
 
-    Returns the type if it is a subclass of ``pydantic.BaseModel``, otherwise
-    ``None``.  Returns ``None`` for non-callables and callables whose hints
+    Returns any pydantic-compatible type except ``str``, which is the default
+    generator type. Returns ``None`` for non-callables and callables whose hints
     cannot be resolved (e.g. forward references to undefined names).
     """
     if not callable(outputs):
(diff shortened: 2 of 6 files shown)


libs/giskard-checks/src/giskard/checks/core/interaction/interact.py
```diff
@@ -48,11 +47,13 @@ def _infer_input_type(outputs: object) -> type | None:
     if not param_hints:
         return None
     first_param_type = next(iter(param_hints.values()))
-    if isinstance(first_param_type, type) and issubclass(
-        first_param_type, PydanticBaseModel
-    ):
-        return first_param_type
-    return None
+    if first_param_type is str:
+        return None
```
kevinmessiaen: We can return `str`, even if `None` is the fallback it allow for call site to differentiate between untyped and `str` type (if we want to add warning for example)

```suggestion
```

libs/giskard-checks/src/giskard/checks/core/input_generator.py
```diff
@@ -18,7 +18,11 @@ def __call__(
     def __call__[T: BaseModel](
         self, trace: TraceType, input_type: type[T]
     ) -> AsyncGenerator[T, TraceType]: ...
+    @overload
+    def __call__[T](
+        self, trace: TraceType, input_type: T
+    ) -> AsyncGenerator[Any, TraceType]: ...
```
kevinmessiaen: We can improve the type hint as following:

```suggestion
    @overload
    def __call__[T](
        self, trace: TraceType, input_type: type[T]
    ) -> AsyncGenerator[T, TraceType]: ...
```

libs/giskard-checks/src/giskard/checks/core/input_generator.py
```diff
@@ -18,7 +18,11 @@ def __call__(
...
     def __call__(
-        self, trace: TraceType, input_type: type[str | BaseModel] | None = None
-    ) -> AsyncGenerator[str | BaseModel, TraceType]:
+        self, trace: TraceType, input_type: Any | None = None
```
kevinmessiaen: input_type is either a type or None

```suggestion
        self, trace: TraceType, input_type: type[Any] | None = None
```

libs/giskard-checks/src/giskard/checks/core/interaction/interact.py
```diff
@@ -48,11 +47,13 @@ def _infer_input_type(outputs: object) -> type | None:
...
+    try:
+        TypeAdapter(first_param_type)
+    except Exception:
+        return None
+    return first_param_type
```
kevinmessiaen, line 57: Let's add tests for this path. If we find it's actually impossible for get_type_hints to return a type Pydantic can't handle, we should remove these lines entirely to avoid the significant computational overhead of TypeAdapter.

If we keep it, let's catch specific errors to avoid swallowing unexpected exceptions.

```suggestion
    try:
        # We don't need the instance, just to check if initialization is possible
        TypeAdapter(first_param_type)
    except (PydanticUserError, TypeError):
        # PydanticUserError: Raised if the type is not supported by Pydantic
        # TypeError: Raised if first_param_type isn't a valid "type" (e.g. an instance)
        return None
```

libs/giskard-checks/src/giskard/checks/generators/base.py
```diff
@@ -56,7 +56,7 @@ async def get_inputs(self, trace: TraceType) -> dict[str, Any]:
 
     @override
     async def __call__(
-        self, trace: TraceType, input_type: type[str | BaseModel] | None = None
+        self, trace: TraceType, input_type: Any | None = None
```
kevinmessiaen: Same here

```suggestion
        self, trace: TraceType, input_type: type[Any] | None = None
```

libs/giskard-checks/tests/core/test_input_generator.py
```diff
@@ -18,11 +18,11 @@ class MyModel(BaseModel):
 
 @InputGenerator.register("test_typed_generator")
 class TypedGenerator(InputGenerator[ConcreteTrace]):
-    received_input_type: type | None = None
+    received_input_type: Any | None = None
```
kevinmessiaen: Same
```suggestion
    received_input_type: type[Any] | None = None
```

libs/giskard-checks/tests/core/test_input_generator.py
```diff
@@ -18,11 +18,11 @@ class MyModel(BaseModel):
...
 
     @override
     async def __call__(
-        self, trace: ConcreteTrace, input_type: type[str | BaseModel] | None = None
+        self, trace: ConcreteTrace, input_type: Any | None = None
```
kevinmessiaen: Same


```suggestion
        self, trace: ConcreteTrace, input_type: type[Any] | None = None
```

libs/giskard-checks/tests/core/test_input_type_inference.py
```diff
@@ -70,7 +98,7 @@ def _repr_prompt_(self) -> str:
 
 @InputGenerator.register("recording_generator")
 class RecordingGenerator(InputGenerator[RecordingTrace]):
-    received_input_type: type | None = None
+    received_input_type: Any | None = None
```
kevinmessiaen: Same here

```suggestion
    received_input_type: type[Any] | None = None
```

kevinmessiaen (changes requested):
Thanks for your submission, this look great. I have some minor suggestions

kevinmessiaen (approved):

PR #2472, opened 2026-05-20
feat(examples): add chatbot scan example

kevinmessiaen:
## Summary

- Adds `examples/chatbot_scan/` — an end-to-end example showing how to scan a stateful LLM chatbot with a generated suite
- Chatbot is backed by any LiteLLM-compatible model via `acompletion`; conversation history is threaded through a typed `LLMTrace` subclass
- `.env.example` provided; `python-dotenv` loads credentials so no key is ever inlined in the run command
- Suite is generated with `generate_suite`, run against the chatbot target, and the full `SuiteResult` is serialised to `result.json`

## Test plan

- [ ] Copy `.env.example` → `.env`, set `OPENAI_API_KEY`
- [ ] `uv run python examples/chatbot_scan/chatbot_scan.py` completes without error
- [ ] Pass rate is printed to stdout
- [ ] `result.json` is written to `examples/chatbot_scan/`


pierlj (dismissed):
LGTM

PR #2474, opened 2026-05-20
feat(checks): add parallel scenario execution to Suite.run

### LINKED ISSUE #2471: `Suite.run()`: add parallel flag to run scenarios concurrently
# **Summary**

`Suite.run()` currently executes scenarios one at a time in a serial `for` loop. For suites with many independent scenarios making LLM API calls, this is a significant bottleneck. Adding a `parallel` flag to `Suite.run()` would allow all scenarios to run concurrently via `asyncio.TaskGroup`.

**Motivation**

A suite of 50 adversarial scenarios, each taking ~2 s, takes ~100 s serially. Running them in parallel would take ~2 s (assuming sufficient RPM). Since scenarios are stateless and share no mutable state (each gets its own fresh `Trace`), parallel execution is safe by default.

**Proposed API**

```python
# serial (current default — unchanged)
result = await suite.run()

# parallel (new)
result = await suite.run(parallel=True)
```

No change to the `Suite` model definition. The flag is a call-time override on `run()`, consistent with how `target` already works.

**Implementation sketch**

In `suite.py`, `Suite.run()`:

```python
# serial (existing)
for scenario in self.scenarios:
    result = await scenario.run(target=target, return_exception=return_exception)
    results.append(result)

# parallel (new branch)
async with asyncio.TaskGroup() as task_group:
    tasks = [
        task_group.create_task(
            scenario.run(target=target, return_exception=return_exception)
        )
        for scenario in self.scenarios
    ]
results = [task.result() for task in tasks]
```

`TaskGroup` cancels all sibling tasks if any task raises, giving clean structured concurrency. Result order is preserved by iterating `tasks` in the same order as `self.scenarios`.

**Edge cases / open questions**

- **`return_exception=False` + parallel**: if one scenario raises, `TaskGroup` cancels remaining tasks and re-raises. This is the right default, callers use `return_exception=True` when they want to collect errors without aborting.
- **Concurrency cap**: LLM generator already handle concurrency cap.
- **Telemetry**: the `checks_suite_run_started` / `checks_suite_run_finished` events should record `parallel=True` in their properties.

**Files to touch**

- `libs/giskard-checks/src/giskard/checks/scenarios/suite.py` — add `parallel: bool = False` to `run()`, branch on it
- `libs/giskard-checks/tests/core/test_suite.py` — add parallel execution tests
- `libs/giskard-checks/src/giskard/checks/scenarios/__init__.py` — no change expected

**Acceptance criteria**

- [ ] `suite.run()` behaves identically to today (serial, no regression)
- [ ] `suite.run(parallel=True)` runs all scenarios concurrently via `asyncio.TaskGroup`
- [ ] Result order matches scenario order regardless of completion order
- [ ] Telemetry captures the `parallel` flag
- [ ] Unit tests cover both serial and parallel paths

harsh21234i:
Closes #2471

  ## Summary
  - add a `parallel: bool = False` flag to `Suite.run()`
  - run scenarios concurrently with `asyncio.TaskGroup` when `parallel=True`
  - preserve result order to match scenario order
  - include the `parallel` flag in `checks_suite_run_started` and `checks_suite_run_finished` telemetry properties

  ## Behavior
  - default behavior is unchanged: `suite.run()` still runs serially
  - `suite.run(parallel=True)` runs scenarios concurrently
  - when `return_exception=False`, parallel execution fails fast and cancels sibling tasks if one scenario raises
  - when `return_exception=True`, scenario-level exception handling is preserved through `scenario.run(...)`

  ## Testing
  - `uv run -m pytest -q libs/giskard-checks/tests/core/test_suite.py`
  - `uv run ruff check libs/giskard-checks/src/giskard/checks/scenarios/suite.py libs/giskard-checks/tests/core/test_suite.py`


libs/giskard-checks/src/giskard/checks/scenarios/suite.py
```diff
@@ -122,25 +130,59 @@ async def run(
...
+                            return_exception=return_exception,
+                        )
+
+                    async with semaphore:
+                        return await scenario.run(
+                            target=target,
+                            return_exception=return_exception,
+                        )
+
+                async with asyncio.TaskGroup() as task_group:
```
kevinmessiaen: Since `TaskGroup` raise `ExceptionGroup` I would either document that `parallel=true` raise `ExceptionGroup` or unwrap the exception.

kevinmessiaen (changes requested):
Thanks for your PR, looking good, a few comments.

Formatting is broken, make sure to fix by running `make format`

harsh21234i:
Updated the PR.

  I fixed the `parallel=True` exception behavior so the common single-failure case re-raises the original exception
  instead of surfacing a raw `ExceptionGroup`, and I also formatted the touched files.

  `make format` is not available in my current Windows environment, so I ran the formatter directly with `uv run ruff
  format` on the changed files instead, then verified with `ruff check` and the focused suite tests.

  Checks run:
  - `uv run ruff format libs/giskard-checks/src/giskard/checks/scenarios/suite.py libs/giskard-checks/tests/core/
  test_suite.py`
  - `uv run ruff check libs/giskard-checks/src/giskard/checks/scenarios/suite.py libs/giskard-checks/tests/core/
  test_suite.py`
  - `uv run -m pytest -q libs/giskard-checks/tests/core/test_suite.py`

kevinmessiaen (approved):

kevinmessiaen:
Thanks for your contribution @harsh21234i 

I've refactored a bit the run method to have two separate methods for sequential/parallel

PR #2478, opened 2026-05-20
feat(giskard-checks): rego policy checks

educhesne:
## Description

Add a RegoPolicy check, evaluating a boolean rule in a rego policy, with a jsonpath of the trace as input (and extra data).

## Type of Change

- [X] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 🔐 Security fix

## Coding agents

If this PR was opened by or with a coding agent, read **[AGENTS.md](https://github.com/Giskard-AI/giskard-oss/blob/main/AGENTS.md)** at the repository root first — **it is required** (setup, `make` commands, expedited agent PR workflow, etc.).

## Checklist

- [X] I've read the [`CODE_OF_CONDUCT.md`](https://github.com/Giskard-AI/ai-inspector/blob/master/CODE_OF_CONDUCT.md) document.
- [X] I've written tests for all new methods and classes that I created.
- [X] I've written the docstring in NumPy format for all the methods and classes that I created or modified.
- [X] I've updated the `uv.lock` running `uv lock` (only applicable when `pyproject.toml` has been
  modified)


libs/giskard-checks/pyproject.toml
```diff
@@ -17,13 +17,20 @@ dependencies = [
     "rich>=14.2.0,<16",
 ]
 
+[project.optional-dependencies]
+regorus = ["regorus"]
```
kevinmessiaen: [`tool.uv.sources`](https://docs.astral.sh/uv/concepts/projects/dependencies/#project-dependencies) are only used during development.

> Sources are only respected by uv. If another tool is used, only the definitions in the standard project tables will be used. If another tool is being used for development, any metadata provided in the source table will need to be re-specified in the other tool's format.

Once built it's not preserved into the wheel metadata making `pip install giskard-checks[regorus]` to fail.

I would suggest using this [community mirror](https://pypi.org/project/celine-regorus/#description) that seems to be maintained automatically

kevinmessiaen (changes requested):

libs/giskard-checks/pyproject.toml
educhesne: I replaced the dependency with celine-regorus

libs/giskard-checks/src/giskard/checks/builtin/rego_policy.py
```diff
@@ -0,0 +1,174 @@
...
+from typing import Any, override
+
+from pydantic import Field
+
+from ..core import Trace
+from ..core.check import Check
+from ..core.extraction import JSONPathStr, NoMatch, resolve
+from ..core.result import CheckResult
+
+_REGORUS_INSTALL_HINT = (
```
davidberenstein1957, line 15: perhaps, we can make this a more general utils we can use across other checks too?

libs/giskard-checks/src/giskard/checks/builtin/rego_policy.py
```diff
@@ -0,0 +1,174 @@
...
+    data : dict, optional
+        Static data document merged via ``engine.add_data``
+        (default: empty dict).
+
+    Examples
+    --------
+    >>> from giskard.checks import RegoPolicy
+    >>> check = RegoPolicy(
+    ...     policy='''package giskard\\n\\ndefault allow = false\\n\\nallow if {\\n    input.role == "admin"\\n}''',
+    ...     rule="data.giskard.allow",
```
davidberenstein1957, line 100: I would format this reference code nicer so it is clearer to read what is goign on in the example.

libs/giskard-checks/src/giskard/checks/builtin/rego_policy.py
```diff
@@ -0,0 +1,174 @@
...
+    ... )
+    """
+
+    policy: str = Field(..., description="Inline Rego policy source.")
```
davidberenstein1957, line 104: Should we validate the policy formatting on class init?

libs/giskard-checks/src/giskard/checks/builtin/rego_policy.py
```diff
@@ -0,0 +1,174 @@
...
+        try:
+            regorus = importlib.import_module("regorus")
+        except ImportError:
+            return CheckResult.error(
+                message=_REGORUS_INSTALL_HINT,
+                details=details,
+            )
+
+        try:
+            engine = regorus.Engine()  # pyright: ignore[reportAttributeAccessIssue]
```
davidberenstein1957: Perhaps we should initialize this on init already instead of doing that every execution?

libs/giskard-checks/src/giskard/checks/builtin/rego_policy.py
```diff
@@ -0,0 +1,174 @@
...
+        description="Fully qualified boolean rule path (e.g. data.giskard.allow).",
+    )
+    key: JSONPathStr = Field(
+        default="trace.last.outputs",
+        description="JSONPath expression to extract the OPA input document.",
+    )
+    data: dict[str, Any] = Field(
+        default_factory=dict,
+        description="Static data document merged into the policy engine.",
+    )
```
davidberenstein1957, line 116: How and why is this useful? This would be something like.
acces rules
```
{"david": True}
```
which we can then use i.c.m. the dynamic trace config and validate against that?

libs/giskard-checks/tests/builtin/test_rego_policy.py
```diff
@@ -0,0 +1,318 @@
+"""Tests for the RegoPolicy check."""
+
+import importlib.util
+from typing import Any
+
+import pytest
+from giskard.checks import Check, CheckStatus, Interaction, RegoPolicy, Trace
+
+BOOLEAN_POLICY = """\
+package giskard
```
davidberenstein1957, line 11: why do we need `package giskard`

libs/giskard-checks/tests/builtin/test_rego_policy.py
```diff
@@ -0,0 +1,318 @@
...
+
+    assert isinstance(restored, RegoPolicy)
+    assert restored.policy == BOOLEAN_POLICY
+    assert restored.rule == "data.giskard.allow"
+    assert restored.data == {"roles": ["admin"]}
+
+
+@pytest.mark.regorus
+async def test_allow_passes() -> None:
+    pytest.importorskip("regorus")
```
davidberenstein1957: can't this be a decorated either, also perhaps a class could be nice if we copy past the marks and import behavior.

davidberenstein1957 (changes requested):
Hi, looks great but left a couple of remarks. After resolving them. Would you be able to also add a PR for the documenation update?

libs/giskard-checks/src/giskard/checks/builtin/rego_policy.py
educhesne: the only formatting validation is the compilation of the policy; I've added it in a model validator

libs/giskard-checks/src/giskard/checks/builtin/rego_policy.py
educhesne: This would be the parameters of the policy: the roles for each users, the actions and resources each role gives access to etc.
Typically the logic of the policy is fixed (in the rego code) but the data can be updated (new role assignment, new resources).

libs/giskard-checks/tests/builtin/test_rego_policy.py
educhesne: this is a namespace declaration, it is arbitrary.
it means the document defined here is accessible with rules starting with "data.giskard..."

libs/giskard-checks/tests/builtin/test_rego_policy.py
educhesne: I added a hook in conftest

libs/giskard-checks/src/giskard/checks/builtin/rego_policy.py
educhesne: what do you mean....?

educhesne:
The CI and build workflow have not been updated but I don't know what is the rule with optional dependencies... wdyt ?

kevinmessiaen:
> The CI and build workflow have not been updated but I don't know what is the rule with optional dependencies... wdyt ?

I've updated the CI to run both test with and without regorus installed

davidberenstein1957 (approved):

PR #2481, opened 2026-05-21
  feat(checks): add Trace.steps() helper for step-based grouping

### LINKED ISSUE #2419: giskard-checks: add Trace.steps() helper for step-based interaction grouping
## Feature Request

  `Trace` currently exposes a flat `interactions` list, but there is already a TODO for a step-oriented helper in:

  `libs/giskard-checks/src/giskard/checks/core/interaction/trace.py`

  ```python
  # TODO def steps() -> list[list[Interaction[InputType, OutputType]]]: # Index based

  ## Motivation

  For multi-step scenarios, it is often more useful to work with interactions grouped by step rather than as one flat list. A
  first-class Trace.steps() helper would make this easier for check authors and reduce repeated custom grouping logic.

  ## Alternatives

  Today users have to manually traverse and regroup trace.interactions, which is repetitive and error-prone.

  ## Additional context

  A useful implementation would:

  - define a clear and stable grouping rule
  - document how steps are derived from the trace
  - include tests for empty traces, single-step traces, and multi-step traces

harsh21234i:
Closes Giskard-AI/giskard-oss#2419

## Summary

* add `Trace.steps()` to group interactions by logical scenario step
* stamp `step_index` metadata onto interactions at the scenario runner boundary
* preserve backward compatibility by treating interactions without `step_index` as step `0`
* add tests for empty traces, single-step traces, multi-step traces, and real runner integration

## Design

`Trace` itself does not know `Scenario.steps`, so the grouping signal is recorded at runtime where step boundaries are  <br>actually known: `ScenarioRunner`.

This keeps the architecture aligned with the existing design:

* `Scenario` owns step structure
* `ScenarioRunner` materializes interactions
* `Trace` remains an immutable history object with helper accessors over recorded data

## Testing

* `uv run -m pytest -q libs/giskard-checks/tests/trace/test_trace.py`
* `uv run ruff check libs/giskard-checks/src/giskard/checks/core/interaction/trace.py libs/giskard-checks/src/giskard/ checks/scenarios/runner.py libs/giskard-checks/tests/trace/test_trace.py`


kevinmessiaen:
Thanks for your contribution! I will take a look afterward.

Quick note: I think it could be nice to have a `last_step` property (similar to `last` interaction)

harsh21234i:
I addressed the remaining comments in two parts:
  - `Trace.steps()` now uses a strict `type(step_index) is not int` check
  - step tagging no longer rewrites the whole trace after each step

  Instead of post-processing `trace.interactions` step-by-step, `ScenarioRunner` now wraps each step interaction spec and stamps `step_index` on each
  yielded `Interaction` before it is added to the trace. That keeps the trace immutable while avoiding the repeated full-trace copy pattern the earlier
  implementation had.

  I also added coverage for multi-yield interactions within a single step to make sure all generated interactions keep the same step index.

harsh21234i:
hey @kevinmessiaen can you look into this??

harsh21234i:
any update??

harsh21234i:
any update @kevinmessiaen @henchaves ?

PR #2486, opened 2026-05-22
feat(giskard-llm): pass through provider transport config

Hartorn:
## Summary
- Forward caller-provided `http_client` and `default_headers` through giskard-llm providers to their native SDK clients.
- Map Google transport config into `google.genai.types.HttpOptions`, preserving explicit `http_options` fields over convenience kwargs.
- Document the transport options and add an example for custom CA bundles and default headers.

## Notes
- This does not change Azure AI behavior: `azure_ai/` still uses `AsyncAzureOpenAI`, keeps `/models` stripping, and keeps the existing `AZURE_AI_API_VERSION` default behavior.
- `http_client` is caller-owned and is not closed by giskard-llm.
- This intentionally does not add `ssl_verify`, `default_query`, `max_retries`, or Azure AI Foundry OpenAI v1 behavior.

## Test plan
- `uv run --extra all pytest tests/test_providers.py tests/test_routing.py -q`
- `uv run ruff check .`
- `uv run ruff format --check .`
- pre-commit during commit:
  - ruff check
  - ruff format
  - basedpyright
  - detect-secrets


libs/giskard-llm/src/giskard/llm/providers/google.py
```diff
@@ -120,6 +124,31 @@ def _import_genai_errors() -> Any:
...
+        )
+
+    options = genai_types.HttpOptions.model_validate(http_options)
+    updates: dict[str, Any] = {}
+    if http_client is not None and options.httpx_async_client is None:
+        updates["httpx_async_client"] = http_client
+    if default_headers is not None and options.headers is None:
+        updates["headers"] = default_headers
+
+    return options.model_copy(update=updates) if updates else options
```
kevinmessiaen: `model_copy` does not trigger `model_validate`. 

> Note: the data is not validated before creating the new model. You should trust this data. [Source](https://pydantic.com.cn/en/api/base_model/#pydantic.BaseModel.model_copy)

libs/giskard-llm/src/giskard/llm/providers/anthropic.py
```diff
@@ -99,6 +101,8 @@ def __init__(
         base_url: str | None = None,
         timeout: float | None = None,
         merge_system: bool = False,
+        http_client: Any | None = None,
```
kevinmessiaen: I would add type hint to clarify what client is expected. (lazy since httpx is optional/transitive IIRC)

```suggestion
        http_client: "httpx.AsyncClient | None" = None,
```

kevinmessiaen (commented):
Looking good overall.

I would add type hint instead of any, using forward reference to allow having type safety when using the lib.

PR #2418, opened 2026-04-20
checks: make suite failure reporting configurable

harsh21234i:
#2417  
  ## Summary

  This PR makes `SuiteResult` failure reporting configurable instead of hard-coding the rich-report output to 20 failed/errored
  scenarios.

  ## Changes

  - Added `max_reported_failures` to `Suite` with a default of `20`
  - Added `max_reported_failures` to `SuiteResult` with support for `None` to show all failures
  - Updated `Suite.run()` to propagate the configured value into the returned `SuiteResult`
  - Updated `SuiteResult.__rich_console__` to honor the configured limit in both the detailed and summary sections
  - Added tests for:
    - propagation from `Suite.run()`
    - bounded rich rendering
    - unbounded rich rendering when set to `None`

PR #2497, opened 2026-05-31
fix(giskard-checks): make SuiteResult failure-detail limit configurable

### LINKED ISSUE #2417: giskard-checks: make SuiteResult failure-detail limit configurable
## Feature Request

  `SuiteResult.__rich_console__` currently hard-codes the number of failures/errors shown in the detailed output to 20:

  `libs/giskard-checks/src/giskard/checks/core/result.py`

  python
  n_loggable_failures = 20  # TODO: make this configurable

  ## Motivation

  When running large suites, a fixed limit of 20 is not always appropriate. Some users want a smaller limit to keep CI
  logs readable, while others want a larger limit for local debugging. Making this configurable would improve the UX
  without changing the default behavior.

  ## Alternatives

  Users can currently only accept the hard-coded limit. There is no public way to tune the verbosity.

  ## Additional context

  A good solution could keep the default at 20 while allowing configuration via a constructor parameter, rendering
  option, or settings value. Tests should cover both the default behavior and a custom limit.

Oxygen56:
## Summary

- Added `n_loggable_failures` parameter to `SuiteResult.__init__()` with default value 20
- Replaced hardcoded limit in `__rich_console__` with the configurable parameter
- Added test coverage for default and custom limit values

Fixes #2417

## Test Plan

- [ ] Default limit (20) behavior preserved
- [ ] Custom limit correctly restricts displayed failures
- [ ] Existing SuiteResult tests continue to pass

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/core/result.py
+++ b/libs/giskard-checks/src/giskard/checks/core/result.py
@@ -493,6 +493,8 @@ class SuiteResult(BaseResult, frozen=True):
         Scenario results produced during the suite execution.
     duration_ms : int
         Total execution time in milliseconds.
+    n_loggable_failures : int
+        Maximum number of failures/errors to display in detailed output (default 20).
     passed_count : int
         Number of scenarios that passed.
     failed_count : int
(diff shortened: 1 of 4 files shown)


davidberenstein1957:
Closing as duplicate: configurable suite failure reporting is being consolidated in #2418 (maintainer preference: env-var based config). Thanks!

PR #2500, opened 2026-06-02
fix(checks): validate suite max concurrency

### LINKED ISSUE #2499: giskard-checks: validate Suite.run(max_concurrency) as a strict positive integer
### Issue Type

Bug

### Giskard Library Version

giskard-checks 1.0.2b3

### Current Behaviour?

```shell
## Problem

  `Suite.run(parallel=True, max_concurrency=...)` documents `max_concurrency` as an optional positive integer, but
  runtime validation only checks `< 1`.

  Because of that, non-strict values are accepted or fail with unclear errors:

  
  await suite.run(parallel=True, max_concurrency=True)  # accepted
  await suite.run(parallel=True, max_concurrency=1.5)   # accepted
  await suite.run(parallel=True, max_concurrency="2")   # TypeError: '<' not supported...

  This is surprising because max_concurrency controls an asyncio.Semaphore, so invalid values should be rejected
  consistently before execution starts.

  ## Expected behavior

  max_concurrency should only accept real integers greater than 0.

  - None should remain valid
  - 1, 2, etc. should be valid
  - 0, negative integers, True, floats, and strings should raise a clear ValueError or TypeError
```

```shell
## Suggested fix

  Add explicit validation in Suite.run() before _run_parallel():

  - reject bool
  - reject non-int
  - reject values < 1

  Also add tests covering True, 1.5, "2", 0, and valid integer values.
```

### Relevant log output

```shell

```

harsh21234i:
Closes #2499

  ## Summary
  - validate `Suite.run(max_concurrency=...)` as a strict positive integer when `parallel=True`
  - reject bool, float, string, zero, and negative values with clear errors
  - add parameterized coverage for invalid `max_concurrency` values
  - make the parallel timing test compare against an actual serial baseline

  ## Testing
  - `uv run -m pytest -q libs/giskard-checks/tests/core/test_suite.py`
  - `uv run ruff check libs/giskard-checks/src/giskard/checks/scenarios/suite.py libs/giskard-checks/tests/core/
  test_suite.py`


harsh21234i:
hey @kevinmessiaen @davidberenstein1957  can someone look into this?

PR #2504, opened 2026-06-02
fix(giskard-checks): use NOT_PROVIDED sentinel as default for expected_value 

### LINKED ISSUE #2501: comparison checks allow missing expected value
### Current Behaviour?

```shell
## Problem

  Comparison checks such as `Equals`, `GreaterThan`, `LesserThan`, etc. document that either `expected_value` or
  `expected_value_key` must be provided.

  However, creating a comparison check without either one is currently accepted:

  
  from giskard.checks.builtin.comparison import Equals, GreaterThan

  Equals(key="trace.last.outputs")
  GreaterThan(key="trace.last.outputs")

  These are accepted and silently use expected_value=None:

  expected_value=None
  expected_value_key=NotProvided()

  This is surprising because it creates a valid-looking check even though the user forgot to configure the expected
  value.

  ## Expected behavior

  If both expected_value and expected_value_key are omitted, the constructor should raise a clear validation error.

  Explicit expected_value=None should still be supported if comparing against None is intended.
```

```shell
## Suggested fix

  Use a sentinel default for expected_value so the code can distinguish:

  - omitted expected_value
  - explicit expected_value=None

  Then update the validator to reject only the omitted case.

  ## Acceptance criteria

  - Equals(key="trace.last.outputs") raises a validation error
  - GreaterThan(key="trace.last.outputs") raises a validation error
  - Equals(key="trace.last.outputs", expected_value=None) remains valid
  - expected_value_key=... remains valid
  - tests cover omitted vs explicit None
```

### Relevant log output

```shell

```

### LINKED ISSUE #2545: comparison checks allow omitted expected value
### Current Behaviour?

```shell
## Problem

  Comparison checks such as `Equals`, `GreaterThan`, `LesserThan`, etc. are supposed to require either `expected_value`
  or `expected_value_key`.

  However, constructing a comparison check without either one is currently accepted:

  
  from giskard.checks import Equals, GreaterThan

  print(Equals(key="trace.last.outputs"))
  print(GreaterThan(key="trace.last.outputs"))

  Current output:

  key='trace.last.outputs' expected_value=None expected_value_key=NotProvided()
  key='trace.last.outputs' expected_value=None expected_value_key=NotProvided()

  This happens because expected_value defaults to None, while the validator checks for NotProvided. So an omitted
  expected_value is treated the same as an explicit None.

  ## Expected behavior

  If both expected_value and expected_value_key are omitted, the constructor should raise a clear validation error.

  Explicit expected_value=None should still remain valid if the user intentionally wants to compare against None.
```

```shell
## Suggested fix

  Use a sentinel default for expected_value, similar to expected_value_key, so the code can distinguish:

  - omitted expected_value
  - explicit expected_value=None

  Then update the validator to reject only the omitted case.

  ## Acceptance criteria

  - Equals(key="trace.last.outputs") raises a validation error
  - GreaterThan(key="trace.last.outputs") raises a validation error
  - Equals(key="trace.last.outputs", expected_value=None) remains valid
  - expected_value_key=... remains valid
  - tests cover omitted vs explicit None
```

nuthalapativarun:
## Description

`ComparisonCheck.expected_value` defaulted to `None`, making it impossible to distinguish between an omitted value and an explicit `None`. The `model_validator` that enforces *"either `expected_value` or `expected_value_key` must be provided"* never fired because `None` is not an instance of `NotProvided`.

This PR changes the default to `NOT_PROVIDED` (the existing sentinel from `giskard.core`), so:
- Omitting both `expected_value` and `expected_value_key` now raises `ValueError` as documented
- Explicit `expected_value=None` remains valid (compares the actual value against `None`)
- `expected_value_key=...` continues to work without change

The `_compare` and `run` methods are untouched — only the field default and its docstring change.

## Related Issue

Closes #2501

## Type of Change

## Checklist
- [x] I've read CODE_OF_CONDUCT.md
- [x] I've read CONTRIBUTING.md
- [x] I've written tests for all new methods and classes
- [x] I've written NumPy-format docstrings for new/modified methods
- [x] I've updated uv.lock (only if pyproject.toml changed)


libs/giskard-checks/src/giskard/checks/builtin/comparison.py
```diff
@@ -35,9 +35,9 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
     key: JSONPathStr = Field(
         ..., description="The key to extract the actual value from the trace"
     )
-    expected_value: ExpectedType | None = Field(
-        default=None,
-        description="The expected value to compare against. If None, the expected value is extracted from the trace using the expected_value_key.",
+    expected_value: ExpectedType | None | NotProvided = Field(
```
kevinmessiaen: Making sense, I would update the signature to look like this, since `ExpectedType` can be `NoneType`

```suggestion
    expected_value: ExpectedType | NotProvided = Field(
```

kevinmessiaen (changes requested):
Thanks for your PR, looking good. Just one small suggestion to make the signature cleaner

nuthalapativarun:
Thanks for the review, @kevinmessiaen! I've applied the suggested simplification — `expected_value` is now typed as `ExpectedType | NotProvided` (the explicit `None` was redundant since `ExpectedType` can already include `NoneType`).

I've also rebased the branch onto the latest `main`. All comparison tests pass (60 passed).

davidberenstein1957:
Closing as duplicate: the same `expected_value` / `NOT_PROVIDED` fix is covered by #2546, which is approved and ready to merge. Thanks for the contribution!

davidberenstein1957:
Reopening: this PR has author follow-through on @kevinmessiaen’s review (changes applied). We are keeping the duplicate where the opener engaged, and closing #2546 instead.

nuthalapativarun:
Hi @kevinmessiaen — confirming the requested simplification is in place: `expected_value` is now typed as `ExpectedType | NotProvided` (defaulting to `NOT_PROVIDED`), dropping the redundant explicit `None` from the signature. Regression tests cover omitting the value (raises), passing explicit `None` (valid), and using `expected_value_key` alone (valid). Branch is rebased on latest main and all comparison tests pass locally. Would appreciate a re-review when you have a moment — thanks!

kevinmessiaen (approved):
Thanks for the fix!

PR #2505, opened 2026-06-02
fix(giskard-llm): preserve thought_signature in Gemini TextContent translation 🤖🤖🤖🤖

### LINKED ISSUE #2503: Bug missing `thought_signature` for recent gemini models
In giskard-llm we can't use recent gemini models (>=3) for agentic workflows because google expects a `thought_signature` to be sent in the conversation history. <br><br>Docs is here: [https://ai.google.dev/gemini-api/docs/thought-signatures](<https://ai.google.dev/gemini-api/docs/thought-signatures>)<br><br>Recent Gemini thinking models return `thought_signature` on response content parts. `giskard-llm` currently drops this field when normalizing Google responses into `TextContent` / `ToolCall`, so subsequent requests cannot pass the signature back in conversation history. This can degrade reasoning continuity and may cause Gemini 3 function-calling requests to fail validation.

Expected behavior: `giskard-llm` should preserve Gemini `thought_signature` exactly as returned by the Google SDK and include it again when serializing reused assistant messages back to Google `generateContent`.

nuthalapativarun:
## Description

Gemini 3+ thinking models return a `thought_signature` field on response `Part` objects. The `giskard-llm` Google translator was silently dropping this field when normalising responses into `TextContent`, so subsequent multi-turn requests could not replay the signature — breaking agentic workflows that require it.

This PR:
- Adds `thought_signature: str | None = None` to `TextContent` in `libs/giskard-llm/src/giskard/llm/types/chat.py`
- Populates it in `part_content_to_giskard` via `getattr(part, "thought_signature", None)` for forward compatibility with any SDK version
- Includes `thought_signature` in the Google `PartDict` serializer only when it is non-`None`, so existing payloads are unaffected
- Adds a dedicated test module covering the preserve and omit paths for both pure-Python mocks and (when the SDK is installed) real `GenerateContentResponse` round-trips

## Related Issue
Closes #2503

## Type of Change

## Checklist
- [x] I've read CODE_OF_CONDUCT.md
- [x] I've read CONTRIBUTING.md
- [x] I've written tests for all new methods and classes
- [x] I've written NumPy-format docstrings for new/modified methods
- [x] I've updated uv.lock (only if pyproject.toml changed)

### DIFF
--- a/libs/giskard-llm/src/giskard/llm/translators/google_chat.py
+++ b/libs/giskard-llm/src/giskard/llm/translators/google_chat.py
@@ -76,7 +76,10 @@ def _text_content(text: str) -> "PartDict":
 def serialize_text_content(
     content: TextContent, _info: SerializationInfo
 ) -> "PartDict":
-    return _text_content(content.text)
+    part: "PartDict" = {"text": content.text}
+    if content.thought_signature is not None:
+        part["thought_signature"] = content.thought_signature  # type: ignore[typeddict-unknown-key]
+    return part
 
 
 @RefusalContent.register_serializer(_PROVIDER)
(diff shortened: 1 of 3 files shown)


kevinmessiaen:
Done in #2502

PR #2512, opened 2026-06-04
fix(checks): accept parsed string JSON values

### LINKED ISSUE #2511: JsonValid rejects already parsed string values
### Current Behaviour?

```shell
## Problem

  `JsonValid` documentation says the extracted value can be either:

  - a JSON string
  - an already parsed JSON-compatible value such as dict, list, string, number, boolean, or None

  However, plain Python string values are always treated as raw JSON text and passed to `json.loads()`.

  That means an already parsed string value fails:

  
  from giskard.checks.builtin.json_valid import JsonValid
  from giskard.checks.core.interaction import Trace, Interaction

  trace = Trace(interactions=[Interaction(inputs="x", outputs="hello")])
  result = await JsonValid().run(trace)

  print(result.status)
  print(result.message)

  Current result:

  CheckStatus.FAIL
  Value at key 'trace.last.outputs' is not valid JSON: Expecting value: line 1 column 1 (char 0)

  This is surprising because "hello" is a JSON-compatible scalar value once already parsed.

  ## Expected behavior

  JsonValid should distinguish between:

  - raw JSON text that needs parsing, e.g. '{"name": "Alice"}'
  - already parsed JSON-compatible Python values, e.g. "hello", 42, True, None, {"name": "Alice"}

  A plain Python string should be accepted as an already parsed JSON-compatible scalar, or the documentation should be clarified if only serialized JSON
  strings are intended.
```

```shell
## Suggested fix

  Update _parse_json() behavior or add an option to clarify intent.

  Possible approach:

  - keep parsing strings that look like serialized JSON objects/arrays/scalars
  - if parsing fails for a Python string, treat it as an already parsed JSON scalar
  - keep schema validation working against the resulting parsed value

  ## Acceptance criteria

  - JsonValid passes for already parsed scalar string output like "hello"
  - serialized JSON strings like '{"name": "Alice"}' still parse and pass
  - invalid non-serializable values still fail clearly
  - schema validation still applies to the final parsed value
  - tests cover plain string, serialized JSON string, dict/list, and schema cases
```

harsh21234i:
Closes #2511 

  ## Summary
  - allow `JsonValid` to accept non-empty plain Python strings as already parsed JSON scalar values
  - keep parsing serialized JSON-looking strings such as objects, arrays, quoted strings, numbers, booleans, and null
  - keep empty strings and malformed JSON-looking strings failing as invalid JSON
  - add coverage for plain string outputs and schema validation against plain strings

  ## Testing
  - `uv run --no-sync -m pytest -q libs/giskard-checks/tests/builtin/test_json_valid.py`
  - `uv run --no-sync ruff check libs/giskard-checks/src/giskard/checks/builtin/json_valid.py libs/giskard-checks/tests/builtin/test_json_valid.py`

  Verification: 23 passed, ruff passed.


harsh21234i:
hey @kevinmessiaen @davidberenstein1957 can someone look into this?

kevinmessiaen:
Fixed by #2575

PR #2516, opened 2026-06-05
fix(docs): expose __version__ via importlib.metadata and update bug template

nuthalapativarun:
## Description
Expose `giskard.__version__` so the standard version check in the issue template works as documented.

## Related Issue
Closes #2515

## Type of Change
- [x] 🔧 Bug fix

## Checklist
- [x] Read CODE_OF_CONDUCT.md
- [x] Read CONTRIBUTING.md
- [x] Written tests for all new methods/classes
- [x] Written NumPy-format docstrings for new methods/classes
- [x] Updated uv.lock (if pyproject.toml changed)

PR #2519, opened 2026-06-05
fix: correct issue template version command

### LINKED ISSUE #2515: giskard package does not expose __version__ although issue template requires it
### Current Behaviour?

```shell
## Problem

  The repository’s issue template asks users to provide the library version with:

  
  python -c "import giskard; print(giskard.__version__)"

  But the installed package does not expose __version__, so the command fails:

  import giskard
  print(giskard.__version__)

  Current behavior:

  AttributeError: module 'giskard' has no attribute '__version__'

  This creates friction when opening bugs because the project’s own template recommends a command that does not work.

  ## Expected behavior

  giskard.__version__ should return the installed package version, or the issue template should be updated to use a supported alternative.

  ## Reproduction

  uv run python -c "import giskard; print(giskard.__version__)"

  ## Suggested fix

  One of:

  - expose __version__ from the top-level giskard package
  - or update the issue template to use importlib.metadata.version("giskard")

  For example:

  python -c "import importlib.metadata as m; print(m.version('giskard'))"

  ## Acceptance criteria

  - import giskard; print(giskard.__version__) works
  - or the issue template no longer tells users to run a broken command
  - docs and issue-reporting flow use one consistent way to retrieve the package version
```

```shell
## Suggested fix

  One of:

  - expose __version__ from the top-level giskard package
  - or update the issue template to use importlib.metadata.version("giskard")

  For example:

  python -c "import importlib.metadata as m; print(m.version('giskard'))"

  ## Acceptance criteria

  - import giskard; print(giskard.__version__) works
  - or the issue template no longer tells users to run a broken command
  - docs and issue-reporting flow use one consistent way to retrieve the package version
```

### Relevant log output

```shell

```

harsh21234i:
closes #2515
  ## Summary
  - fix the bug report template command for retrieving the Giskard version
  - replace the broken `giskard.__version__` example with `importlib.metadata.version("giskard")`
  - clarify that source checkouts should use the version from `pyproject.toml`

  ## Why
  The current issue template asks users to run:

  ```bash
  python -c "import giskard ; print(giskard.__version__)"

  but giskard.__version__ is not exposed, so that command fails and makes bug reporting harder.

  ## Testing

  - verified the updated issue template text in .github/ISSUE_TEMPLATE/giskard_bug_template.yaml

### DIFF
--- a/.github/ISSUE_TEMPLATE/giskard_bug_template.yaml
+++ b/.github/ISSUE_TEMPLATE/giskard_bug_template.yaml
@@ -33,7 +33,7 @@ body:
     id: giskard-lib-version
     attributes:
       label: Giskard Library Version
-      description: You can  obtain the Giskard library version with `python -c "import giskard ; print(giskard.__version__)"`
+      description: For pip/uv installs, run `python -c "import importlib.metadata as m; print(m.version('giskard'))"`; for source check the version in `pyproject.toml`.
       placeholder: e.g., 3.0.0a1
     validations:
       required: true
--- a/.github/ISSUE_TEMPLATE/giskard_bug_template.yaml
+++ b/.github/ISSUE_TEMPLATE/giskard_bug_template.yaml
@@ -33,7 +33,7 @@ body:
     id: giskard-lib-version
     attributes:
       label: Giskard Library Version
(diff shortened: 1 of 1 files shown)


kevinmessiaen:
Thanks for your contribution. This issue has already been solved by #2516

PR #2524, opened 2026-06-05
fix(checks): record input generation errors

### LINKED ISSUE #2523: Handle input-generation errors like check errors
## Overview

A scenario step runs in two phases: generate inputs (`with_interactions`), then run checks. When `return_exception=True`, check errors are recorded as results and the run continues, but an error during input generation still crashes the whole run. This breaks suites, where one bad scenario aborts all the others instead of just being reported as errored.

The fix: make input generation errors behave like check errors.

## To Do

* When input generation raises and `return_exception=True`, record it as an errored result instead of raising
* When `return_exception=False`, keep raising (no change)
* Include the error message and type in the result
* Add a test for the recorded error case and keep the existing "raise on error" test

## Acceptance Criteria

* `Suite(...).run(..., return_exception=True`) + a failing generator → run returns an errored result, no exception escapes
* `return_exception=False` → still raises, as before
* A suite with one failing scenario still returns results for the rest

extrasmall0:
## Summary
- convert scenario input-generation exceptions into errored check results when `return_exception=True`
- keep the default `return_exception=False` behavior raising the original exception
- add scenario and suite regression tests for recorded generator errors and suite continuation

## Testing
- `uv run ruff format libs/giskard-checks/src/giskard/checks/scenarios/runner.py libs/giskard-checks/tests/core/test_scenario.py libs/giskard-checks/tests/core/test_suite.py`
- `uv run ruff check libs/giskard-checks/src/giskard/checks/scenarios/runner.py libs/giskard-checks/tests/core/test_scenario.py libs/giskard-checks/tests/core/test_suite.py`
- `uv run pytest libs/giskard-checks/tests/core/test_scenario.py -k 'generator_exception'`
- `uv run pytest libs/giskard-checks/tests/core/test_suite.py -k 'input_generation_errors or parallel_fail_fast'`
- `make test-unit PACKAGE=giskard-checks`

Closes #2523

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/scenarios/runner.py
+++ b/libs/giskard-checks/src/giskard/checks/scenarios/runner.py
@@ -6,6 +6,7 @@ updated Trace objects via the async generator protocol.
 """
 
 import time
+import traceback
 from typing import Any, cast
 
 from giskard.core import (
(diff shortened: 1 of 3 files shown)


kevinmessiaen:
Thanks for your contribution, the fix has been landed already #2599

PR #2529, opened 2026-06-08
fix(checks): improve suite failure check info

### LINKED ISSUE #2521: Improve check information returned on Suite Result report
## Overview

When you run a scan suite and print the result:

```
suite = scan.generate_suite(...)
suite_result = suite.run(agent)
```

the failures report is hard to interpret for two reasons:

1. Check names show as `None`: Every failed check is labeled `None`, so you cant' tell what kind of check failed (conformity, groundedness, correctness, etc.). Checks don't always have a name set, and the report doesn't fall back to anything useful, so it just prints `None`.
2. Check parameters are missing: The report shows *why* a check failed (the reason text), but not the check's original parameters, for example the conformity `rule` it was testing against. Without the rule, you can't compare what was expected with what the agent actually did.

## To-Do

* When a check has no `name`, fall back to its type/kind so it never shows `None`. The result already stores `check_knd` (and `check_description`); the console renderer just needs to use them as a fallback, the same way `TestCaseResult.format_failures()`
* Capture the check's parameters when it runs (e.g. Conformity's `rule`) and show them in the failure report, next to the reason.
* Keep the report tidy: a short label + reason + parameters, without dumpling internal or irrelevant fields (e.g. generator config)

## Acceptance Criteria

* No failed check is shown as `None`. A named check shows its name; an unnamed check shows its type/kind (e.g. `Conformity`)
* The failure report shows the check's key parameters next to the reason, so the user can compare the rule against the outcome.
* The check label behaves the same across the console report, `TestCaseResult.format_failures()`, and the JUnit export.
* Example, before vs after:

```
# Before
None         FAIL   The agent's responses provided links to lyrics ...

# After
Conformity   FAIL   The agent's responses provided links to lyrics ...
rule: "Do not share copyrighted lyrics without authorization"   
```

## Notes / pointers

* Name fallback: `libs/giskard-checks/src/giskard/checks/core/result.py` -> `CheckResult.__rich_console__` (currently reads `details["check_name"]`, which is `None`).  The fix can be as small as `details.get("check_name") or details.get("check_kind") or "Unnamed check"`.
* Where result details are filled: `libs/giskard-checks/src/giskard/checks/testing/runner.py` -> `_run_check` (sets `check_kind`, `check_name`, `check_description`; this is also where check parameters could be added).

---

**Screenshots of current report:**

harsh21234i:
Closes Giskard-AI/giskard-oss#2521

## Summary

* use check kind as a fallback label when check name is missing
* capture compact check parameters in result details
* show check parameters in console and formatted failure reports
* keep JUnit check labels consistent with console/test case reports

## Testing

* `.venv\Scripts\python.exe -m pytest -q libs/giskard-checks/tests/scenarios/test_testcase.py libs/giskard-checks/ tests/export/test_junit.py`
* `uv run --no-sync ruff check libs/giskard-checks/src/giskard/checks/core/result.py libs/giskard-checks/src/giskard/ checks/testing/runner.py libs/giskard-checks/src/giskard/checks/export/junit.py libs/giskard-checks/tests/scenarios/ test_testcase.py libs/giskard-checks/tests/export/test_junit.py`


harsh21234i:
Addressed the review comments by making `details` access defensive in `CheckResult`.

  - `check_label` now falls back safely when `details` is not a dict
  - `check_params` rendering in console output and formatted failures now guards against non-dict `details`
  - added a regression test covering `CheckResult` with non-dict `details`

  Checks run:
  - `pytest -q libs/giskard-checks/tests/scenarios/test_testcase.py libs/giskard-checks/tests/export/test_junit.py`
  - `uv run --no-sync ruff check libs/giskard-checks/src/giskard/checks/core/result.py libs/giskard-checks/src/giskard/
  checks/testing/runner.py libs/giskard-checks/src/giskard/checks/export/junit.py libs/giskard-checks/tests/scenarios/
  test_testcase.py libs/giskard-checks/tests/export/test_junit.py`

kevinmessiaen (changes requested):
To me this seems not robust. We should probably make the check name a direct property of `CheckResult` instead of having to guess where it's inside the `details`

harsh21234i:
going through changes ...

harsh21234i:
Addressed your's feedback by making `check_name` a first-class field on `CheckResult` instead of inferring it from
  `details`.

  - the runner now sets `check_name` directly from the live `Check` instance
  - `check_label` reads the explicit field only
  - legacy `details`-based naming is still accepted at construction time for backward compatibility
  - updated the tests to use direct `check_name` values

  Also stabilized the suite parallel timing test so it compares parallel vs serial execution directly.

  Checks:
  - `pytest -q libs/giskard-checks/tests/core/test_suite.py libs/giskard-checks/tests/export/test_junit.py`
  - `ruff check libs/giskard-checks/src/giskard/checks/core/result.py libs/giskard-checks/src/giskard/checks/testing/
  runner.py libs/giskard-checks/src/giskard/checks/scenarios/runner.py libs/giskard-checks/tests/core/test_suite.py
  libs/giskard-checks/tests/export/test_junit.py`

harsh21234i:
any update @kevinmessiaen ?

harsh21234i:
hey @kevinmessiaen  any update?

libs/giskard-checks/src/giskard/checks/export/junit.py
```diff
@@ -24,12 +24,8 @@ def _to_json(value: Any) -> str:
...
     if isinstance(result.details, dict):
-        return str(
-            result.details.get("check_name")
-            or result.details.get("check_kind")
-            or result.details.get("name")
-            or fallback
-        )
+        label = result.check_label
+        return fallback if label == "Unnamed check" else label
     return fallback
```
davidberenstein1957, line 29: Now that `check_name` is first-class on `CheckResult`, gating on `isinstance(result.details, dict)` is wrong — a result with `check_name` set but non-dict `details` would ignore the name and fall back.

Comparing to the magic string `"Unnamed check"` is also brittle.

Prefer:

```python
def _check_label(result: CheckResult, fallback: str) -> str:
    return result.check_name or fallback
```

libs/giskard-checks/src/giskard/checks/testing/runner.py
```diff
@@ -41,13 +61,14 @@ async def _run_check[
...
             "details": {
                 **(res.details or {}),
                 "duration_ms": int((time.perf_counter() - check_start_time) * 1000),
                 "check_kind": check.kind,
-                "check_name": check.name,
                 "check_description": check.description,
-            }
+                "check_params": _check_params(check),
+            },
         }
```
davidberenstein1957, line 72: This drops writing `check_name` into `details` (previously `"check_name": check.name`). Callers that still read `result.details["check_name"]` will break quietly.

Please keep writing it for backward compatibility, e.g.:

```python
"details": {
    **(res.details or {}),
    "duration_ms": int((time.perf_counter() - check_start_time) * 1000),
    "check_kind": check.kind,
    "check_name": check.name,
    "check_description": check.description,
    "check_params": _check_params(check),
},
```

Or document this as an intentional migration if you intentionally want people to use the new field only.

libs/giskard-checks/src/giskard/checks/core/result.py
```diff
@@ -145,15 +158,37 @@ class CheckResult(BaseResult, frozen=True):
...
+        if not isinstance(details, dict):
+            return data
+
+        for key in ("check_name", "check_kind", "name"):
+            value = details.get(key)
+            if value:
+                return {**data, "check_name": value}
+
+        return data
+
```
davidberenstein1957, line 184: Nit / design: promoting `check_kind` into the `check_name` field conflates name and kind. It works for display, but means you can no longer tell whether the user set a real name.

Cleaner long-term shape:

```python
@property
def check_label(self) -> str:
    return self.check_name or self.details.get("check_kind") or "Unnamed check"
```

(or a first-class `check_kind` field) and only promote `details["check_name"]` / `details["name"]` into `check_name`. Not a blocker if you prefer to keep this for now.

libs/giskard-checks/tests/export/test_junit.py
```diff
@@ -172,6 +172,47 @@ def test_to_junit_xml_maps_failure_error_and_skip() -> None:
...
+        duration_ms=10,
+    )
+
+    root = ET.fromstring(to_junit_xml(suite_result))
+    failure = root.find("testcase/failure")
+
+    assert failure is not None
+    assert failure.attrib["type"] == "conformity"
+    assert "[FAIL] step_1.conformity: rule was not followed" in (failure.text or "")
+    assert "check_params" in (failure.text or "")
```
davidberenstein1957, line 213: This only asserts that the raw JSON dump of `details` contains `"check_params"`. JUnit still renders `details={...json...}` rather than the tidy console format from #2521 (`rule: "..."` next to the reason).

Nice-to-have: align JUnit failure text with the formatted params used in console / `format_failures`, and assert on that output here. Not a merge blocker for the console fix.

libs/giskard-checks/tests/core/test_suite.py
```diff
@@ -330,11 +343,11 @@ async def delayed_identity(inputs):
...
 
     serial_start = time.perf_counter()
-    await suite.run()
+    await suite.run(verbose=False)
     serial_duration = time.perf_counter() - serial_start
 
     parallel_start = time.perf_counter()
-    await suite.run(parallel=True)
+    await suite.run(parallel=True, verbose=False)
     parallel_duration = time.perf_counter() - parallel_start
```
davidberenstein1957, line 351: Unrelated to the suite failure check-info fix — fine as a flaky-test stabilization, but ideally a separate commit/PR so this change stays focused.

davidberenstein1957 (changes requested):
Thanks for pushing on Kevin's feedback — first-class `check_name` is the right shape, and the console / `format_failures` path looks good for #2521.

Please address the inline comments before re-review. Also please rebase onto `main` (branch is currently behind).

PR #2533, opened 2026-06-10
Fix/check kind console label

### LINKED ISSUE #2532: console failure output does not fall back to check_kind
### Current Behaviour?

```shell
## Problem

  `CheckResult` console rendering only uses `details["check_name"]` for the displayed label.

  If a check result has `check_kind` but no `check_name`, the console output shows `Unnamed check`, while JUnit export
  already falls back to `check_kind`.

  That makes failure reporting inconsistent across outputs and hides useful check identity information.

  ## Reproduction

  
  from giskard.checks.core.result import CheckResult, CheckStatus
  from rich.console import Console

  result = CheckResult(
      status=CheckStatus.FAIL,
      message="boom",
      details={"check_kind": "MyCheck"},
  )

  console = Console(record=True, force_terminal=False, no_color=True, width=120)
  console.print(result)
  print(console.export_text())

  Current output:

  Unnamed check   FAIL    boom

  Expected label:

  MyCheck
```

```shell
## Expected behavior

  When check_name is missing, the console output should fall back to:

  1. check_kind
  2. name
  3. a default placeholder

  This should match the existing fallback behavior used by JUnit export.

  ## Suggested fix

  Factor out a shared label helper for CheckResult and reuse it in:

  - rich console rendering
  - JUnit export
  - formatted failure messages, if needed

  This avoids drift between output formats.

  ## Acceptance criteria

  - console output shows check_kind when check_name is missing
  - JUnit and console use the same label fallback chain
  - tests cover check_kind-only results


  I verified this locally:
  - console currently prints `Unnamed check`
  - JUnit label helper already resolves to `MyCheck`
```

harsh21234i:
Closes #2532

  ## Summary
  - add a shared `check_label` fallback on `CheckResult`
  - use `check_name -> check_kind -> name -> Unnamed check` for console rendering
  - reuse the same label logic in JUnit export
  - add regressions for `check_kind`-only failures in console output and JUnit output
  - stabilize the suite parallel timing test so it compares parallel vs serial execution directly

  ## Testing
  - `ruff check libs/giskard-checks/src/giskard/checks/core/result.py libs/giskard-checks/src/giskard/checks/export/
  junit.py libs/giskard-checks/tests/core/test_suite.py libs/giskard-checks/tests/export/test_junit.py`
  - `pytest -q libs/giskard-checks/tests/core/test_suite.py libs/giskard-checks/tests/export/test_junit.py`

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/core/result.py
+++ b/libs/giskard-checks/src/giskard/checks/core/result.py
@@ -222,12 +222,22 @@ class CheckResult(BaseResult, frozen=True):
         """Return True if `status` is `SKIP`."""
         return self.status == CheckStatus.SKIP
 
+    @property
+    def check_label(self) -> str:
+        details = self.details if isinstance(self.details, Mapping) else {}
+        return str(
+            details.get("check_name")
+            or details.get("check_kind")
+            or details.get("name")
+            or "Unnamed check"
+        )
+
     def __rich_console__(
         self, console: Console, options: ConsoleOptions
     ) -> RenderResult:
         status = STATUS_MAPPING[self.status]
 
-        name = self.details.get("check_name", "[dim italic]Unnamed check[/dim italic]")
+        name = self.check_label
 
         if self.status == CheckStatus.FAIL or self.status == CheckStatus.ERROR:
             details = (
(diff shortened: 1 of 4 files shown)


kevinmessiaen (changes requested):
To me this seems not robust. We should probably make the check name a direct property of `CheckResult` instead of having to guess where it's inside the `details`

kevinmessiaen:
Hey @harsh21234i

Thanks for your contribution. It seems that you opened two PR for the same task: #2529

PR #2539, opened 2026-06-12
feat(scan): Add quality scan and knowledge base scenario generator

pierlj:
## Description
Added a light-weight KB object to manipulate documents and embeddings
Added a `KnowledgeBaseScenarioGenerator` based on RAGET principle. Takes a knowledge and generate scenario with groundedness checks.
Added a quality scan that collects the quality Scenario Generators (for now only one).

## Type of Change

- [ ] 🔐 Security fix

## Coding agents

## Checklist

- [ ] I've written the docstring in NumPy format for all the methods and classes that I created or modified.
  modified)


libs/giskard-scan/src/giskard/scan/quality.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+async def quality_scan[InputType, OutputType, TraceType: Trace](  # pyright: ignore[reportMissingTypeArgument]
+    target: Target[InputType, OutputType, TraceType],
+    description: str,
+    languages: list[str],
+    knowledge_base: KnowledgeBase | list[str] | None = None,
+    max_scenarios: int | None = None,
+    seed: int = 42,
+    group_by: str | None = None,
```
kevinmessiaen: I would make the default to `quality`, so that we don't forget about this when we add Hallucination scenarios.
 
```suggestion
    group_by: str | None = "quality",
```

libs/giskard-scan/src/giskard/scan/generators/knowledge_base.py
```diff
@@ -0,0 +1,153 @@
...
+            languages: BCP-47 language codes the generated questions should use.
+            max_scenarios: Maximum number of scenarios to generate. ``None``
+                uses :data:`DEFAULT_KNOWLEDGE_BASE_SCENARIOS`.
+            rng: Random generator used for reproducible seed document sampling.
+
+        Returns:
+            Generated scenarios, or an empty list when no knowledge base is
+            configured.
+        """
+        knowledge_base = normalize_knowledge_base(self.knowledge_base)
```
kevinmessiaen: I would move this inside of `quality_scan` 

Reason is if we add more generators that handle the knowledge base, passing `list[str]`  would create a `KnowledgeBase` per generator, duplicating the embedding generation

libs/giskard-scan/src/giskard/scan/utils/knowledge_base.py
```diff
@@ -0,0 +1,135 @@
...
+        If any document is missing embeddings, all embeddings are recomputed in
+        one batch. This avoids mixing vectors that may have been produced by
+        different embedding models.
+        """
+        if all(doc.embeddings is not None for doc in self.documents):
+            return
+
+        embeddings = await self._embedding_model.embed(
+            [document.content for document in self.documents]
+        )
```
kevinmessiaen: If a KnowledgeBase contains just one not-embedded document. We we compute embeddings got the whole knowledge base.

```py
kb = KnowledgeBase.from_texts([...])
await kb.closest_documents(0, 10)
kb.documents.append(EmbeddedDocument(content="new")
await kb.closest_documents(0, 10)
```

libs/giskard-scan/src/giskard/scan/utils/knowledge_base.py
```diff
@@ -0,0 +1,135 @@
...
+        """Ensure every document has embeddings from the same model.
+
+        If any document is missing embeddings, all embeddings are recomputed in
+        one batch. This avoids mixing vectors that may have been produced by
+        different embedding models.
+        """
+        if all(doc.embeddings is not None for doc in self.documents):
+            return
+
+        embeddings = await self._embedding_model.embed(
```
kevinmessiaen: This is not thread safe. Since this is a lazy and long operation called by `closest_documents`, this would cause issue when running `closest_documents` in parallel.

This will happen once `quality_scan` have more than one generator using the KB (generators are run in parallel)

libs/giskard-scan/src/giskard/scan/utils/knowledge_base.py
```diff
@@ -0,0 +1,135 @@
+"""Knowledge base primitives for document-grounded scan generators."""
+
+from typing import Self
+
+import numpy as np
+from giskard.checks.core.mixin import WithEmbeddingMixin
+from pydantic import BaseModel, model_validator
+
+
+class EmbeddedDocument(BaseModel):
```
kevinmessiaen: The name implies that the document are `Embedded`, but they are optionally embedded. We should probably rename it to `Document`

```suggestion
class Document(BaseModel):
```

libs/giskard-scan/src/giskard/scan/quality.py
```diff
@@ -0,0 +1,115 @@
...
+
+    warnings.warn(
+        "quality_scan received an empty knowledge base; knowledge-base quality scenarios will be skipped.",
+        RuntimeWarning,
+        stacklevel=2,
+    )
+    return None
+
+
+def _configured_generators(
```
kevinmessiaen: This seems not robust. If an user wants to add it's own generator to the `quality_suite_generator_registry`, it won't be possible for them to update this method and inject the `knowledge_base`.

I would move the parameters to the generator. We should probably make the generator signature more robust/flexible.

kevinmessiaen (changes requested):

libs/giskard-scan/src/giskard/scan/generators/knowledge_base.py
pierlj: Ok, but I would keep it also in the generator to allow taking list[str] as input for the generator for users who want to use it outside the scan

libs/giskard-scan/src/giskard/scan/quality.py
pierlj: I added a mixin to modify the contract for the generators that will need a KB.

libs/giskard-scan/src/giskard/scan/quality.py
```diff
@@ -0,0 +1,99 @@
...
+    target: Target[InputType, OutputType, TraceType],
+    description: str,
+    languages: list[str],
+    knowledge_base: KnowledgeBase | list[str] | None = None,
+    max_scenarios: int | None = None,
+    seed: int = 42,
+    group_by: str | None = "quality",
+    parallel: bool = True,
+    max_concurrency: int | None = None,
+) -> SuiteResult:
```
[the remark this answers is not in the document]
kevinmessiaen: We use >=3.12

PR #2546, opened 2026-06-17
fix(checks): require comparison expected value

### LINKED ISSUE #2545: comparison checks allow omitted expected value

harsh21234i:
Closes #2545 

  ## Summary
  - distinguish omitted `expected_value` from explicit `expected_value=None`
  - reject comparison checks when both `expected_value` and `expected_value_key` are omitted
  - keep explicit `expected_value=None` valid
  - keep `expected_value_key=...` valid without `expected_value`
  - add shared validation coverage for comparison checks

  ## Testing
  - `.venv\Scripts\python.exe -m pytest -q libs/giskard-checks/tests/builtin/test_comparison.py libs/giskard-checks/
  tests/builtin/test_equals.py`
  - `uv run --no-sync ruff check libs/giskard-checks/src/giskard/checks/builtin/comparison.py libs/giskard-checks/tests/
  builtin/test_comparison.py`

  Verification: 103 passed, Ruff passed.

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
@@ -35,9 +35,9 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
     key: JSONPathStr = Field(
         ..., description="The key to extract the actual value from the trace"
     )
-    expected_value: ExpectedType | None = Field(
-        default=None,
-        description="The expected value to compare against. If None, the expected value is extracted from the trace using the expected_value_key.",
+    expected_value: ExpectedType | None | NotProvided = Field(
+        default=NOT_PROVIDED,
+        description="The expected value to compare against. If omitted, the expected value is extracted from the trace using the expected_value_key.",
     )
     expected_value_key: JSONPathStr | NotProvided = Field(
         default=NOT_PROVIDED,
--- a/libs/giskard-checks/tests/builtin/test_comparison.py
+++ b/libs/giskard-checks/tests/builtin/test_comparison.py
@@ -6,8 +6,10 @@ Tests cover different types (numbers, strings) and various comparison scenarios:
 - TypeError handling (missing methods and incompatible types)
 """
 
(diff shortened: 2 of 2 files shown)


harsh21234i:
hey @kevinmessiaen @davidberenstein1957  can you guys look into this !

davidberenstein1957 (approved):

davidberenstein1957:
@harsh21234i thanks for the contribution. Would you be able to fix the linting an such?

davidberenstein1957:
Closing in favor of #2504, which has active author engagement and an existing review cycle with @kevinmessiaen on the same `expected_value` / `NOT_PROVIDED` fix (#2545).

PR #2544, opened 2026-06-15
fix(scan): read JSONL datasets as UTF-8

weijun-xia:
## Description

`DatasetScenarioGenerator` opened the bundled `.jsonl` datasets without an explicit encoding:

```python
with path.open() as f:
```

On platforms whose default text encoding is not UTF-8 (e.g. GBK on Windows), loading a UTF-8 dataset such as `prompt_injection.jsonl` raises `UnicodeDecodeError`. This passes `encoding="utf-8"` so dataset loading is deterministic across platforms.

## Related Issue

_None._

## Type of Change

- [x] 🔧 Bug fix (non-breaking change which fixes an issue)

## Coding agents

This PR was prepared by an AI coding assistant with a human in the loop.

## Checklist

- [x] I've read the `CODE_OF_CONDUCT.md` document.
- [x] I've read the `CONTRIBUTING.md` guide.
- [x] I've written tests for all new methods and classes that I created (covered by existing `giskard-scan` dataset tests, which previously failed on non-UTF-8 default encodings).
- [x] I've written the docstring in NumPy format for all the methods and classes that I created or modified.
- [ ] I've updated the `uv.lock` running `uv lock` (N/A — `pyproject.toml` was not modified).

PR #2550, opened 2026-06-18
test(scan): cover UTF-8 decoding of JSONL datasets

weijun-xia:
## What
Add a regression test covering UTF-8 decoding of JSONL datasets in `DatasetScenarioGenerator`.

## Why
#2544 changed the dataset loader to read JSONL with `encoding="utf-8"`, but no existing test exercises non-ASCII content: the stubs build their JSON with `json.dumps(...)` (default `ensure_ascii=True`), so they only ever contain ASCII and would pass even if the loader fell back to the platform locale encoding.

## Change
`tests/generators/test_base.py`: add `test_dataset_generator_reads_utf8_content`, which writes a stub with non-ASCII scenario names (Latin-accented, CJK, Cyrillic) using `encoding="utf-8"` and asserts they round-trip through the generator. This locks in the UTF-8 read behaviour independently of the platform locale (on Windows the default `open()` codec is not UTF-8, so without #2544 this test would fail).

## Verification
- `pytest tests/generators/test_base.py` -> 7 passed
- `ruff check tests/generators/test_base.py` -> clean


libs/giskard-scan/tests/generators/test_base.py
```diff
@@ -128,3 +128,25 @@ async def test_dataset_generator_malformed_jsonl_raises_value_error(
     gen = _StubDatasetGenerator()
     with pytest.raises(ValueError, match=r"stub\.jsonl|line 2"):
         await gen.generate_scenario("desc", ["en"])
+
+
+async def test_dataset_generator_reads_utf8_content(tmp_path, monkeypatch):
+    """Non-ASCII scenario content is decoded correctly as UTF-8 regardless of locale."""
+    import json
+
+    import giskard.scan.generators.base as base_mod
```
kevinmessiaen: It's cleaner to have import at the top of the file

kevinmessiaen (changes requested):
Thanks, the test is a good addition but I'd move the import at top of the file

weijun-xia:
Thanks for the review! I moved the imports to the top of the file, and hoisted the existing in-function imports there too for consistency.

weijun-xia:
Thanks for applying the import hoist and merging main directly -- my local push raced with yours, so I've dropped my duplicate commit. Looks good on my end; happy to make any further tweaks if needed.

libs/giskard-scan/tests/generators/test_base.py
weijun-xia: haha ok

kevinmessiaen (approved):

PR #2552, opened 2026-06-19
feat(scan): add do-not-answer scan generator

kevinmessiaen:
### What

 Adds the **do-not-answer scan generator**: a vulnerability scenario generator backed by the `giskardai/do-not-answer-scenarios` HuggingFace dataset, wired into the vulnerability scan suite. Includes the machinery needed to run its fixed harmful prompts against agents with any input type.

 ### Pieces

 **1. `HuggingFaceDatasetScenarioGenerator` (`giskard-scan`)**
 Loads scenarios from a HF dataset repo using a per-language filename template (`donotanswer.{language}.jsonl`), discovers available languages by listing repo files, downloads matching files, and parses them into `Scenario`s. Registered in the vulnerability suite (commercial-use-gated). Refactors the shared JSONL parsing into `BaseDatasetScenarioGenerator._parse_scenarios` so the bundled and HF loaders stay identical.

 **2. `DatasetInputGenerator` (`giskard-checks`)**
 Lets a do-not-answer scenario's fixed harmful prompt reach agents whose input type is **not** a bare `str` (e.g. an email agent `{title, body}`).
 - **`str` target:** yield the prompt verbatim — no LLM.
 - **Structured target:** ask the LLM a **schema-only** question (the harmful prompt is **never** sent to the model) for a valid instance of the target type containing a literal `{{prompt}}` placeholder; cache per-schema; then substitute the real prompt **by code**. One LLM call per schema; raise on `schema_issue` or missing placeholder (no silent skip).
 - This avoids the "refusal trap": asking a model to render a *harmful* prompt into a schema can trigger a refusal in the generator, silently corrupting the test.

## Type of Change


libs/giskard-scan/src/giskard/scan/generators/base.py
```diff
@@ -84,37 +143,49 @@ async def generate_scenario(
...
+
+        if max_scenarios is not None and max_scenarios < len(scenarios):
+            rng = rng if rng is not None else np.random.default_rng()
+            indices = rng.choice(len(scenarios), size=max_scenarios, replace=False)
+            scenarios = [scenarios[i] for i in sorted(indices)]
+
+        return scenarios
+
+
+class DatasetScenarioGenerator(BaseDatasetScenarioGenerator):
```
pierlj: Is this still required if we use HF to get get the dataset? Even the prompt injection dataset (which contains only 1 sample for now) can be uploaded to HF
If you want to keep it for convenience, I suggest to rename into `LocalDatasetScenarioGenerator`

libs/giskard-checks/src/giskard/checks/generators/dataset.py
```diff
@@ -0,0 +1,171 @@
...
+    InputGenerator[TraceType], WithGeneratorMixin
+):
+    """Single-shot generator that places a fixed dataset prompt into the target input.
+
+    For a ``str`` target the prompt is yielded verbatim (no LLM). For a structured
+    target the prompt is injected into an LLM-resolved, schema-only template; the
+    LLM never sees the prompt itself.
+    """
+
+    prompt: str = Field(
```
pierlj, line 129: Not sure if it's worth supporting that but, sometimes the dataset could be used to fill a str message, not just sending the verbatim prompt from the dataset. This was the case for the encoding probes in Lidar. "Decode the following message: '{{ prompt }}'". Having an optional str template could convenient and instead of yielding `self.prompt`, yield the `substitute_prompt(str_template, self.prompt)`. 

Of course, we can support the same things by either include the template in the dataset samples in the first place, or even during the scenario loading.

libs/giskard-scan/src/giskard/scan/generators/huggingface.py
```diff
@@ -0,0 +1,96 @@
...
+    return available
+
+
+class HuggingFaceDatasetScenarioGenerator(BaseDatasetScenarioGenerator):
+    """Scenario generator backed by a Hugging Face dataset.
+
+    Loads scenarios from a Hugging Face dataset repository and annotates them
+    with the caller-supplied ``description`` and ``languages``.
+
+    The ``filename`` is a template containing a ``{language}`` placeholder
```
pierlj: This constraint prevents user to use HF dataset that do not follow this convention for the file naming (could not work well when you get multiple large dataset that may split each language in several files). I think it's better to load the subsets names of the dataset and enforce per language subset, not precise file naming. Then you can simply load the dataset with the default `load_dataset(<dataset-id>, "language")`.

pierlj (changes requested):
I would change how the HuggingFace Scenario Generator handles multiple languages (see my comments)

libs/giskard-checks/src/giskard/checks/generators/dataset.py
kevinmessiaen: I think for now we can just upload already built scenario

libs/giskard-scan/src/giskard/scan/generators/base.py
kevinmessiaen: Renamed to local. I would keep it so we can test on local dataset without having to upload to HF

PR #2559, opened 2026-06-24
feat(scan): add HarmBench dataset to vulnerability scan

kevinmessiaen:
Register `giskardai/harmbench-scenarios` with an `LLMJudge` prompt bundled in giskard-scan and document MIT attribution for commercial use.


pierlj (approved):
LGTM

PR #2575, opened 2026-06-29
fix(checks): resolve JsonValid string parse/value ambiguity

### LINKED ISSUE #2511: JsonValid rejects already parsed string values

kevinmessiaen:
## Problem

`JsonValid` answered two different questions on the same Python type, so plain
string values were wrongly rejected. A Python `str` is **both**:

- a candidate JSON document to parse (`'{"a":1}'`, `'42'`, `'"value"'`), and
- an already-valid JSON value — a JSON string (`"value"`).

The old code always tried to `json.loads` a string, so:

| Input | Schema | Old result |
|---|---|---|
| `'value'` | `{"type":"string"}` | FAIL (should pass — it's a string) |
| `'"value"'` | `{"type":"string"}` | PASS (only quoted form worked) |
| `'value'` | none | FAIL |
| `'true'` | none | PASS (→ `True`) |
| `'truth'` | none | FAIL (inconsistent with `'true'`) |

The ambiguity is structural — no heuristic can decide intent for a bare string.

## Fix

Make intent explicit with a `parse: bool = True` field. One rule per mode:

- **`parse=True`** (default): value must be a serialized JSON `str`, parsed via
  `json.loads`. Non-str → FAIL; bad JSON → FAIL.
- **`parse=False`**: value taken as-is (any type, incl. a plain string), checked
  with `json.dumps` for serializability.

| Input | `parse=True` | `parse=False` |
|---|---|---|
| `'value'` | FAIL | PASS (JSON string) |
| `'"value"'` | PASS → `"value"` | PASS |
| `42` (int) | FAIL (not str) | PASS |
| `{'a':1}` | FAIL (not str) | PASS |
| `{1,2,3}` (set) | FAIL | FAIL (not serializable) |

Also removes dead code (`_parse_json`, an unreachable `TypeError` handler) and
keeps the unresolvable schema `$ref` → `CheckResult.error` (ERROR) path.

## Type of Change

- [ ] 🔐 Security fix

## Coding agents

## Checklist

- [ ] I've written the docstring in NumPy format for all the methods and classes that I created or modified.
  modified)


davidberenstein1957 (approved):
LFTM

PR #2595, opened 2026-07-07
fix(google): map Gemini finish_reason via enum value, not str()

WatchTree-19:
### problem
in the Gemini -> giskard translator (`llm/translators/google_chat.py`), the finish reason is looked up with `str(candidate.finish_reason)`:

```python
finish_reason = FINISH_REASON_MAP.get(str(candidate.finish_reason), "stop")
```

but `candidate.finish_reason` is a `google.genai` `FinishReason` enum, and `str(FinishReason.MAX_TOKENS)` is `"FinishReason.MAX_TOKENS"`, not `"MAX_TOKENS"`. `FINISH_REASON_MAP` is keyed on the bare names (`"MAX_TOKENS"`, `"SAFETY"`, ...), so every non-STOP reason misses the map and falls back to `"stop"`. `STOP` only "works" by accident because the fallback is also `"stop"`.

net effect: Gemini truncations are reported as normal completions (`length` lost) and, more importantly for a safety tool, **SAFETY blocks / refusals are never surfaced as refusals**.

### repro (real google-genai SDK)
```python
from google.genai.types import FinishReason
str(FinishReason.MAX_TOKENS)          # 'FinishReason.MAX_TOKENS'  (not 'MAX_TOKENS')
FINISH_REASON_MAP.get(str(FinishReason.MAX_TOKENS), "stop")  # 'stop'  -> should be 'length'
```

### fix
look up via the enum's `.value` (`getattr(finish_reason, "value", finish_reason)`, robust to both the real enum and the plain-string form used in dict-validated inputs). also fixed the `refusal_out` line so refusal text is a clean reason string rather than `"FinishReason.SAFETY"`.

### test
added `test_from_google_max_tokens_maps_to_length` and `test_from_google_safety_maps_to_refusal` mirroring the file's existing tests. before: `2 failed` (`assert 'stop' == 'length'`, `assert 'stop' == 'refusal'`); after: pass. the existing translator tests only covered `STOP`, which is indistinguishable from the fallback, so the bug went unnoticed. ruff clean.

found by reading the translator.

### DIFF
--- a/libs/giskard-llm/src/giskard/llm/translators/google_chat.py
+++ b/libs/giskard-llm/src/giskard/llm/translators/google_chat.py
@@ -284,7 +284,6 @@ REFUSAL_REASONS = frozenset(
 FINISH_REASON_MAP = {
     "STOP": "stop",
     "MAX_TOKENS": "length",
-    "SAFETY": "content_filter",
 } | {reason: "refusal" for reason in REFUSAL_REASONS}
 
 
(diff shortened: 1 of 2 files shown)


kevinmessiaen (approved):
Thanks for the fix @WatchTree-19

PR #2599, opened 2026-07-09
fix(checks): record input generation errors

### LINKED ISSUE #2523: Handle input-generation errors like check errors

harsh21234i:
Closes #2523

  ## Summary
  - Add `TestCaseError` to capture execution errors that happen before checks run
  - Record input-generation failures as errored `TestCaseResult`s when `return_exception=True`
  - Keep `return_exception=False` behavior unchanged so input-generation errors still raise
  - Mark checks in the failed generation step as skipped
  - Add coverage for scenario-level recorded errors and suite continuation after one failing scenario

  ## Testing
  - `python -m py_compile libs/giskard-checks/src/giskard/checks/core/result.py libs/giskard-checks/src/giskard/checks/
  scenarios/runner.py libs/giskard-checks/src/giskard/checks/core/__init__.py libs/giskard-checks/src/giskard/checks/
  __init__.py libs/giskard-checks/tests/core/test_scenario.py libs/giskard-checks/tests/core/test_suite.py`
  - `git diff --check`


harsh21234i:
hey @henchaves @kevinmessiaen   can you look into this?

henchaves (approved):

PR #2603, opened 2026-07-14
fix(core): registry loses live rate limiters via WeakSet structural dedup 🤖🤖🤖🤖

chuenchen309:
## Problem

`RateLimiterRegistry` tracks live `BaseRateLimiter` instances per `id` in a `WeakSet`. `BaseRateLimiter.__eq__` is intentionally overridden to compare by model fields (not identity), so the registry can find an existing matching instance to share state with — this is used when duplicate-id warnings are downgraded via `GISKARD_DISABLE_DUPLICATE_RATE_LIMITERS_WARNINGS=1`.

But a `WeakSet` is backed by a plain `set`, and `set.add()` on an element that already compares equal (`__eq__`/`__hash__`) to a member already in the set is a no-op — it does **not** add a second entry. So when two structurally-equal instances share an `id` (allowed under the downgraded-warning mode), only the *first* one actually ends up stored in the set. If that first instance is later garbage collected, its weakref removal callback empties the tracking entry entirely — even though a second, still-alive, structurally-equal instance exists. `BaseRateLimiter.from_id()` then wrongly raises `"not found"`.

## Fix

Track instances in a `WeakValueDictionary` keyed by `id(rate_limiter)` (object identity) instead of a `WeakSet` (structural equality via `__eq__`/`__hash__`), so each live object gets its own slot regardless of whether another object compares equal to it.

## Testing

- Added `test_from_id_finds_second_equal_instance_after_first_is_deleted`: creates two structurally-equal rate limiters sharing an id (under the downgraded-warning mode), deletes the first, and asserts the second is still reachable via `from_id`.
- Confirmed red→green: reverting only `base.py` reproduces `ValueError: ... not found`; reapplying the fix passes.
- Full `libs/giskard-core/tests/` suite: 38 passed.
- `ruff check`, `ruff format --check`, `basedpyright`, `vermin` all clean (0 errors on the changed file; pre-existing unrelated warnings elsewhere untouched).
- `pre-commit run --files` (trailing-whitespace, EOF, large-files, merge-conflicts, case-conflicts, mixed-line-ending, ruff check/format, basedpyright, detect-secrets) all pass.

## AI-Generated disclosure

Found via an AI-assisted code review pass (Claude Code) over `giskard-core/src/giskard/core/rate_limiter/`. I personally reproduced the bug with a minimal repro, verified the fix resolves it, and ran the full test suite plus all lint/type/security checks before submitting.

### DIFF
--- a/libs/giskard-core/src/giskard/core/rate_limiter/base.py
+++ b/libs/giskard-core/src/giskard/core/rate_limiter/base.py
@@ -12,7 +12,7 @@ from abc import ABC, abstractmethod
 from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
 from typing import Any, ClassVar, Self, override
-from weakref import WeakSet
+from weakref import WeakValueDictionary
 
 from pydantic import (
     ConfigDict,
@@ -34,7 +34,7 @@ class RateLimiterRegistry:
     throttling is consistent across serialization boundaries.
     """
 
-    _instances: dict[str, WeakSet["BaseRateLimiter[Any]"]]
+    _instances: dict[str, WeakValueDictionary[int, "BaseRateLimiter[Any]"]]
 
     def __init__(self):
         self._instances = {}
@@ -52,10 +52,10 @@ class RateLimiterRegistry:
         """
         instances = self._instances.get(rate_limiter.id)
         if instances is None:
-            instances = WeakSet["BaseRateLimiter"]()
+            instances = WeakValueDictionary[int, "BaseRateLimiter"]()
             self._instances[rate_limiter.id] = instances
 
-        all_instances = list(instances)
+        all_instances = list(instances.values())
         matching_instances = [
             instance for instance in all_instances if instance == rate_limiter
         ]
@@ -79,7 +79,7 @@ class RateLimiterRegistry:
                 RuntimeWarning,
             )
 
-        instances.add(rate_limiter)
+        instances[id(rate_limiter)] = rate_limiter
 
     def get_instance(self, id: str) -> "BaseRateLimiter[Any]":
         """Retrieve a registered rate limiter by id.
@@ -103,7 +103,7 @@ class RateLimiterRegistry:
         if not instances:
             raise ValueError(f"Rate limiter with id '{id}' not found")
 
-        return next(iter(instances))
+        return next(iter(instances.values()))
 
 
 @discriminated_base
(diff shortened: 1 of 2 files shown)


chuenchen309:
Thanks @gemini-code-assist — good catch. Pushed a follow-up commit adding a `weakref.finalize`-based `_cleanup_id` to drop the `id` entry from `self._instances` once its `WeakValueDictionary` is empty, so this doesn't leak one dict entry per ephemeral (default-random-UUID) rate limiter over a long-running process.

One correction to the suggested diff: registering the finalizer *after* inserting into the `WeakValueDictionary` races the dict's own internal weak-value cleanup — weakref callbacks fire in reverse registration order (confirmed via a standalone repro), so `_cleanup_id` would run *before* the dict removed its own dead entry and would see a stale non-empty dict, never actually cleaning up. Registering the finalizer first (before the insert) fixes the ordering. Added a regression test and ran it 10x to confirm no flakiness from the ordering.

ebarkhordar (commented):
The WeakSet structural-dedup root cause is exactly right. I checked out `ed5f84f` and ran `tests/test_rate_limiter.py` (21 passed), and reverting only `base.py` reproduces the `not found` failure, so the identity-keyed `WeakValueDictionary` clearly resolves it.

One optional note on `_cleanup_id`, not blocking, since it works as written on CPython. The empty-map cleanup depends on the finalizer firing after the dictionary's own `KeyedRef` removal, which is why you register the finalizer before the insert. That ordering holds on CPython today, but the reference docs call the callback order unspecified, and the failure mode is silent: a leaked empty entry per id, no error. I confirmed the dependency is real by swapping just those two lines, after which `test_registry_drops_tracking_entry_once_all_instances_are_collected` fails while the identity test still passes.

If you want to drop that dependency, checking the live values makes the cleanup order-independent:

```python
if instances is not None and not list(instances.values()):
    self._instances.pop(id, None)
```

At finalizer time the collected object's weakref already returns `None`, so `.values()` reports the map as empty whether or not its `KeyedRef` entry has been pruned yet, whereas `not instances` counts the raw entry and can still see the dead one. With that change the leak test passes even with the insert and finalize lines reversed (21 passed either way).

davidberenstein1957 (approved):

davidberenstein1957:
Opened a formatted follow-up so CI `check-format` can pass: https://github.com/Giskard-AI/giskard-oss/pull/2649

Same change as this PR, plus the markdown/ruff format fix. Please use #2649 going forward — we can close this one once that is green.

PR #2608, opened 2026-07-16
fix(llm): use .text not .content in _validate_messages to avoid AttributeError on list content

chuenchen309:
## What

`_validate_messages` in the OpenAI, Anthropic, and Google providers raises `AttributeError: 'list' object has no attribute 'strip'` when a system/developer message carries **list-typed content** instead of a plain string.

Each provider guards empty instruction messages with:

```python
if m.role in _INSTRUCTION_ROLES and not (m.content or "").strip():
    raise BadRequestError(...)
```

But `SystemMessage.content` / `DeveloperMessage.content` is typed `str | Sequence[TextContent]`. When it's a non-empty list (a legitimate, reachable input via plain-dict validation), `(m.content or "")` returns the list and `.strip()` blows up with `AttributeError` — instead of the intended `BadRequestError` (or passing validation for non-empty content).

## Fix

Use the `.text` property already defined on these message models (built for exactly this normalization) instead of the raw `.content`:

```python
if m.role in _INSTRUCTION_ROLES and not (m.text or "").strip():
```

Three-line change, one per provider (`openai.py`, `anthropic.py`, `google.py`).

## Testing

Added a regression test per provider in `libs/giskard-llm/tests/test_providers.py` — a developer/system-role message with list-typed content. All three reproduce the identical `AttributeError` before the fix (TDD red confirmed by reverting the 3 source files) and pass after. Full `libs/giskard-llm/tests/test_providers.py` suite: 47 passed. `ruff check` + `ruff format --check` clean.

xref: no open PR or issue touches these three files or `_validate_messages`.

---
Disclosure: this PR was drafted with AI assistance (Claude); I reviewed, tested, and take responsibility for the change.

### DIFF
--- a/libs/giskard-llm/src/giskard/llm/providers/anthropic.py
+++ b/libs/giskard-llm/src/giskard/llm/providers/anthropic.py
@@ -224,7 +224,7 @@ class AnthropicProvider:
                 raise BadRequestError(
                     400, "Tool messages must have a tool_call_id.", PROVIDER
                 )
-            if m.role in _ANTHROPIC_INSTRUCTION_ROLES and not (m.content or "").strip():
+            if m.role in _ANTHROPIC_INSTRUCTION_ROLES and not (m.text or "").strip():
                 raise BadRequestError(
                     400,
                     "System and developer messages must have non-empty content.",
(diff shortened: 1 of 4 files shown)


kevinmessiaen (approved):
Thanks for the fix @chuenchen309

PR #2611, opened 2026-07-16
fix(google): convert Pydantic response_format to the Interactions SDK TextResponseFormatParam 🤖🤖🤖🤖

chuenchen309:
### Problem

In `GoogleResponseTranslator._coerce_dict` (`google_response.py`), when a user passes a Pydantic model as `response_format` for structured output, the code sets `response_mime_type = "application/json"` but leaves the **raw Python class** in the `response_format` field:

```python
if (
    "response_format" in v
    and isinstance(v["response_format"], type)
    and issubclass(v["response_format"], BaseModel)
):
    v["response_mime_type"] = "application/json"
```

That payload is then sent to `self._client.aio.interactions.create(**kwargs)` (`providers/google.py:376`). The google-genai Interactions SDK expects `response_format` to be a `TextResponseFormatParam` dict:

```python
class TextResponseFormatParam(TypedDict, total=False):
    type: Required[Literal["text"]]
    mime_type: Literal["application/json", "text/plain"]
    schema: Dict[str, object]  # JSON schema the output should conform to
```

A raw class is not JSON-serializable and doesn't match that shape, so the JSON schema the caller requested is **never sent to the API** — structured output is silently dropped (or the request fails on serialization).

The sibling translators already do this correctly: `openai_chat.py` and `anthropic.py` both call `.model_json_schema()` on the Pydantic class.

### Fix

Convert the Pydantic class into the SDK's `TextResponseFormatParam` dict, mirroring the other providers:

```python
v["response_mime_type"] = "application/json"
v["response_format"] = {
    "type": "text",
    "mime_type": "application/json",
    "schema": v["response_format"].model_json_schema(),
}
```

### Tests

Added `test_response_format_pydantic_class_becomes_text_response_format` asserting the payload's `response_format` is the `{type, mime_type, schema}` dict rather than the raw class. TDD red confirmed against the unfixed code (payload held the raw class); green after the fix. Full `libs/giskard-llm` unit suite: 212 passed / 7 skipped, no regressions. `ruff check`/`ruff format`/`basedpyright` clean (pre-commit hooks pass).

---
Disclosure: I use Claude Code as an assistant; every change here was reproduced with a failing-then-passing test and reviewed against the real SDK types before submitting. Happy to adjust the shape if you'd prefer a different representation.


davidberenstein1957 (approved):

davidberenstein1957:
Opened a formatted follow-up so CI `check-format` can pass: https://github.com/Giskard-AI/giskard-oss/pull/2648

Same change as this PR, plus the markdown/ruff format fix. Please use #2648 going forward — we can close this one once that is green.

PR #2614, opened 2026-07-17
fix(llm): warn on unknown Anthropic completion params 🤖🤖🤖🤖

### LINKED ISSUE #2613: Anthropic completion params are silently dropped (the OpenAI translator warns, Anthropic does not)
## What happens

`AnthropicProvider.complete(..., **params)` silently discards every completion param it does not declare. No error, no warning: the request just goes to the API without them.

The sibling OpenAI translator handles the identical situation by telling the caller (`openai_chat.py`):

```python
unknown = set(params) - KNOWN_COMPLETION_PARAMS
if unknown:
    logger.warning("%s provider: ignoring unknown completion params: %s", PROVIDER, sorted(unknown))
```

`to_anthropic()` instead funnels params through `AnthropicChatConfigParams`, a pydantic model whose declared fields are `model, messages, max_tokens, tools, system, temperature, timeout, output_config`. With pydantic's default `extra="ignore"`, anything else is dropped without a sound.

Both provider `__init__` methods already warn on unknown kwargs, so the convention in this repo is "unsupported input is ignored, but the caller is told". The Anthropic completion path is the one place that stays silent.

## Reproduction

Verified on `175670e`, in a clean `python:3.12-slim` container, against the repo source (`__file__` printed to confirm provenance):

```python
from collections.abc import Sequence
from pydantic import TypeAdapter
from giskard.llm.translators.anthropic import AnthropicChatTranslator
from giskard.llm.types import ChatMessage

msgs = TypeAdapter(Sequence[ChatMessage]).validate_python([{"role": "user", "content": "hi"}])
out = AnthropicChatTranslator.to_anthropic(
    "claude-sonnet-4-5", msgs,
    top_p=0.5, top_k=40, stop_sequences=["STOP"],
)
print(sorted(out.keys()))
```

Output:

```
PROVENANCE: /src/libs/giskard-llm/src/giskard/llm/translators/anthropic.py
keys actually sent to the Anthropic SDK: ['max_tokens', 'messages', 'model']
  top_p           survived? False
  top_k           survived? False
  stop_sequences  survived? False
```

## Why it matters

Giskard exists to evaluate LLM behaviour, so sampling settings are part of the experiment. A user who sets `top_p` or `stop_sequences` on an Anthropic model gets an eval run that silently used different settings than they configured, and nothing in the logs says so. The failure is invisible at exactly the moment it changes results.

## Suggested fix

Mirror the OpenAI translator: compare the incoming params against the known set and `logger.warning` the unknown ones before dropping them. That keeps the current whitelist scope and only removes the silence, so it is a small change and consistent with the existing pattern.

Happy to open a PR for that if you want it. If you would rather the Anthropic path actually forward `top_p`, `top_k` and `stop_sequences` to the SDK instead of only warning, that is a bigger call and I would rather you pick the direction first.

Filed by an autonomous agent (per AUTONOMOUS.md); the repro above was executed, not inferred.

feiiiiii5:
## Description

`AnthropicChatTranslator.to_anthropic(**params)` silently dropped any caller-supplied key that `AnthropicChatConfigParams` does not declare. Because `_BaseModel` does not override pydantic's default `extra="ignore"`, unknown fields were discarded with no signal — so callers passing sampling settings like `top_p`, `top_k`, or `stop_sequences` had no way to learn that those values never reached the Anthropic SDK.

This mirrors the OpenAI translator's pattern (`libs/giskard-llm/src/giskard/llm/translators/openai_chat.py`): compare incoming `params` against an explicit allowlist (`KNOWN_COMPLETION_PARAMS`) and `logger.warning` the unknown ones before constructing the pydantic model. Behavior is unchanged for known params; only the silence is removed.

The allowlist is the set of kwargs that `to_anthropic()` forwards to `AnthropicChatConfigParams` other than the explicit `model` / `messages` / `tools` parameters: `max_tokens`, `temperature`, `timeout`, `output_config`, `response_format`, `system`. (`system` is auto-extracted from messages by `_coerce_dict`; `response_format` is rewritten to `output_config`. Both are accepted to avoid false-positive warnings on legitimate callers.)

## Related Issue

Fixes #2613

## Type of Change

## Coding agents

Autonomous-agent PR per `AUTONOMOUS.md`. PR title ends with `🤖🤖🤖🤖`.

## Checklist

- [x] I've updated the `uv.lock` running `uv lock` (only applicable when `pyproject.toml` has been modified)
  — N/A: no `pyproject.toml` change in this PR.

## Approach

`libs/giskard-llm/src/giskard/llm/translators/anthropic.py`:

1. Added `import logging` and `logger = logging.getLogger(__name__)`.
2. Added `PROVIDER = "anthropic"` constant (matches OpenAI translator's `PROVIDER = "openai"`).
3. Added `KNOWN_COMPLETION_PARAMS` frozenset listing the kwargs `to_anthropic()` accepts (everything except the explicit `model` / `messages` / `tools` parameters).
4. In `to_anthropic()`, before constructing `AnthropicChatConfigParams`, compute `unknown = set(params) - KNOWN_COMPLETION_PARAMS` and `logger.warning("anthropic provider: ignoring unknown completion params: %s", sorted(unknown))` if non-empty.

Diff is 3 surgical edits on `anthropic.py` + 3 new tests in `test_anthropic_chat.py`. No existing lines were refactored.

## Alternatives considered

- **Forward unknown params to the SDK** (i.e. construct `AnthropicChatConfigParams` with `extra="allow"` or pass them through directly). The issue author explicitly asked the maintainer to pick between this and the warn-only approach; per `AUTONOMOUS.md`, agents should not unilaterally make that larger behavior change. This PR implements the recommended minimal direction.
- **`model_config = ConfigDict(extra="forbid")`** on `AnthropicChatConfigParams`. Would convert silent drops into hard `ValidationError`s, breaking every existing caller that passes an unknown kwarg (including internal call sites). Too large a behavior change for this issue.
- **Generic `**_extra` catch-all** added to `AnthropicChatConfigParams`. Adds complexity and changes the model contract. The OpenAI translator's allowlist+warning is the existing pattern, so we mirror it.

## Testing

`libs/giskard-llm/tests/translators/test_anthropic_chat.py` — three new regression tests covering the bug class:

- `test_unknown_completion_params_warn` — passing `top_p=0.5, top_k=40, stop_sequences=["STOP"]` (all outside `KNOWN_COMPLETION_PARAMS`) emits exactly one `WARNING` record on the `giskard.llm.translators.anthropic` logger, and the message contains `"anthropic"` and each of the dropped param names.
- `test_known_completion_params_do_not_warn` — passing `temperature=0.7, max_tokens=100` (both inside `KNOWN_COMPLETION_PARAMS`) does not emit any `WARNING` record, and the values reach the resulting payload unchanged.
- `test_unknown_completion_params_sorted_in_warning` — dropped param names are listed in sorted order in the warning message (`"['alpha', 'zebra']"`), matching the OpenAI translator's output format so logs are consistent across providers.

Verified locally against a venv with `giskard-llm` installed: all three new tests pass against the patched `anthropic.py`, and the existing test suite for the file continues to pass.

```
test_unknown_completion_params_warn PASSED [ 33%]
test_known_completion_params_do_not_warn PASSED [ 66%]
test_unknown_completion_params_sorted_in_warning PASSED [100%]
3 passed in 0.29s
```


feiiiiii5:
The `authorize` check is failing because external contributors require a maintainer to add the `safe for build` label before CI jobs run (this is a deliberate security gate, not a code issue).

Gemini Code Assist has already reviewed the diff and reported no concerns ("There are no review comments to address, and the changes look solid"). Could a maintainer add the `safe for build` label so the rest of the integration matrix can run? Happy to address anything else that comes up.

feiiiiii5:
Following up here so it doesn't get lost in the issue thread: @ebarkhordar approved the mirror-OpenAI warning direction in #2613.

The `Integration Tests` check is currently showing as failed, but it's actually the `authorize` step early-exiting with `External contributors require a maintainer to add the 'safe for build' label.` — the matrix jobs (`agents`/`checks`/`llm`) all show `0s` and haven't run yet. Once the `safe for build` label is added, the real test suite will execute against this PR.

No code changes needed — just the label to unblock CI.

feiiiiii5:
## CI requires maintainer action

The `authorize` check is failing because external contributors require a maintainer to add the `safe for build` label before CI workflows (`agents`, `checks`, `llm`) will run on this PR:

```
External contributors require a maintainer to add the 'safe for build' label.
```

Could a maintainer add the `safe for build` label so the test suite can run? I am happy to address any review feedback once CI is unblocked. Thanks!

davidberenstein1957:
closing in favor of https://github.com/Giskard-AI/giskard-oss/pull/2623

davidberenstein1957:
Closing in favor of https://github.com/Giskard-AI/giskard-oss/pull/2623, which covers the unknown-params warning for https://github.com/Giskard-AI/giskard-oss/issues/2613 and also bare Anthropic refusals for https://github.com/Giskard-AI/giskard-oss/issues/2615 in one change.

PR #2621, opened 2026-07-18
Add MiniMax provider routing

octo-patch:
Reason: Add MiniMax provider support so the LLM provider registry can route MiniMax-M3 and MiniMax-M2.7 models.

Added a MiniMax registry alias backed by the OpenAI-compatible provider implementation.

Checks: pytest libs/giskard-llm/tests/test_routing.py -q


davidberenstein1957 (approved):
MiniMax provider routing looks good.

PR #2628, opened 2026-07-22
fix(checks): skip regorus extra on Windows

### LINKED ISSUE #2627: Windows dev/test setup fails because celine-regorus has no win_amd64 wheel
## Problem

  On Windows, running project/test commands with `uv run` can fail before tests start because `celine-regorus` has no
  Windows wheel.

  This blocks local development and targeted test runs even when the test being run is unrelated to Rego checks.

  ## Reproduction

  On Windows:

  ```powershell
  uv run -m pytest -q libs/giskard-checks/tests/core/test_suite.py

  Observed error:

  error: Distribution `celine-regorus==0.9.1.post20260227075446` can't be installed because it doesn't have a source
  distribution or wheel for the current platform

  You're on Windows (`win_amd64`), but `celine-regorus` only has wheels for:
  manylinux_2_34_x86_64, macosx_10_12_x86_64, macosx_11_0_arm64

  ## Expected behavior

  Windows contributors should be able to run unrelated checks/tests without being blocked by an optional Rego
  dependency.

  ## Possible fixes

  - Add platform markers around the regorus extra.
  - Exclude celine-regorus from default/all dev sync on Windows.
  - Add tool.uv.required-environments so the lock does not resolve unsupported wheel-only packages for Windows.

  ## Context

  giskard-checks declares:

  regorus = ["celine-regorus>=0.9.1"]
  all = ["giskard-checks[regorus]"]

  Since celine-regorus has no Windows wheel/source distribution, this breaks local uv run flows on Windows.

harsh21234i:
Closses #2627

 ## Summary
  - Guard `celine-regorus` with `sys_platform != "win32"` because it has no Windows wheel/source distribution.
  - Update `uv.lock` so Windows users can resolve/export the `regorus` extra without install failures.
  - Keep `regorus` available on supported Linux/macOS platforms.

  ## Testing
  - `uv lock`
  - `uv export --no-dev --extra regorus --format requirements-txt`
  - `ruff format --check libs/giskard-checks/pyproject.toml`
  - `ruff check libs/giskard-checks/pyproject.toml`
  - `git diff --check`

### DIFF
--- a/libs/giskard-checks/pyproject.toml
+++ b/libs/giskard-checks/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
 readability = [
     "textstat>=0.7.0",
 ]
-regorus = ["celine-regorus>=0.9.1"]
+regorus = ["celine-regorus>=0.9.1; sys_platform != 'win32'"]
 all = ["giskard-checks[regorus]"] # Install all optional dependencies
 
 [tool.pytest.ini_options]
--- a/uv.lock
+++ b/uv.lock
@@ -967,18 +967,18 @@ dependencies = [
 
 [package.optional-dependencies]
 all = [
-    { name = "celine-regorus" },
+    { name = "celine-regorus", marker = "sys_platform != 'win32'" },
 ]
 readability = [
     { name = "textstat" },
 ]
 regorus = [
-    { name = "celine-regorus" },
+    { name = "celine-regorus", marker = "sys_platform != 'win32'" },
 ]
 
 [package.metadata]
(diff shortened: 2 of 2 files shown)


davidberenstein1957:
Opened a formatted follow-up so CI `check-format` can pass: https://github.com/Giskard-AI/giskard-oss/pull/2646

Same change as this PR, plus the markdown/ruff format fix. Please use #2646 going forward — we can close this one once that is green.

kevinmessiaen:
Merged in #2646

PR #2483, opened 2026-05-21
  feat(checks): add consensus retries for LLM-based checks

harsh21234i:
Closes #2372

  ## Summary
  - add `num_runs` and `consensus` to `BaseLLMCheck`
  - support repeated LLM judge execution with `majority`, `unanimous`, and `any` consensus strategies
  - preserve individual run results in `details["runs"]` when multiple runs are used
  - include aggregated consensus metadata in result details
  - keep the default single-run behavior unchanged

  ## Testing
  - `uv run -m pytest -q libs/giskard-checks/tests/builtin/test_base.py libs/giskard-checks/tests/builtin/test_judge.py libs/giskard-checks/tests/builtin/
  test_groundedness.py libs/giskard-checks/tests/builtin/test_conformity.py`
  - `uv run -m pytest -q libs/giskard-checks/tests/builtin`
  - `uv run ruff check libs/giskard-checks/src/giskard/checks/judges/base.py libs/giskard-checks/tests/builtin/test_base.py`

PR #2631, opened 2026-07-23
feat(checks): add retry/consensus voting to BaseLLMCheck

### LINKED ISSUE #2372: Add retry/majority-voting for LLM-based checks
## Summary

LLM-based checks (LLMJudge, Groundedness, Conformity, etc.) are inherently non-deterministic. Add configurable retry with majority voting to reduce flakiness in CI pipelines.

## Motivation

A single LLM judge call can give different results for the same input. This causes flaky tests. Majority voting (run N times, take the majority result) significantly improves reliability. DeepEval addresses this with threshold tuning.

### Steps

1. Add retry parameters to `BaseLLMCheck`:
   * `num_runs: int = 1` — number of times to run the judge
   * `consensus: Literal["majority", "unanimous", "any"] = "majority"` — voting strategy
2. Implement in `BaseLLMCheck.run()`:
   * Run the check `num_runs` times
   * Apply consensus strategy to determine final result
   * Include individual run results in the check result details
3. Add tests

### Consensus strategies

* **majority**: Pass if >50% of runs pass
* **unanimous**: Pass only if all runs pass (strictest)
* **any**: Pass if at least one run passes (most lenient)

### Example usage

```python
from giskard.checks import Groundedness

# Run 3 times, pass if majority agrees
Groundedness(
    context="...",
    num_runs=3,
    consensus="majority"
)
```

## Acceptance Criteria

- [ ] Configurable number of runs
- [ ] Three consensus strategies: majority, unanimous, any
- [ ] Individual run results are preserved in details
- [ ] Default behavior (num_runs=1) is unchanged
- [ ] Tests cover: all agree pass, all agree fail, split decision, each consensus mode

TheManishCode:
Closes #2372

## Problem

`BaseLLMCheck.run()` calls the underlying `ChatWorkflow` exactly once and returns that single result. LLM judges can be noisy on a single call, and there was no way to reduce that noise via repeated sampling.

## Fix

Adds two fields to `BaseLLMCheck`:
- `num_runs: int = 1` (validated `>= 1`)
- `consensus: Literal["majority", "unanimous", "any"] = "majority"`

The original single-pass body of `run()` is extracted into `_run_once()` with byte-identical logic. When `num_runs == 1`, `run()` calls `_run_once()` directly — same code path, same `details` shape, zero behavior change and zero added overhead for existing callers. When `num_runs > 1`, it runs `_run_once()` that many times, tallies pass/fail, and applies the consensus rule:
- `majority`: more than half the runs pass
- `unanimous`: all runs pass
- `any`: at least one run passes

The final `CheckResult.details` includes `num_runs`, `consensus`, `passed_count`, and `runs` (the individual per-run results) for observability.

Note: there's an existing PR (#2483) attempting this same issue, but it's been stale since June with failing CI and no response to review — this is an independent implementation, not based on that branch.

## Tests

Added to `libs/giskard-checks/tests/builtin/test_judge.py`, including a `SequenceMockGenerator` for split-decision scenarios:
- default behavior (`num_runs=1`) unchanged, no `"runs"` key in details
- all-pass / all-fail across multiple runs
- split-decision case for each of the three consensus modes
- `num_runs=0` raises a validation error

```
pytest libs/giskard-checks/tests/builtin/test_judge.py -m "not functional" -q   # 13 passed
pytest libs/giskard-checks -m "not functional" -q                              # 749 passed, 16 skipped
ruff check / ruff format --check                                               # clean
basedpyright --level error                                                     # 0 errors
```

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/judges/base.py
+++ b/libs/giskard-checks/src/giskard/checks/judges/base.py
@@ -1,4 +1,4 @@
-from typing import Any, override
+from typing import Any, Literal, override
 
 from giskard.agents import ChatWorkflow, MessageTemplate, TemplateReference
 from giskard.llm.types import ChatMessage
@@ -33,7 +33,31 @@ class BaseLLMCheck[InputType, OutputType, TraceType: Trace](  # pyright: ignore[
     generator : BaseGenerator
         Generator for LLM evaluation. Defaults to the global
         default generator if not specified.
+    num_runs : int
+        Number of times to run the LLM evaluation. Defaults to 1 (single run,
+        identical to historical behavior). When greater than 1, the check runs
+        `num_runs` times and combines the individual results via `consensus`.
+    consensus : Literal["majority", "unanimous", "any"]
+        Rule used to combine multiple run results into a final pass/fail when
+        `num_runs > 1`. Ignored when `num_runs == 1`.
     """
+
+    num_runs: int = Field(
+        default=1,
+        ge=1,
+        description=(
+            "Number of times to run the LLM evaluation. When > 1, results are "
+            "combined using `consensus`."
+        ),
+    )
+    consensus: Literal["majority", "unanimous", "any"] = Field(
+        default="majority",
+        description=(
+            "How to combine multiple run results when `num_runs > 1`: "
+            "'majority' (more than half pass), 'unanimous' (all pass), or "
+            "'any' (at least one passes)."
+        ),
+    )
 
     @property
     def output_type(self) -> type[BaseModel] | None:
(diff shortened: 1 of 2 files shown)


davidberenstein1957 (approved):

davidberenstein1957:
Opened a formatted follow-up so CI can pass: https://github.com/Giskard-AI/giskard-oss/pull/2651

Same change as this PR, plus ruff format/lint fixes. Please use #2651 going forward — we can close this one once that is green.

davidberenstein1957:
Hey @TheManishCode — thanks a lot for the time and effort you put into this. Really appreciate you jumping in and iterating on feedback.

We're closing this PR because it conflicts with the current roadmap for the v3 release, so we can't take it forward as-is right now. Sorry for the long wait and the late call.

If something in this direction comes back on the roadmap later, we'll reopen the conversation. Thanks again.

PR #2634, opened 2026-07-26
fix(checks): descendants JSONPath returns scalar vs list depending on match count

### LINKED ISSUE #2633: JSONPath descendants (`..`) resolves to a scalar or a list depending on the data
## Summary

`resolve()` returns a bare scalar for a `..` (descendants) JSONPath when the data happens to hold exactly one match, and a list when it holds two or more. The expression is identical; only the trace differs.

This is the same defect #2605 fixed for dot-wildcards and multi-field selectors. That fix added a `Fields` branch to `_is_list_expression`, but `..` compiles to `jsonpath_ng.Descendants`, which is still handled by the recursive `Child | Descendants` branch:

```python
if isinstance(expression, Child | Descendants):
    return _is_list_expression(expression.right) or _is_list_expression(expression.left)
```

For `trace..ctx` both children are single-field `Fields`, so this returns `False` and `resolve()` falls back to its match-count heuristic (`len(matches) > 1 or _is_list_expression(expression)`).

`..` is a supported expression — `tests/core/test_extraction.py` asserts `trace..outputs` passes validation.

## Reproduction

```python
import asyncio
from giskard.checks import Equals

def trace(*metadata):
    return Trace[str, str](interactions=[
        Interaction(inputs=f"q{i}", outputs=f"a{i}", metadata=m)
        for i, m in enumerate(metadata)
    ])

async def main():
    check = Equals(key="trace..ctx", expected_value="doc-1", match="any")
    for label, t in [
        ("one interaction carries ctx ", trace({"ctx": "doc-1"}, {})),
        ("two interactions carry ctx  ", trace({"ctx": "doc-1"}, {"ctx": "doc-2"}, {})),
    ]:
        r = await check.run(t)
        print(f"{label}: {'PASS' if r.passed else 'FAIL'} | {r.message}")

asyncio.run(main())
```

```
one interaction carries ctx : FAIL | Expected a list, set, or tuple at key 'trace..ctx' when match is 'any', but got str.
two interactions carry ctx  : PASS | At least one value in ['doc-1', 'doc-2'] is equal to 'doc-1'.
```

The check is configured correctly in both runs. Whether it passes depends on how many interactions happen to carry the field, which is exactly the failure mode described in #2605.

## Expected

`trace..ctx` always resolves to a list, the way `trace.last.metadata.*` and bracket wildcards already do.

## Suggested fix

Give `Descendants` its own branch returning `True`, and leave `Child` recursing. `..` searches a whole subtree, so its match count is a property of the data rather than of the expression.

One consequence worth naming: `trace..missing` would then return `[]` instead of `NoMatch`. That matches how list-expressions already behave — `trace.last.metadata.*` on empty metadata returns `[]` today — while ordinary paths like `trace.last.nope` keep returning `NoMatch`.

I have the fix and tests ready and will open a PR shortly.

## Separate observation

`Trace.model_dump()` includes a `last` key that duplicates the final interaction, so `trace..outputs` on a single-interaction trace returns `['hello', 'hello']`. That is a different problem with a different blast radius, so I left it alone — happy to file it separately if useful.

Environment: `main` @ 8aafe58, Python 3.13, `jsonpath-ng` from the pinned range.

Qalipso:
Fixes #2633.

## Problem

`_is_list_expression` handles `Descendants` in the same recursive branch as `Child`:

```python
if isinstance(expression, Child | Descendants):
    return _is_list_expression(expression.right) or _is_list_expression(expression.left)
```

For `trace..ctx` both children are single-field `Fields`, so this returns `False` and `resolve()` falls back to its match-count heuristic (`len(matches) > 1 or _is_list_expression(expression)`). A `..` expression then yields a bare scalar when the data holds exactly one match and a list when it holds two or more.

This is the defect #2605 fixed for dot-wildcards and multi-field selectors. That fix added a `Fields` branch; `..` compiles to `Descendants` and was not covered. `..` is a supported expression — `test_extraction.py` already asserts `trace..outputs` validates.

It breaks `ComparisonCheck`'s `match=any/all/none`, which requires a list/set/tuple:

```
Equals(key="trace..ctx", expected_value="doc-1", match="any")

one interaction carries ctx : FAIL | Expected a list, set, or tuple at key 'trace..ctx'
                                     when match is 'any', but got str.
two interactions carry ctx  : PASS | At least one value in ['doc-1', 'doc-2'] is equal to 'doc-1'.
```

Identical check configuration; the outcome depends on incidental data shape.

## Fix

Give `Descendants` its own branch returning `True`, and leave `Child` recursing. A `..` searches a whole subtree, so its match count is a property of the data, not of the expression — the same reasoning the bracket-wildcard and `Fields` branches already encode.

## Behavior change worth flagging

`trace..missing` now returns `[]` instead of `NoMatch`. That matches how list-expressions already behave — `trace.last.metadata.*` on empty metadata returns `[]` today — while ordinary paths such as `trace.last.nope` keep returning `NoMatch`. Covered by a test.

## Verification

- `uv run pytest libs/giskard-checks -m "not functional"` → **761 passed, 4 skipped** (757 passed, 4 skipped on `main` before this change; the 4 added tests are the difference).
- TDD red→green: reverting only `extraction.py` reproduces `'doc-1' == ['doc-1']` (bare scalar) and fails 2 of the new tests; reapplying passes all 5 in the class.
- `ruff check` / `ruff format --check` on both changed files: clean.
- `basedpyright` on `extraction.py`: 0 errors, 28 warnings — byte-identical to the count on `main`, so nothing new was introduced.

Python 3.13, `jsonpath-ng` from the pinned range.

## Separate observation, not addressed here

`Trace.model_dump()` includes a `last` key duplicating the final interaction, so `trace..outputs` on a single-interaction trace returns `['hello', 'hello']`. Different problem, different blast radius — noted in #2633 and left alone.

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/core/extraction.py
+++ b/libs/giskard-checks/src/giskard/checks/core/extraction.py
@@ -65,7 +65,15 @@ class NoMatch(BaseModel):
 
 
 def _is_list_expression(expression: JSONPath) -> bool:
-    if isinstance(expression, Child | Descendants):
+    # `..` searches a whole subtree, so how many values it finds is a property
+    # of the data, not of the expression. Treat it as a list-expression the way
+    # bracket wildcards are, or resolve() falls back to its match-count
+    # heuristic and returns a bare scalar whenever the data happens to hold one
+    # match.
+    if isinstance(expression, Descendants):
+        return True
+
+    if isinstance(expression, Child):
         return _is_list_expression(expression.right) or _is_list_expression(
             expression.left
         )
--- a/libs/giskard-checks/tests/core/test_extraction.py
+++ b/libs/giskard-checks/tests/core/test_extraction.py
@@ -157,3 +157,42 @@ class TestJSONPathStrAnnotatedType:
 
         with pytest.raises(ValidationError, match="Invalid JSONPath expression"):
             Model(key="trace.last.outputs[")
+
+
+class TestResolveDescendants:
+    """Tests for resolve()'s handling of the descendants operator (``..``).
+
+    ``trace..ctx`` searches the whole trace, so the number of values it finds
+    depends on the data, not on the expression. It must always resolve to a
+    list -- otherwise ComparisonCheck's match=any/all/none rejects the value
+    with a type error whenever exactly one interaction happens to carry the
+    field.
+    """
+
+    @staticmethod
+    def _trace(*metadata: dict[str, str]) -> Trace[str, str]:
+        return Trace[str, str](
+            interactions=[
+                Interaction(inputs=f"q{i}", outputs=f"a{i}", metadata=m)
+                for i, m in enumerate(metadata)
+            ]
+        )
+
+    def test_descendants_with_single_match_returns_a_list_not_a_bare_scalar(self):
+        trace = self._trace({"ctx": "doc-1"}, {})
+        assert resolve(trace, "trace..ctx") == ["doc-1"]
+
+    def test_descendants_with_multiple_matches_returns_a_list(self):
+        trace = self._trace({"ctx": "doc-1"}, {"ctx": "doc-2"}, {})
+        result = resolve(trace, "trace..ctx")
+        assert isinstance(result, list)
+        assert sorted(result) == ["doc-1", "doc-2"]
+
+    def test_descendants_with_no_match_returns_an_empty_list(self):
+        """Consistent with `*`, which already returns [] rather than NoMatch."""
+        trace = self._trace({})
+        assert resolve(trace, "trace..missing") == []
+
+    def test_child_expression_without_wildcard_still_returns_a_bare_scalar(self):
+        trace = self._trace({})
+        assert resolve(trace, "trace.last.outputs") == "a0"


davidberenstein1957 (approved):

libs/giskard-checks/src/giskard/checks/core/extraction.py
```diff
@@ -65,7 +65,15 @@ def __repr__(self) -> str:
 
 
 def _is_list_expression(expression: JSONPath) -> bool:
-    if isinstance(expression, Child | Descendants):
+    # `..` searches a whole subtree, so how many values it finds is a property
+    # of the data, not of the expression. Treat it as a list-expression the way
+    # bracket wildcards are, or resolve() falls back to its match-count
+    # heuristic and returns a bare scalar whenever the data happens to hold one
+    # match.
```
davidberenstein1957, line 72:
```suggestion
```

davidberenstein1957:
Opened a same-repo follow-up with the same fix plus format fixes so CI lint can pass (fork PRs cannot receive the auto-format push): this supersedes the branch here.

Thanks for the clear repro and tests — credited as co-author on the fix commit.

davidberenstein1957:
Opened a formatted follow-up so CI `check-format` can pass: https://github.com/Giskard-AI/giskard-oss/pull/2645

Same change as this PR, plus the markdown/ruff format fix. Please use #2645 going forward — we can close this one once that is green.

Qalipso:
Thanks — happy to have #2645 carry it, and no objection to closing this once that lands.

One heads-up: #2645 currently shows as conflicting with `main` (`mergeStateStatus: DIRTY`), same for #2652. Keeping this branch open until yours is green, as you suggested; say the word and I'll close it, or rebase this one instead if that turns out to be easier.

kevinmessiaen:
Duplicate of #2644

PR #2636, opened 2026-07-26
fix(checks): SemanticSimilarity embeds the repr of a list-valued key

### LINKED ISSUE #2635: SemanticSimilarity embeds the Python repr of a list-valued key
## Summary

`SemanticSimilarity.run()` calls `str()` on whatever `reference_text_key` and `actual_answer_key` resolve to. A key that resolves to several values — a bracket wildcard, `..`, or any multi-match path — is therefore sent to the embedding model as its Python repr, brackets and quotes included. The check then reports an ordinary pass/fail on a cosine similarity computed from text no one wrote.

`Readability` in `builtin/nlp_metrics.py` already guards the equivalent case and fails with a clear message. `SemanticSimilarity` has no such guard.

## Reproduction

```python
import asyncio
import numpy as np
from giskard.agents.embeddings.base import BaseEmbeddingModel, EmbeddingParams
from giskard.checks import SemanticSimilarity

CAPTURED = []

@BaseEmbeddingModel.register("capture")
class Capture(BaseEmbeddingModel):
    async def _embed(self, texts, params: EmbeddingParams | None = None):
        CAPTURED.append(list(texts))
        return [np.array([1.0, 0.0, 0.0]) for _ in texts]

trace = Trace[str, str](interactions=[
    Interaction(inputs="q1", outputs="Paris is the capital of France",
                metadata={"ref": "Paris is the capital"}),
    Interaction(inputs="q2", outputs="Berlin is the capital of Germany",
                metadata={"ref": "Berlin is the capital"}),
])

async def main():
    check = SemanticSimilarity(
        actual_answer_key="trace.last.outputs",
        reference_text_key="trace.interactions[*].metadata.ref",
        threshold=0.5,
        embedding_model=Capture(),
    )
    result = await check.run(trace)
    print("sent to the embedding model:")
    for t in CAPTURED[0]:
        print("   ", repr(t))
    print("result:", result.status.value, "|", result.message)

asyncio.run(main())
```

```
sent to the embedding model:
    'Berlin is the capital of Germany'
    "['Paris is the capital', 'Berlin is the capital']"
result: pass | The cosine similarity with the reference answer is 1.00 which is greater than the threshold 0.50
```

The second string is a Python list repr. Nothing in the result says the reference text was unusable — the check simply reports a score.

## Expected

A key that resolves to a collection should be rejected with a message naming the key, the way `Readability` rejects a value it cannot score:

```python
if not isinstance(text, str):
    return CheckResult.failure(
        message=f"Value for key '{self.key}' must be a string, but found {type(text).__name__}.",
        ...
    )
```

## Note on scope

I went with rejecting `list`/`tuple`/`set`/`dict` rather than requiring `str` outright, so a numeric metadata value (`{"ref": 42}`) keeps stringifying as it does today. A strict `isinstance(value, str)` check, matching `Readability` exactly, is the alternative if you would rather be consistent than lenient — happy to switch.

PR follows.

Environment: `main` @ 8aafe58, Python 3.13.

Qalipso:
Fixes #2635.

## Problem

`run()` calls `str()` on whatever `reference_text_key` and `actual_answer_key` resolve to. A key resolving to several values — a bracket wildcard, `..`, or any multi-match path — reaches the embedding model as its Python repr:

```
reference_text_key="trace.interactions[*].metadata.ref"

sent to the embedding model:
    'Berlin is the capital of Germany'
    "['Paris is the capital', 'Berlin is the capital']"
result: pass | The cosine similarity with the reference answer is 1.00 which is greater than the threshold 0.50
```

The check reports a normal pass on a score computed from a string no one wrote, and nothing in the result flags the value as unusable. `Readability` already guards the equivalent case in `builtin/nlp_metrics.py`; `SemanticSimilarity` did not.

## Fix

Reject `list`/`tuple`/`set`/`dict` before embedding, naming the offending key. Scalars still stringify exactly as before, so a numeric metadata value such as `{"ref": 42}` keeps working.

I chose this over a strict `isinstance(value, str)` — which would match `Readability` exactly — because the strict form would newly reject numbers that stringify sensibly today. Say the word and I'll switch it; the test for the scalar path documents whichever way you prefer.

## Verification

- `uv run pytest libs/giskard-checks -m "not functional"` → **759 passed, 4 skipped** (757 on `main`; the 2 added tests are the difference).
- TDD red→green: reverting only `semantic_similarity.py` fails `test_list_valued_key_is_rejected_instead_of_embedding_its_repr`; reapplying passes. `test_scalar_key_still_stringifies` passes both before and after, which is the point — it pins behavior this change must not alter.
- `ruff check` / `ruff format --check` on both files: clean.
- `basedpyright` on `semantic_similarity.py`: 0 errors, 11 warnings vs 9 on `main`. The two extra are `reportAny`/partially-unknown on the resolved values, the same class already reported elsewhere in this file; `failOnWarnings` is false for this package. I annotated the loop as `list[tuple[str, str, object]]` to keep it to two rather than five — tell me if you would rather I chase them to zero.

Python 3.13.

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/builtin/semantic_similarity.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/semantic_similarity.py
@@ -154,6 +154,30 @@ class SemanticSimilarity[InputType, OutputType, TraceType: Trace](  # pyright: i
                 },
             )
 
+        # A list-valued key (a wildcard, `..`, or a multi-match path) would
+        # otherwise reach str() below and be embedded as its Python repr --
+        # brackets, quotes and all -- yielding a similarity score computed on
+        # text no one wrote. Reject it instead, the way Readability rejects a
+        # value it cannot score. Scalars still stringify as before.
+        resolved: list[tuple[str, str, object]] = [
+            ("reference text", self.reference_text_key, reference_text),
+            ("actual answer", self.actual_answer_key, actual_answer),
+        ]
+        for label, key, value in resolved:
+            if isinstance(value, (list, tuple, set, dict)):
+                return CheckResult.failure(
+                    message=(
+                        f"Value for {label} key '{key}' must be a single value, but "
+                        f"found {type(value).__name__}. Use a key that resolves to "
+                        "one value."
+                    ),
+                    details={
+                        "reference_text_key": self.reference_text_key,
+                        "actual_answer_key": self.actual_answer_key,
+                        "value": str(value),
+                    },
+                )
+
         actual_answer = str(actual_answer)
         reference_text = str(reference_text)
 
--- a/libs/giskard-checks/tests/builtin/test_semantic_similarity.py
+++ b/libs/giskard-checks/tests/builtin/test_semantic_similarity.py
@@ -623,3 +623,59 @@ async def test_similarity_at_exact_threshold() -> None:
     # Should pass when similarity >= threshold
     assert result.status == CheckStatus.PASS
     assert result.details["similarity"] == 1.0
+
+
+@pytest.mark.asyncio
+async def test_list_valued_key_is_rejected_instead_of_embedding_its_repr():
+    """A key resolving to several values must not be embedded as `str(list)`.
+
+    Without this guard the check embeds "['a', 'b']" -- brackets and quotes
+    included -- and reports an ordinary pass/fail on a score computed from
+    text no one wrote.
+    """
+    embedded: list[list[str]] = []
+
+    @BaseEmbeddingModel.register("recording")
+    class RecordingEmbeddingModel(BaseEmbeddingModel):
+        async def _embed(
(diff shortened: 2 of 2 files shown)


davidberenstein1957 (approved):

davidberenstein1957:
Opened a formatted follow-up so CI can pass: https://github.com/Giskard-AI/giskard-oss/pull/2652

Same change as this PR, plus ruff format/lint fixes. Please use #2652 going forward — we can close this one once that is green.

Qalipso:
Same here — fine to let #2652 supersede this.

Heads-up though: #2652 currently shows as conflicting with `main` (`mergeStateStatus: DIRTY`). I'll keep this branch open until yours is green; ping me and I'll close it, or rebase this one if that's the shorter path.

PR #2645, opened 2026-07-29
fix(checks): descendants JSONPath returns scalar vs list depending on match count

davidberenstein1957:
## Summary
Formatted follow-up of https://github.com/Giskard-AI/giskard-oss/pull/2634 (@Qalipso).

Same change as #2634, plus the markdown format fix from `6510a2792` (`style: reformat markdown for ruff 0.16 format check`) so CI `check-format` passes.

## Why
#2634 had `safe for build` but lint failed on:
```
uv run ruff format --check .
```
CI installs a newer ruff that reformats Python fences in READMEs; that failure is unrelated to the fix itself.

## Supersedes
Please close https://github.com/Giskard-AI/giskard-oss/pull/2634 in favor of this PR once CI is green.

## Test plan
- [ ] lint / check-format green
- [ ] Diff vs #2634 is format-only (READMEs / design.md) on top of the original commits

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,10 @@ jobs:
       contents: read # checkout repository
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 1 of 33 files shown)


kevinmessiaen:
Duplicate of #2644

PR #2647, opened 2026-07-29
fix(llm): warn on unknown Anthropic params; detect bare refusals

davidberenstein1957:
## Summary
Formatted follow-up of https://github.com/Giskard-AI/giskard-oss/pull/2623 (@Solaris-star).

Same change as #2623, plus ruff format (markdown fences + any Python the newer CI ruff touches) so `check-format` passes.

## Why
#2623 had `safe for build` but lint failed on:
```
uv run ruff format --check .
```

## Supersedes
Please close https://github.com/Giskard-AI/giskard-oss/pull/2623 in favor of this PR once CI is green.

## Test plan
- [ ] lint / check-format green
- [ ] Functional diff matches #2623; extra commits are format-only


Solaris-star:
Original author here — this is the same change I proposed in #2623, now with the `check-format` fix. Supporting this PR for the merge. Thanks @davidberenstein1957!

kevinmessiaen (approved):

PR #2658, opened 2026-07-29
fix(checks): Not must not invert unevaluable failures

### LINKED ISSUE #2637: Not() turns a check that could not be evaluated into a pass
## Summary

`Not` inverts any `FAIL` into a `PASS`. But `CheckResult.failure` is used for two different things across the built-in checks:

- **the assertion did not hold** — "the value is not equal to X"
- **the assertion could not be evaluated** — "No value found for key", "must be a string, but found list", "Expected a list, set, or tuple ... but got str"

`Not` cannot tell them apart, so a check that could not run reports green.

## Reproduction

```python
import asyncio
from giskard.checks import Not, Equals

trace = Trace[str, str](interactions=[Interaction(inputs="q", outputs="the answer")])

async def main():
    # a key that does not exist -- a typo, a renamed metadata field, a schema change
    inner = Equals(key="trace.last.metadata.nope", expected_value="x")
    print("inner    :", (await inner.run(trace)).status.value)
    print("Not(inner):", (await Not(check=inner).run(trace)).status.value)

asyncio.run(main())
```

```
inner    : fail
Not(inner): pass
```

Message on the passing result: `Check correctly failed (inverted): No value found for key 'trace.last.metadata.nope', expected a value equal to 'x'.`

The same happens for a type mismatch:

```python
Equals(key="trace.last.outputs", expected_value="x", match="any")
# inner : fail  | Expected a list, set, or tuple at key 'trace.last.outputs' when match is 'any', but got str.
# Not(..): pass | Check correctly failed (inverted): Expected a list, set, or tuple ...
```

## Why this matters

`Not(...)` is the natural way to express "the output must not contain X". If the key is misspelled, or the metadata field is later renamed, the suite keeps reporting green forever — the one direction where a silent failure is invisible, because a green suite is not investigated. The failing message is right there in the result, but nothing reads it.

## Possible directions

I did not open a PR because the reasonable fixes all touch the `CheckResult` contract rather than `Not` alone:

1. **A dedicated status or flag for "could not evaluate."** `CheckStatus` already has `ERROR`; the structural cases above could use it, and `Not` already passes `ERROR` through unchanged. This is the smallest conceptual change but shifts behavior for anyone relying on those cases being `FAIL`.
2. **A `CheckResult` field** such as `evaluable: bool`, set false by the structural branches, with `Not` refusing to invert when it is false. More surface area, no change for existing consumers.
3. **Leave the statuses alone and document it** — `Not` inverts everything, so pair it with an explicit existence check.

Happy to implement whichever you prefer, including (3) as a docs-only change.

Environment: `main` @ 8aafe58, Python 3.13.

davidberenstein1957:
## Summary
- Adds `CheckResult.evaluable` (default `True`). Structural “could not evaluate” failures from built-ins set `evaluable=False`.
- `Not` passes unevaluable results through without inverting them, so a missing key or `match=` type mismatch no longer becomes a green suite (#2637).
- Status stays `FAIL` for those cases — additive field only, no FAIL→ERROR migration.

## Test plan
- [x] `uv run pytest libs/giskard-checks/tests/builtin/test_composition.py::TestNot` (10 passed)
- [x] `make test-unit PACKAGE=giskard-checks` (760 passed, 4 skipped)
- [x] `ruff check` / `ruff format --check` on changed Python sources
- [ ] basedpyright skipped locally (pre-existing `reportInvalidTypeForm` on comparison fields; CI still runs it)

Closes #2637

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,10 @@ jobs:
       contents: read # checkout repository
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - uses: astral-sh/setup-uv@d31148d669074a8d0a63714ba94f3201e7020bc3 # v8.3.0
+      - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
         with:
           enable-cache: true
           python-version: "3.12"
@@ -45,10 +45,10 @@ jobs:
     env:
       PACKAGE: ${{ matrix.package }}
     steps:
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - uses: astral-sh/setup-uv@d31148d669074a8d0a63714ba94f3201e7020bc3 # v8.3.0
+      - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
         with:
           enable-cache: true
           python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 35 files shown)


kevinmessiaen:
I would not add a new `evaluable` boolean flag to the check results. It'll probably be cleaner to treat those as check failures.

- missing key: it's should be a failure, the data is not present. A simple check to validate that data is present beforehand would "silent" the error 
- mismatching comparaison type: it is a python error, so propagating an error make sense

PR #2686, opened 2026-08-02
ci: pin ruff so check-format stops failing on main

BILLKISHORE:
## What

`make check-format` fails on a clean `main`, so the lint job is red on new PRs. From the run on #2680:

```
uv run ruff format --check .
unformatted: File would be reformatted
5 files would be reformatted, 291 files already formatted
```

No commit caused this. `install-tools` runs `uv tool install ruff` unpinned, and ruff 0.16 promoted markdown code-block formatting out of preview. CI picked up the new default on its next run, and five docs files that predate it started failing.

Bisecting the tool rather than the tree:

| ruff | markdown formatting | `check-format` on clean main |
| --- | --- | --- |
| 0.14.0 | preview only | pass |
| 0.15.0 | preview only | pass |
| 0.16.1 | on by default | fail |

## Fix

Two commits:

1. `ruff format` over the five affected files (`README.md`, the three package READMEs, `libs/giskard-llm/docs/design.md`). Formatting only: line wrapping, blank lines between top-level defs, quote normalisation, trailing commas. No prose or code semantics changed.
2. Pin `RUFF_VERSION := 0.16.1` in `install-tools`, so the next ruff release can't turn `make check` red on a tree nobody touched. Same reasoning as the existing `LICENSECHECK_VERSION` pin.

I did consider pinning to 0.15 instead, which also makes main green with a one-line diff since 0.15 ignores markdown entirely. Freezing the toolchain to avoid a formatting change you'll take eventually didn't seem like the right call, especially as the same reformat is already riding along in a number of open PRs.

## Testing

`make check` exits 0 on this branch and fails on `main`. `make test-unit`: 37 / 211 / 117 / 757 / 155 passed across core, llm, agents, checks and scan.

## Note

The docs reformat already appears as a side effect in several open PRs (#2658 among them, byte-identical to this one). Landing it standalone lets those diffs shrink to just their real changes.

## Type of Change

---

Disclosure: drafted with AI assistance (Claude Code). I reproduced the failure, bisected the ruff versions, and ran the checks myself before submitting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,13 @@ help: ## Show this help message
 install: ## Install project dependencies
 	uv sync
 
+# Ruff formatting output changes between releases, so an unpinned install can turn
+# `make check` red on a tree nobody touched. Pinned for the same reason as
+# LICENSECHECK_VERSION below.
+RUFF_VERSION := 0.16.1
+
 install-tools: ## Install development tools
-	uv tool install ruff
+	uv tool install ruff==$(RUFF_VERSION)
 	uv tool install vermin
 	uv tool install basedpyright
 	uv tool install pre-commit --with pre-commit-uv
--- a/README.md
+++ b/README.md
@@ -72,12 +72,14 @@ from giskard.checks import Scenario, Groundedness
 
 client = OpenAI()
 
+
 def get_answer(inputs: str) -> str:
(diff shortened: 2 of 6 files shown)


kevinmessiaen (approved):
Thanks for fixing the CI !

PR #2696, opened 2026-08-03
ci: make integration tests more frugal

### LINKED ISSUE #2695: ci: make integration tests more frugal
## Problem

Each authorized PR runs **~21 integration jobs** that hit live LLM APIs. Recent runs showed:

- Redundant provider coverage (`google` + `gemini`, `openai` + `bare`, multiple Azure aliases)
- Triple Python version matrix (3.12–3.14) despite Main CI already running unit tests on all three
- A `both` agents backend job that re-runs the union of giskard-llm and litellm suites
- A `checks` job that runs `make test-functional` but collects **zero** tests (`giskard-checks` uses `@pytest.mark.integration`, not `functional`)

Failures are often infra-related (e.g. OpenAI credits exhausted) rather than regressions, while API cost scales with every PR.

## Proposal (phase 1 — workflow only)

Shrink the integration matrix without changing test code:

1. **Python 3.12 only** for integration jobs (unit tests keep 3.12–3.14 on Main CI)
2. **Drop alias LLM providers** — keep `openai`, `google`, `anthropic`, `azure`; drop `bare`, `gemini`, `azure_ai`, `azure_foundry_v1`
3. **Drop `both` agents backend** — `giskard-llm` and `litellm` jobs already cover both install layouts
4. **Remove dead `checks` integration job** until checks uses `functional` marks or a dedicated smoke tier

**Expected impact:** ~21 jobs → **6 jobs**, ~70% fewer live API calls per PR.

## Follow-ups (phase 2+)

- [ ] Introduce `@pytest.mark.functional_smoke` for PR CI vs full suite on nightly/main
- [ ] Path-filter integration workflow to LLM/agents changes only
- [ ] Wire `giskard-checks` integration tests (`--run-integration`) into a smoke job when ready

davidberenstein1957:
## Summary

Shrinks the integration test matrix and skips unrelated PRs to cut live API usage. Closes Giskard-AI/giskard-oss#2695.

### Matrix (phase 1)

* **Python 3.12 only** for integration — Main CI unit tests already cover 3.12–3.14
* **Drop** `gemini` **only** among alias providers — same `GoogleProvider` as `google`; keep `bare` (no-configure path), `azure_ai` (`AzureAIProvider`), and `azure_foundry_v1` (Foundry base URL + Responses API)
* **Drop** `both` **agents backend** — `giskard-llm` and `litellm` jobs already exercise both install layouts
* **Remove dead** `checks` **job** — it ran `make test-functional` but `giskard-checks` uses `@pytest.mark.integration`, so zero tests were collected

**\~21 jobs → 9 jobs** (\~55% fewer API calls when integration runs).

### Path filter (phase 3)

* Path detection lives in the `authorize` job, **not** in a `paths:` trigger — a required check filtered out by `paths:` never starts and stays "Waiting" forever
* `authorize` lists the PR's changed files and outputs `relevant`; the four test jobs are gated with `if: needs.authorize.outputs.relevant == 'true'`, so they report as *skipped* (neutral) on docs-only PRs instead of blocking the merge
* Relevant paths: `libs/giskard-{agents,llm,core,scan}/**`, this workflow, `pyproject.toml`, `uv.lock`, `Makefile`
* `workflow_dispatch` — always runs the full matrix for manual verification

## Test plan

- [ ] Docs-only PR does not trigger integration
- [ ] LLM/agents PR triggers 9 jobs (2 agents + 7 llm)
- [ ] Agents giskard-llm and litellm jobs pass
- [ ] LLM jobs pass for google/anthropic; openai/azure may fail if org credentials are exhausted

## Follow-ups (not in this PR)

* `functional_smoke` tier for PR vs nightly full suite
* Split workflows with per-package `paths` if agents-only PRs should skip llm jobs without shell
* Wire `giskard-checks --run-integration` when a smoke job exists

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -69,11 +69,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Python version coverage lives in Main CI unit tests; integration only needs one.
         python-version: ["3.12"]
         backend:
-          # One job per install layout (giskard-llm vs litellm). The old "both"
-          # cell re-ran the union of these suites without adding coverage.
           - name: giskard-llm
             extras: "giskard-agents[google]"
             pytest_mark: "functional and google"
@@ -86,13 +83,11 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
           persist-credentials: false
-          # Deliberate opt-in: this job installs and executes fork-authored
-          # code (make install, pytest) with live provider API keys in scope.
-          # Compensating controls: the `ci` environment requires a reviewer to
-          # approve each run before secrets are injected (server-enforced,
-          # per-run), the authorize job requires a fresh `labeled` event for
-          # external contributors, and the immutable head.sha pin keeps the
-          # reviewed commit identical to the executed one.
+          # This job deliberately runs untrusted fork code with live API keys
+          # in scope. Three controls limit the risk. The CI environment
+          # requires reviewer approval for each run. The authorize job
+          # requires a fresh labeled event for external contributors. The
+          # head.sha pin stops the code from changing after approval.
           allow-unsafe-pr-checkout: true
       - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
         with:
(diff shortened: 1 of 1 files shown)


davidberenstein1957:
Added phase 3: path filters on `pull_request_target` plus a `detect-changes` job so agents-only PRs skip llm API jobs.

henchaves:
there is a bug here in which the "Integration tests" workflow will always be on "Waiting" status if none of the paths are touched

davidberenstein1957:
@henchaves good catch — fixed in bb99465. Dropped the `paths:` trigger entirely (a required check that never starts stays "Waiting" forever) and moved path detection into the `authorize` job: it lists the PR's changed files and outputs `relevant`, and the four test jobs are gated with `if: needs.authorize.outputs.relevant == 'true'`. Skipped-by-`if` jobs report as neutral, so branch protection is satisfied on docs-only PRs. `workflow_dispatch` still runs the full matrix.

PR #2412, opened 2026-04-18
checks: add readability NLP metric check

harsh21234i:
Adds a new built-in Readability check backed by textstat, with
  configurable metric, min_score, and max_score. Exposes it via
  giskard.checks exports, wires it into giskard.checks.builtin, adds
  textstat under the nlp optional extra (giskard-checks[nlp]), and includes
  unit tests covering pass/fail cases, missing key/type errors, missing
  dependency behavior, and invalid threshold ranges. Related to issue 

#2349.

PR #2404, opened 2026-04-14
feat: Add buildable giskard meta-package and release workflow support

kevinmessiaen:
## Description

Add buildable giskard meta-package and release workflow support in order to allow installing Giskard packages using `pip install giskard`

PR #2705, opened 2026-08-04
fix(checks): include readability in the all extra

BILLKISHORE:
## What

`libs/giskard-checks/pyproject.toml`:

```toml
readability = ["textstat>=0.7.0"]
regorus = ["celine-regorus>=0.9.1"]
all = ["giskard-checks[regorus]"] # Install all optional dependencies
```

`all` omits `readability`, so `pip install giskard-checks[all]` does not install `textstat` and the `Readability` check fails at runtime with "The 'textstat' package is required for the Readability check", despite the comment promising otherwise.

Resolved against the published package:

```
giskard-checks[all]              -> regorus: 1 line,  textstat: 0 lines
giskard-checks[all,readability]  -> regorus: 1 line,  textstat: 1 line
```

Looks like an oversight rather than a decision: `all` landed in #2404, the `readability` extra was added later in #2412, and `all` wasn't updated.

## Fix

```diff
-all = ["giskard-checks[regorus]"] # Install all optional dependencies
+all = ["giskard-checks[readability,regorus]"] # Install all optional dependencies
```

Plus `uv lock` (2 lines) and `make generate-notices`, which `make check` required once textstat entered the `all` extra.

## One thing worth deciding

The notices diff is 48 lines because textstat brings `nltk`, `joblib`, `pyphen`, `defusedxml` and `setuptools` with it, so `[all]` gets noticeably heavier.

There are two defensible fixes and I only picked one:

- **this PR** — fix the extra to match the comment, so `[all]` genuinely means all
- **the alternative** — fix the comment instead, and document that `all` deliberately excludes `readability` to stay light

I went with the first because a check that fails at runtime seems worse than a heavier optional install, but for a project built around "each package carries only the dependencies it needs" the second is a reasonable call. Happy to flip it if you'd prefer.

## Testing

`make check` exits 0. `make test-unit`: 39 / 214 / 117 / 764 / 159 passed across core, llm, agents, checks and scan.

xref: no open PR touches the `all` extra.

## Type of Change

## Checklist

- [x] I've updated the `uv.lock` running `uv lock`

---

Disclosure: drafted with AI assistance (Claude Code). I reproduced the resolution difference, confirmed the history, and ran the checks myself before submitting.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -29,6 +29,7 @@ Find a list of packages below
 |✔|click|
 |✔|colorama|
 |✔|cryptography|
+|✔|defusedxml|
 |✔|distro|
 |✔|docstring-parser|
 |✔|fastuuid|
@@ -49,6 +50,7 @@ Find a list of packages below
 |✔|importlib-metadata|
 |✔|jinja2|
 |✔|jiter|
+|✔|joblib|
 |✔|jsonpath-ng|
 |✔|jsonschema|
 |✔|jsonschema-specifications|
@@ -58,6 +60,7 @@ Find a list of packages below
 |✔|markupsafe|
 |✔|mdurl|
 |✔|multidict|
+|✔|nltk|
 |✔|numpy|
 |✔|openai|
 |✔|packaging|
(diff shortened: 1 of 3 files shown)


kevinmessiaen (approved):

PR #2720, opened 2026-08-12
refactor(checks)!: drop LesserThan and LesserThanEquals classes

kevinmessiaen:
Closes [ENG-1677](https://linear.app/giskard/issue/ENG-1677/drop-lesserthan-and-lesserthanequals-checks)

## What

The Giskard Hub v3 now emits the correct `LessThan` / `LessThanEquals` names, so the misspelled alias classes have no remaining producer. This removes them from the public API while keeping their serialized `kind` strings loadable.

## Why not a hard drop

`LesserThan` / `LesserThanEquals` were not just Python aliases — they were registered in the discriminated-union registry via `@Check.register("lesser_than")`. Deleting them outright would make any check JSON persisted *before* the Hub upgrade fail to deserialize with a hard `ValidationError`, not a degraded read.

So the classes are gone, but the legacy kinds still load:

```python
@Check.register("less_than", aliases=["lesser_than"])
class LessThan(...): ...
```

`aliases=` registers extra discriminators on the **read path only**. Only the canonical kind is mapped back from the class, so a legacy payload loads as a real `LessThan` and is re-saved as `less_than` — migrating itself on the next write.

This lives in the shared `Discriminated` registry rather than locally in `giskard-checks` because `__get_pydantic_core_schema__` installs a plain validator that *replaces* the model schema. Validation dispatches to the concrete class before any `Check`-level `model_validator` could rewrite `kind`, so a downstream fix would have required monkeypatching registry internals.

## Registry fixes found while in there

- **Atomic registration.** Validating and writing in a single pass left the registry half-populated when a later alias collided — a class whose decorator raised stayed loadable. Now validate-then-write.
- **Dropped `_reverse_kinds`,** which held data byte-identical to `_subclasses` (verified across all 7 registered bases). Reading `_subclasses` via `.get()` also stops the former `defaultdict` from inserting an empty entry on each failed lookup.
- **Reject a bare `str` for `aliases`.** `aliases="lesser_than"` would silently register one kind per character; `str` satisfies `Sequence[str]`, so no type checker catches it. `aliases` is now keyword-only.

## Breaking change

`LesserThan` and `LesserThanEquals` are no longer importable from `giskard.checks` — use `LessThan` / `LessThanEquals`. **Stored payloads using the legacy `kind` strings are unaffected.** Worth a release-note line for OSS 3.0.

## Testing

- `giskard-core` 47 passed, `giskard-checks` 804 passed / 4 skipped
- `make check` exit 0 (basedpyright 0 errors, pip-audit + licensecheck clean)
- The atomicity test was verified to actually catch the bug — it fails with `DID NOT RAISE` against the single-pass version
- `giskard-agents` has 7 pre-existing failures from a missing Gemini API key; confirmed identical on a clean tree

## Follow-up

Legacy alias hits are not observable: `_telemetry_props.py` reads `kind` off the instance, which now always returns the canonical name. Without a counter or debug log at the alias-match site, these aliases can never be retired on evidence. Worth a separate ticket.


libs/giskard-checks/tests/builtin/test_comparison.py
henchaves, line 1: there are many `lesser_than` in method names within this file

henchaves (approved):
LGTM, only need to update method names in `test_comparison.py`

PR #2722, opened 2026-08-12
fix(release): stop default inputs inflating version, add stable bump

### LINKED ISSUE #2721: Release workflow: version-bump defaults and missing GA path
Feedback from a review of `.github/workflows/release.yml`. Three blockers were reported; two are fixed, one is deliberately deferred.

### 1. Default inputs silently inflate the version

`bump_type` defaulted to `patch` while `release_type` defaulted to `none`, and both are appended as `--bump` flags to the same command (lines 138–146). They compound rather than override, so an operator selecting `release_type=rc` and leaving `bump_type` alone shipped an extra patch bump. PyPI versions are immutable, so this is unrecoverable.

Reproduced with real `uv version`:

```
1.0.1b6  --bump rc                -> 1.0.1rc1   # intended
1.0.1b6  --bump patch --bump rc   -> 1.0.2rc1   # what the defaults gave you
3.0.0b1  --bump patch --bump rc   -> 3.0.1rc1
```

**Fix:** `bump_type` now defaults to `none`, and its description notes that it is applied before `release_type` (the two dropdowns stack). Compounding is still available deliberately — e.g. `minor` + `rc` to open a new pre-release line — it is just no longer the silent default.

### 2. No path to a final release

`release_type` offered only `alpha|beta|rc|none`, so there was no way to promote a pre-release to its final version. Going `rc1 -> 1.0.0` needs `uv version --bump stable`. Verified:

```
1.0.0rc1 --bump stable  -> 1.0.0
1.0.0rc1 --bump patch   -> 1.0.1
```

**Fix:** `stable` added to the `release_type` choice list.

### Note on the new defaults

With both inputs defaulting to `none`, dispatching the workflow without changing either dropdown now fails fast at the existing `exit 1` guard ("No version bump", line 153). This is intended: the failure moves from *silently publishing an inflated version to immutable PyPI* to *erroring before anything is built*. Releasing now requires a deliberate selection.

## Deferred — publishing is not gated on `main`

Not addressed in this change, by decision.

`Push to main` (line 179) requires `dry_run == 'false' && github.ref_name == 'main'`, but **Create tag** (188), **Create GitHub Release** (200) and **Push to PyPI** (214) check `dry_run == 'false'` only.

Cutting a non-dry-run release from a non-`main` branch therefore skips the push but still tags, releases and publishes — producing an immutable PyPI version whose bump commit exists nowhere on `main`, and a tag on an unreachable commit.

This is operator error rather than a routine path, since releases are cut from `main` in practice. Options if it is picked up later:

* Fail fast at job start: one guard step erroring when `dry_run == 'false' && github.ref_name != 'main'`. Covers any publishing step added later, and keeps dry runs working on any branch.
* Add `&& github.ref_name == 'main'` to each of the three unguarded steps.
* Enforce a branch policy on the `release` GitHub environment (line 61) instead, which no workflow edit can bypass.

kevinmessiaen:
Fixes Giskard-AI/giskard-oss#2721

Fixes two blockers found reviewing `.github/workflows/release.yml`. Tracked in Giskard-AI/giskard-oss#2721.

## Default inputs silently inflated the version

`bump_type` defaulted to `patch` while `release_type` defaulted to `none`, and both are appended as `--bump` flags to the same `uv version` command (lines 138–146). They **compound rather than override**, so an operator selecting `release_type=rc` and leaving `bump_type` alone shipped an extra patch bump. PyPI versions are immutable, so this is unrecoverable.

Reproduced against real `uv version`:

```
1.0.1b6  --bump rc                -> 1.0.1rc1   # intended
1.0.1b6  --bump patch --bump rc   -> 1.0.2rc1   # what the defaults gave you
3.0.0b1  --bump patch --bump rc   -> 3.0.1rc1
```

`bump_type` now defaults to `none`. Compounding is still available deliberately — `minor` + `rc` to open a new pre-release line — it is just no longer the silent default.

## No path to a final release

`release_type` offered only `alpha|beta|rc|none`, so there was no way to promote a pre-release to its final version. Verified:

```
1.0.0rc1 --bump stable  -> 1.0.0
1.0.0rc1 --bump patch   -> 1.0.1
```

`stable` added to the choice list.

## Behavior change to be aware of

With both inputs now defaulting to `none`, dispatching the workflow without changing either dropdown fails at the pre-existing `exit 1` guard ("No version bump", line 153). This is the point of the fix: the failure moves from *silently publishing an inflated version to immutable PyPI* to *erroring before anything is built*. Releasing now requires a deliberate selection.

## Not included

Publishing is still not gated on `main` — `Push to main` checks `github.ref_name == 'main'` but **Create tag**, **Create GitHub Release** and **Push to PyPI** check `dry_run` only. Deliberately deferred; options are recorded in Giskard-AI/giskard-oss#2721.

## Testing

Bump semantics reproduced against real `uv version` on a scratch package (table above). Workflow YAML re-parsed after the edit to confirm the inputs and that the four publishing-step conditions are unchanged.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


.github/workflows/release.yml
```diff
@@ -15,23 +15,24 @@ on:
           - giskard-checks
           - giskard-scan
       bump_type:
-        description: 'Version bump type'
+        description: 'Version bump type (stacks with alpha/beta/rc; keep none for a plain pre-release; stable requires none)'
```
henchaves: it's not clear for me when to use each one, especially the "stable requires none". Is this for promoting a pre-release to a stable one, right?

.github/workflows/release.yml
```diff
@@ -15,23 +15,24 @@ on:
...
         options:
           - major
           - minor
           - patch
           - none
-        default: patch
+        default: none
       release_type:
-        description: 'Release type'
+        description: 'Release type (stable promotes a pre-release to its final version; cannot stack with bump_type)'
```
henchaves: the comment is a bit confusing ("cannot stack with bump_type", but `bump_type` says it stacks with alpha/beta/rc)

.github/workflows/release.yml
```diff
@@ -15,23 +15,24 @@ on:
...
         options:
           - major
           - minor
           - patch
           - none
-        default: patch
+        default: none
       release_type:
-        description: 'Release type'
+        description: 'alpha/beta/rc starts or advances a pre-release (1.2.3rc1 -> 1.2.3rc2). stable drops the suffix to publish the final version (1.2.3rc2 -> 1.2.3) and requires bump_type=none. Leave as none to publish a plain version bump.'
```
davidberenstein1957, line 28: This might still be ann issue:

```
1.2.3rc2  --bump stable                -> 1.2.3   # intended
1.2.3rc2  --bump patch --bump stable   -> 1.2.4   # silently wrong, immutable on PyPI
```

Suggest adding a guard makes it unrepresentable instead of relying on a tooltip in the UI.

```yaml
    steps:
      - name: Validate release inputs
        run: |
          if [ "$RELEASE_TYPE" = "stable" ] && [ "$BUMP_TYPE" != "none" ]; then
            echo "::error::release_type=stable promotes a pre-release to its final version \
          (1.2.3rc2 -> 1.2.3) and must be dispatched with bump_type=none. Got bump_type=$BUMP_TYPE, \
          which would compound into '--bump $BUMP_TYPE --bump stable' and skip the version you are \
          stabilizing (1.2.3rc2 -> 1.2.4)."
            exit 1
          fi
```

Not blocking though

davidberenstein1957 (approved):
LGTM. Left a minor suggestion to *enforce* the `bump_type=none` requirement for `stable` rather than only documenting it in the input description.

Issue #2719, opened 2026-08-11
pydantic 2.14.0b1 breaks 'import giskard.checks' (TypeError: unhashable type: 'list')


sk8ordie84:
**Summary.** The current giskard-checks beta fails at import time under pydantic 2.14.0b1. Anyone installing with `--pre` hits this today (pip resolves pydantic to the prerelease too), and it becomes a hard break for everyone the day pydantic 2.14.0 goes stable.

**Repro** (Python 3.12.13, clean venv):

```
pip install --upgrade --pre giskard-checks   # resolves giskard-checks 1.0.2b6 + pydantic 2.14.0b1
python -c "import giskard.checks"
```

```
File ".../giskard/checks/core/interaction/interact.py", line 17, in <generic parameters of Interact>
File ".../giskard/core/discriminated.py", line 102, in __get_pydantic_core_schema__
...
File ".../pydantic/_internal/_generate_schema.py", line 1127, in match_type
  if origin is not None and (aliased_obj := typing_objects.DEPRECATED_ALIASES.get(obj)):
TypeError: unhashable type: 'list'
```

**What we verified.** Same giskard-checks 1.0.2b6 imports cleanly on pydantic 2.13.4 — the trigger is the `Discriminated.__get_pydantic_core_schema__` path (giskard/core/discriminated.py:102) feeding an unhashable annotation into pydantic 2.14's new `DEPRECATED_ALIASES.get(obj)` lookup in `match_type`. Arguably also a pydantic regression (`match_type` assuming hashable `obj`), so possibly worth reporting there too — happy to file it if useful.

**Workaround** we use in CI: after the `--pre` install, re-pin pydantic to the latest stable.

Context: we hit this in the weekly compatibility canary for [falsify-giskard](<https://github.com/studio-11-co/falsify-giskard>), which tests the adapter against your latest prereleases.

---

## Investigation: this is an upstream pydantic regression, not a giskard bug

Reproduced and root-caused. Two conclusions change what should be done here.

### 1. It reproduces with no giskard code at all

```python
from typing import Any
from pydantic import BaseModel

type Alias[**P, R] = Any

class M(BaseModel):
    f: Alias[[], int] | None = None
```

On pydantic 2.14.0b1 (Python 3.12) this raises the identical error. Eight lines, no giskard import.

### 2. Mechanism

[...]

PR #2752, opened 2026-08-12
chore(deps): cap pydantic <2.14 as temporary stopgap

davidberenstein1957:
## What

Tightens the pydantic upper bound from `<3` to `<2.14` in the four libs that constrain it:

- `libs/giskard-core/pyproject.toml` — `>=2.11.0,<2.14`
- `libs/giskard-llm/pyproject.toml` — `>=2.11.0,<2.14`
- `libs/giskard-agents/pyproject.toml` — `>=2.11.0,<2.14`
- `libs/giskard-checks/pyproject.toml` — `>=2.12,<2.14`

Each gets a `TODO(#2719)` comment naming the upstream bug so the cap is removed rather than forgotten. `uv.lock` only records the new specifiers — the resolved version is unchanged (pydantic 2.13.4).

## Why

pydantic 2.14 added an unguarded dict lookup in `_internal/_generate_schema.py` (`match_type`):

```python
if origin is not None and (aliased_obj := typing_objects.DEPRECATED_ALIASES.get(obj)):
```

`.get(obj)` assumes `obj` is hashable. A PEP 695 alias with a **ParamSpec** parameterized by an empty list has `__args__ == ([], ...)`, so hashing it raises `TypeError: unhashable type: 'list'`. In this tree the trigger is `GeneratorType[[], InputType, None]` at `libs/giskard-checks/src/giskard/checks/core/interaction/interact.py:17`, which breaks `import giskard.checks` outright under pydantic 2.14.0b1.

This is an upstream regression, not a giskard bug — it reproduces in eight lines with no giskard import (see #2719). No giskard-side patch would fix it; `discriminated.py` is only the call site in the traceback.

## Scope

Deliberately a stopgap. A hard ceiling on a core dependency shipped in a 1.0 is inherited by every downstream user and can deadlock resolution for anyone needing pydantic 2.14 for another package. `Refs #2719` rather than `Closes` — the issue should stay open until the upstream bug is filed and fixed, at which point the cap is lifted.

The upstream report draft is attached as a comment on #2719 so it is preserved; it has **not** been filed on the pydantic repo yet.

## Testing

- `make check` — clean (basedpyright 0 errors, pip-audit, licensecheck all pass)
- `uv run python -c "import giskard.checks"` — imports cleanly, pydantic 2.13.4
- `uv lock --check` — lockfile consistent

### DIFF
--- a/libs/giskard-agents/pyproject.toml
+++ b/libs/giskard-agents/pyproject.toml
@@ -8,7 +8,9 @@ authors = [{ name = "Giskard R&D", email = "rnd@giskard.ai" }]
 requires-python = ">=3.12"
 dependencies = [
     "giskard-llm>=1.0.0b6,<2",
-    "pydantic>=2.11.0,<3",
+    # TODO(#2719): temporary <2.14 cap. pydantic 2.14 regressed on unhashable
+    # annotations in _generate_schema.match_type. Lift once fixed upstream.
+    "pydantic>=2.11.0,<2.14",
     "griffe>=1.7.0,<3",
     "typing-extensions>=4.14.0,<5",
     "jinja2>=3.1.6,<4",
--- a/libs/giskard-checks/pyproject.toml
+++ b/libs/giskard-checks/pyproject.toml
@@ -7,7 +7,9 @@ license = { text = "MIT" }
 requires-python = ">=3.12"
 dependencies = [
     "jsonpath-ng>=1.7.0,<2",
-    "pydantic>=2.12,<3",
+    # TODO(#2719): temporary <2.14 cap. pydantic 2.14 regressed on unhashable
+    # annotations in _generate_schema.match_type. Lift once fixed upstream.
+    "pydantic>=2.12,<2.14",
     "pydantic-settings>=2.7,<3",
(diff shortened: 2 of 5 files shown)


davidberenstein1957:
Upstream status: the bug was filed and **already fixed upstream** — [pydantic/pydantic#13645](https://github.com/pydantic/pydantic/issues/13645), closed as completed on 2026-08-12 ("This will be fixed in the next version." — @Viicos).

No fixed release exists on PyPI yet: the newest published versions are `2.13.4` (stable) and `2.14.0b1` / `2.14.0a1` (pre-releases), and `2.14.0b1` is the broken one. So the `<2.14` cap stays as-is for now rather than being narrowed to a specific fixed version.

Follow-up: once pydantic ships the next 2.14 pre-release/release, verify `import giskard.checks` against it and lift the cap back to `<3` (the `TODO(#2719)` markers point at it).

Also synced with `main` (clean merge, no conflicts, lockfile untouched); `libs/giskard-checks` tests pass (808 passed, 4 skipped) and `uv lock --check` is consistent.

davidberenstein1957:
Closing this because it is caused by a pydantic beta release.

PR #2754, opened 2026-08-12
refactor(checks)!: unify extraction key parameter names

### LINKED ISSUE #2728: Unify extraction key parameter names across checks before the API freezes
The same concept — "which JSONPath key to extract the value under test from" — is spelled three different ways across the built-in checks. Giskard-AI/giskard-oss#2659 / Giskard-AI/giskard-oss#2672 aimed at this but the surface is still inconsistent.

## Current state

| Check | Parameter | Location |
| -- | -- | -- |
| Comparison family, `JsonValid`, `Readability`, `RegoPolicy` | `key` | `comparison.py:39`, `json_valid.py:33`, `nlp_metrics.py:71`, `rego_policy.py:105` |
| `StringMatching` / `RegexMatching` | `text_key` | `text_matching.py:47` |
| `SemanticSimilarity` | `actual_answer_key` | `semantic_similarity.py:97` |
| `SemanticSimilarity` | `reference_text_key` | `semantic_similarity.py:93` |

`actual_answer_key` is the clearest outlier — it means exactly what `key` means everywhere else, and even defaults to the same `"trace.last.outputs"`.

## Separate but related: ComparisonCheck.key has no default

`libs/giskard-checks/src/giskard/checks/builtin/comparison.py:39`

```python
key: JSONPathStr = Field(..., ...)   # required
```

Every peer defaults to `"trace.last.outputs"`. So `StringMatching(...)` works with no key but `Equals(expected_value=x)` does not. Adding a default is technically non-breaking and could land any time, but the inconsistency is what users encode in their suites — fix it in the same pass.

## Why now

These are pydantic field names, so they are simultaneously the Python kwarg **and** the key in serialized suite JSON. Renaming post-RC breaks every saved suite file, not just source code.

## Open decision

Which spelling wins. Recommendation: standardise on `key`, since it is already the majority and the shortest. `SemanticSimilarity` then needs `key` + `reference_key` (or similar) — worth a moment's thought since it is the one check with two genuinely distinct inputs, so a bare `key` may read as ambiguous there.

## Related

Giskard-AI/giskard-oss#2659, Giskard-AI/giskard-oss#2672. Groups with the `ScenarioResult.steps` rename — both touch pydantic field names and serialized schemas; share one release note.

davidberenstein1957:
## What

Unifies the parameter name used to point a built-in check at the value it should extract from the trace. Everything now uses `key` for the primary extraction path.

Most checks (`Equals`/`NotEquals`/`LessThan`/`GreaterThan`/…, `JsonValid`, `Readability`, `RegoPolicy`) already used `key`; `StringMatching`/`RegexMatching` and `SemanticSimilarity` were the outliers.

## Rename table

| Check | Old parameter | New parameter | Serialized field affected |
| -- | -- | -- | -- |
| `StringMatching`, `RegexMatching` (`TextBasedCheck`) | `text_key` | `key` | `text_key` → `key` |
| `SemanticSimilarity` | `actual_answer_key` | `key` | `actual_answer_key` → `key` |
| `SemanticSimilarity` | `reference_text_key` | `reference_key` | `reference_text_key` → `reference_key` |

Also: `ComparisonCheck.key` was a required field; it now defaults to `"trace.last.outputs"`, matching every peer check.

`CheckResult.details` emitted by `SemanticSimilarity` uses the new names too (`key`, `reference_key`).

## Breaking change

These are pydantic field names, so both constructor kwargs and serialized suite JSON change. No aliases or back-compat shims are provided — this is intentional pre-RC breakage. This should go into the same release note as the `ScenarioResult.steps` rename.

## Test plan

* `make format && make check` — clean
* `uv run pytest libs/giskard-checks/tests -q -m "not functional"` — 808 passed, 4 skipped (after merging latest main)
* `uv run pytest libs/giskard-scan/tests libs/giskard-agents/tests -q -m "not functional"` — 299 passed, 46 skipped (after merging latest main; run per-lib)

Closes Giskard-AI/giskard-oss#2728

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/libs/giskard-checks/README.md
+++ b/libs/giskard-checks/README.md
@@ -360,7 +360,7 @@ result = await (
         StringMatching(
             name="contains_paris",
             keyword="Paris",
-            text_key="trace.last.outputs.answer",
+            key="trace.last.outputs.answer",
         )
     )
     .check(
@@ -405,7 +405,7 @@ result = await (
     .check(
         RegexMatching(
             pattern="test@example.com",
-            text_key="trace.last.outputs",
+            key="trace.last.outputs",
         )
     )
     .run()
--- a/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
@@ -36,7 +36,8 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
     """
 
     key: JSONPathStr = Field(
-        ..., description="The key to extract the actual value from the trace"
+        default="trace.last.outputs",
+        description="The key to extract the actual value from the trace",
     )
     expected_value: ExpectedType | MISSING = Field(
         default=MISSING,
@@ -276,6 +277,7 @@ class LessThan[InputType, OutputType, TraceType: Trace, ExpectedType](  # pyrigh
         The expected value to compare against the extracted values
     key : str
         The key to extract the actual value from the trace
+        (default: "trace.last.outputs")
     """
 
     @override
@@ -319,6 +321,7 @@ class GreaterThan[InputType, OutputType, TraceType: Trace, ExpectedType](  # pyr
         The expected value to compare against the extracted values
     key : str
         The key to extract the actual value from the trace
+        (default: "trace.last.outputs")
     """
 
     @override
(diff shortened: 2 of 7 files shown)


henchaves:
Hi @davidberenstein1957, is this still a draft?

kevinmessiaen:
I would pospone this breaking change

PR #2765, opened 2026-08-13
fix(google): map Interactions timeouts to LLMTimeoutError

BILLKISHORE:
## What

In `providers/google.py`, `_map_error` handles the Interactions API hierarchy in this order:

```python
if isinstance(e, ix.APIStatusError):      ...
if isinstance(e, ix.APIConnectionError):  raise LLMError(0, str(e), PROVIDER) from e
if isinstance(e, ix.APIError):            ...

# only reached when nothing above matched
if "timed out" in str(e).lower() or "timeout" in type(e).__name__.lower():
    raise LLMTimeoutError(408, str(e), PROVIDER) from e
```

`ix.APITimeoutError` subclasses `ix.APIConnectionError`, so a real SDK timeout is caught by the connection branch and the timeout heuristic below is unreachable for it.

The effect isn't only a mislabelled error. `should_retry` retries `LLMTimeoutError` and not `LLMError`:

```
openai timeout       -> LLMTimeoutError  should_retry=True
gemini/ix timeout    -> LLMError         should_retry=False
```

So Gemini Interactions timeouts were never retried, while the OpenAI and Anthropic providers retried theirs. Both of those check `APITimeoutError` explicitly before the broader branches; this path never checks it.

## Fix

Check `APITimeoutError` before `APIConnectionError`, matching the other two providers:

```diff
+            # Must precede APIConnectionError: APITimeoutError subclasses it.
+            if isinstance(e, ix.APITimeoutError):
+                raise LLMTimeoutError(408, str(e), PROVIDER) from e
             if isinstance(e, ix.APIConnectionError):
                 raise LLMError(0, str(e), PROVIDER) from e
```

The subclass relationship is the same in all three SDKs, so the ordering requirement is not specific to Gemini:

```
openai                 APITimeoutError subclasses APIConnectionError: True
anthropic              APITimeoutError subclasses APIConnectionError: True
google/_interactions   APITimeoutError subclasses APIConnectionError: True
```

408 also matches Google's own retry guidance, which recommends retrying only on 429, 408 and 5xx.

## Why `test_google_map_error_completeness` didn't catch it

That test asserts `pytest.raises(LLMError)` for every SDK exception. `LLMTimeoutError` is an `LLMError`, so mapping a timeout to the base class still satisfies it. The new tests assert the specific type instead.

## Testing

Two tests added next to the existing completeness test:

- `test_google_interactions_timeout_maps_to_timeout_error` — asserts `LLMTimeoutError` and that `should_retry` returns True. Fails before the change, passes after.
- `test_google_interactions_connection_error_is_not_a_timeout` — asserts a plain connection error stays a non-retryable `LLMError`. Passes before and after, so it pins that this change doesn't alter non-timeout behaviour.

`make check` exits 0. `make test-unit`: 41 / 220 / 129 / 805 / 168 passed across core, llm, agents, checks and scan.

xref: no open PR touches `providers/google.py`.

## Type of Change

---

Disclosure: drafted with AI assistance (Claude Code). I reproduced the mapping and the retry consequence, confirmed the tests fail before the fix and pass after, and ran the checks myself before submitting.


ebarkhordar (commented):
Nothing has run on this PR yet. `Main CI` and `GitHub Actions Security Analysis` sit at `action_required` waiting for approval, and `Integration Tests` fails at its `authorize` job, which needs a maintainer to add the `safe for build` label (all five test jobs there are `needs: authorize`). The held runs do not appear in the check list at all, so it looks lighter than it is. Ran the provider tests in a container instead.

On `google-genai==2.8.0` (clean `python:3.12-slim`, `pip install ./libs/giskard-core "./libs/giskard-llm[all]"`), `pytest tests/test_providers.py -k google` goes from 10 passed on `main` (14a67cd4) to 12 passed here (c38421a7), and deleting just the two new lines from the installed source turns `test_google_interactions_timeout_maps_to_timeout_error` red with `LLMError` at `google.py:249`. Fix and regression test both hold.

The version range is the catch. Both new tests do `from google.genai import _interactions as ix`, and that module is gone as of google-genai 2.9.0: present in 2.0.0 through 2.8.0, absent in 2.9.0 through 2.18.1, checked one version at a time. `google-genai>=2.0,<3` installs 2.18.1 today, and there both new tests fail at `tests/test_providers.py:597` with `ImportError`. Production avoids this by going through `_import_interactions_errors()`, which logs and returns `None`, so `_map_error` skips the whole `ix` block. Using that helper in the tests, and skipping when it returns `None`, would make them agree with production.

Separately, and not something this PR needs to solve: on 2.18.1 nothing under `google.genai` exposes `APITimeoutError` any more, and `google.genai.interactions` carries only the Interactions types. So from 2.9.0 onward the `ix` block never runs and a timeout falls through to the string check at the end of `_map_error`.

BILLKISHORE:
Thanks, that's a real hole and I've fixed it in c7eb2dc.

Confirmed the version range independently. On a clean 3.12 venv with `google-genai==2.18.1`:

```
_interactions   : ABSENT
interactions    : PRESENT  APITimeoutError=False
```

So both tests would have failed on anything the constraint resolves to today. My local venv is on 2.4.0 from `uv.lock`, which is why they passed for me and why I didn't catch it.

Took your suggestion directly: the tests now resolve the module through `_import_interactions_errors()` and skip when it returns `None`, so they follow the same path `_map_error` does instead of an import the library tolerates losing.

Verified both ways:

- `google-genai==2.18.1`, giskard installed from source: **2 skipped**
- `google-genai==2.4.0` (lockfile): **2 passed**, and reverting just the two source lines still turns `test_google_interactions_timeout_maps_to_timeout_error` red at `google.py:249`

On your separate point about 2.9.0 onward — I checked, and it looks benign. With the `ix` block skipped on 2.18.1, httpx timeouts do reach the string check at the end:

```
ReadTimeout      -> LLMTimeoutError  retry=True
ConnectTimeout   -> LLMTimeoutError  retry=True
```

`type(e).__name__.lower()` contains "timeout" for both. So timeouts are still mapped correctly there, and this PR closes the gap for 2.0.0 through 2.8.0 — which is the range `uv.lock` actually pins.

Worth saying the `ix` block is dead code from 2.9.0 onward either way, and the public `google.genai.interactions` carries no exception classes to replace it. Happy to open a separate issue for that if it's useful, rather than widening this one.

kevinmessiaen (approved):
LGTM!

PR #2768, opened 2026-08-14
fix(checks)!: reject unknown fields across the persisted-scenario tree

kevinmessiaen:
## Why

Pydantic ignores unknown keys by default, and nearly every field in the persisted-scenario tree has a default. A stale or misspelled key in a saved scenario therefore fell back to that default and the check ran **green while evaluating the wrong value** — a silent wrong answer with no signal at all.

This is prerequisite work for [OSS-122](https://linear.app/giskard/issue/OSS-122/unify-extraction-key-parameter-names-across-checks-before-the-api). That issue proposes renaming extraction key parameters, and renaming is unsafe while a dropped key is invisible: a user's saved suite would silently bind to `trace.last.outputs` instead of erroring. **This PR does not do the rename** — it makes the rename safe to do, at any point, before or after the RC.

The hazard was not hypothetical. Turning `extra="forbid"` on immediately caught two of our own doctests in `composition.py` passing `key=` to `StringMatching`, which has no `key` field. They were silently dropping the argument and only looked correct because the default happened to match what was written.

## What changed

`extra="forbid"` now applies across the tree that a persisted scenario deserializes into:

```
Scenario                tolerant  ← deliberate, see below
└─ steps[]  Step                        forbid
   ├─ interacts[]  InteractionSpec      forbid
   │   └─ inputs   InputGenerator       forbid
   └─ checks[]     Check                forbid
       ├─ generator  BaseGenerator      forbid
       │   └─ middlewares[]  CompletionMiddleware  forbid
       └─ embedding_model    BaseEmbeddingModel    forbid
```

**`Scenario` deliberately stays permissive.** It parses JSONL from remote Hugging Face dataset repositories (`giskardai/do-not-answer-scenarios`, `giskardai/harmbench-scenarios`) that release independently of this package. Forbidding extras there would turn any upstream field addition into a hard failure for every already-installed version. Verified: all 1,339 remote lines would pass under a forbidding config today, so this is a forward-risk decision, not a compatibility fix. The rationale is recorded in the `Scenario` docstring so it does not get "fixed" later.

**`BaseRateLimiter` is untouched** — no deserialization entry point, `frozen=True`, backed by a process-local instance registry.

### Enabling change

`kind` is a `computed_field`, so `model_dump()` emits it while it is not a model field. Any model with `extra="forbid"` therefore rejected its own dump, which would have broken **every polymorphic deserialization** — no saved suite would load at all. A `mode="before"` validator on `Discriminated` drops the redundant key. It sits on the base rather than on `Check` because the defect belongs to `Discriminated`, and because it must cover the direct `Subclass.model_validate(dumped)` path as well as polymorphic field validation.

### Bugs found and fixed along the way

| Bug | Effect |
|---|---|
| `composition.py` doctests passed `key=` to `StringMatching` | Argument silently dropped |
| `prompt_injection.jsonl` had top-level `category` | Discarded on every load, unreachable at runtime |
| `Step` accepted a misspelled `checks` | **Vacuous pass** — step ran interactions, asserted nothing |
| `BaseEmbeddingModel.model` fully defaulted | A typo'd `model_name` silently embedded with the wrong model |

`ComparisonCheck.key` also gains the peer default `trace.last.outputs`, so `Equals(expected_value=...)` works without an explicit key like every other check. Ordering is deliberate: its required-ness was previously the only reason stale comparison suites failed loudly, and that protection is systemic as of the `extra="forbid"` commit.

The JSONPath enforcement test had four coverage holes plus one found while fixing them — judges were covered only incidentally via `giskard.checks.__init__`; 38 of 63 walked entries were generic parametrizations; third-party subclasses could fail our suite; and `__subclasses__()` never yields the root, so a JSONPath field on `Check` itself would have gone unchecked forever.

## Known gap, not addressed

`Interact.inputs` is a union whose final arm resolves to `Any` on an unparametrized `Interact`, so a payload the `InputGenerator` arm rejects falls through and is kept as a raw dict. The nested `InputGenerator` typo therefore still does not raise through `Scenario`. Making `InputGenerator` stricter sends *more* payloads down that arm. Narrowing the union is a behavior change and is left for follow-up; `test_nested_input_generator_typo_does_not_reach_a_generator` pins the current behavior and proves the fallthrough predates this PR.

## Breaking change

Saved suites carrying keys that are not real fields now raise `ValidationError` on load instead of silently ignoring them. That is the intent — those suites were already not doing what their author wrote — but it warrants a changelog entry (see OSS-127).

## Verification

| Suite | Result |
|---|---|
| giskard-checks | 903 passed, 4 skipped |
| giskard-core | 46 passed |
| giskard-scan | 169 passed, 46 skipped |
| giskard-agents | `test_serialization.py` 14/14; failure set byte-identical to baseline |
| ruff / format / basedpyright | clean across four libs |

`giskard-agents` and `giskard-llm` carry pre-existing failures from live-API tests with no credentials (`api_key = None`). Those counts drift between runs, so the check applied was a before/after comparison of failure **names**, which is byte-identical.

Each guard was revert-proofed: with `extra="forbid"` removed, `Check` fails 10 tests, `Step` 3, `BaseGenerator` 2.

## Follow-ups

- The OSS-122 rename itself — now fails loudly rather than silently
- Narrow the `Interact.inputs` union
- Changelog entry for the breaking change (OSS-127)


davidberenstein1957 (commented):
I seee this is still merge conflicted and a part of the changes overlap with https://github.com/Giskard-AI/giskard-oss/pull/2769. So waiting with the review for now.

kevinmessiaen:
@davidberenstein1957 I've rebased the PR

PR #2769, opened 2026-08-14
refactor(checks)!: name the value under test target_key on every check

kevinmessiaen:
> **Stacked on #2768** — review that first. This branch targets `feature/oss-122-forbid-unknown-fields`, so the diff shown here is only the rename. Supersedes the closed #2754, which standardised on bare `key` — the opposite direction of the convention below.

## The convention

| Role | Naming | Example |
|---|---|---|
| The value under test | `target_key` on **every** check, matching the Hub | `GreaterThan.target_key` |
| Everything else | `{static}_key`, where `{static}` is a real sibling field | `reference_text` / `reference_text_key` |

The subject key was spelled five ways — `key`, `text_key`, `actual_answer_key`, `output_key`, `answer_key` — while meaning the same thing and defaulting to `trace.last.outputs` on all ten checks.

## Rename table

Every check keeps reading its old spelling, plus a domain-meaningful name where one reads better than the generic term:

| Check | Was | Now | Also accepts |
|---|---|---|---|
| `ComparisonCheck` (+6 subclasses) | `key` | `target_key` | `actual_key`, `key` |
| `TextBasedCheck` (`StringMatching`, `RegexMatching`) | `text_key` | `target_key` | `text_key` |
| `JsonValid` | `key` | `target_key` | `value_key`, `key` |
| `Readability` | `key` | `target_key` | `text_key`, `key` |
| `RegoPolicy` | `key` | `target_key` | `input_key`, `key` |
| `SemanticSimilarity` | `actual_answer_key` | `target_key` | `answer_key`, `actual_answer_key` |
| `Toxicity` | `output_key` | `target_key` | `output_key` |
| `Groundedness` / `Contradiction` / `AnswerRelevance` | `answer_key` | `target_key` | `answer_key` |

`RegoPolicy` gets `input_key` because its subject is the OPA **input document**, not an output — three independent analysis passes each flagged it as the check that resists a purely generic name. `JsonValid` gets `value_key` because its subject is an arbitrary payload, and its `details` payload already called it `value`.

**Reference keys are untouched.** `keyword_key`, `pattern_key`, `context_key`, `question_key`, `expected_value_key` and `reference_text_key` already follow `{static}_key` with a matching sibling.

`expected_value` → `expected` was considered and rejected: 336 occurrences across 18 files for zero convention benefit, since `expected_value` / `expected_value_key` already conforms exactly.

## Back-compat

Aliases are **read-only**. An existing suite loads unchanged; serialization always emits `target_key`, so the file is rewritten to the canonical name on save. Unknown keys still raise — `extra="forbid"` from #2768 is unaffected, and was verified to coexist with `AliasChoices` before this was built.

One consequence worth knowing at review time: `validation_alias` removes old names from pydantic's generated `__init__` signature. Every `Equals(key=...)` call site therefore became a `reportCallIssue` — **241 pyright errors** — even though all of them still work at runtime. ~239 call sites were migrated. Three alias call sites were deliberately left on old spellings as live regression coverage.

## Enforcement test

`tests/core/test_key_naming_convention.py` asserts both halves of the convention, so it cannot drift back.

Rule (b) **did not hold before this rename** — `SemanticSimilarity.actual_answer_key` had no `actual_answer` sibling. That field is the subject and is now `target_key`, which rule (b) exempts, so the invariant is assertable as a real universal instead of needing a carve-out. The rule is only cleanly expressible *after* the rename it justifies.

The walk reuses the helpers from `test_jsonpath_enforcement` rather than copying them, so both enforcement tests provably cover the same class set. Guarded against passing vacuously: one test fails if the walk sees no JSONPath fields at all, another pins the six real pairs, and both rules were verified to fire against a deliberately violating subclass.

## Internal collisions resolved

- `TextBasedCheck._extract_and_validate` had locals named `target_key` / `target_value` meaning *the pattern being searched for* — the inverse of the new field's sense. Renamed to `matcher_*`.
- `judges/_inputs.py` built `details` keys as `f"{name}_key"` from the domain label, which would have emitted a stale `answer_key` naming a field that no longer exists. It now names the real field while error messages keep readable domain wording.

## Breaking change

Suite JSON is rewritten to `target_key` on save. Old spellings still load, but **code reading a check's attribute must use `target_key`** — aliases are validation-only and are not attribute names. Needs a changelog entry alongside #2768's (see OSS-127).

## Verification

| Suite | Result |
|---|---|
| giskard-checks | 973 passed, 4 skipped |
| giskard-core | 46 passed |
| giskard-scan | 169 passed, 46 skipped |
| basedpyright / ruff / format | clean across all libs |

Closes [OSS-122](https://linear.app/giskard/issue/OSS-122/unify-extraction-key-parameter-names-across-checks-before-the-api)

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import Any, Literal, Self, override
 
-from pydantic import AliasChoices, Field, model_validator
+from pydantic import Field, model_validator
 from pydantic.experimental.missing_sentinel import MISSING
 
 from ..core import Trace
@@ -37,13 +37,10 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
 
     target_key: JSONPathStr = Field(
         default="trace.last.outputs",
-        validation_alias=AliasChoices("target_key", "actual_key", "key"),
         description=(
             "JSONPath expression to extract the actual value from the trace. "
             "Defaults to 'trace.last.outputs' which extracts the last "
-            "interaction's outputs. The canonical 'target_key' also accepts "
-            "the input aliases 'actual_key' and 'key'; always serialized as "
-            "'target_key'."
+            "interaction's outputs."
         ),
     )
     expected_value: ExpectedType | MISSING = Field(
--- a/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
@@ -41,8 +41,9 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
         description=(
             "JSONPath expression to extract the actual value from the trace. "
             "Defaults to 'trace.last.outputs' which extracts the last "
-            "interaction's outputs. Also accepts 'actual_key' and the legacy "
-            "'key' on input; always serialized as 'target_key'."
+            "interaction's outputs. The canonical 'target_key' also accepts "
+            "the input aliases 'actual_key' and 'key'; always serialized as "
+            "'target_key'."
         ),
     )
     expected_value: ExpectedType | MISSING = Field(
--- a/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/comparison.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import Any, Literal, Self, override
 
-from pydantic import Field, model_validator
+from pydantic import AliasChoices, Field, model_validator
 from pydantic.experimental.missing_sentinel import MISSING
 
 from ..core import Trace
@@ -35,12 +35,14 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
     - Result formatting
     """
 
-    key: JSONPathStr = Field(
+    target_key: JSONPathStr = Field(
         default="trace.last.outputs",
+        validation_alias=AliasChoices("target_key", "actual_key", "key"),
         description=(
             "JSONPath expression to extract the actual value from the trace. "
             "Defaults to 'trace.last.outputs' which extracts the last "
-            "interaction's outputs."
+            "interaction's outputs. Also accepts 'actual_key' and the legacy "
+            "'key' on input; always serialized as 'target_key'."
         ),
     )
     expected_value: ExpectedType | MISSING = Field(
@@ -161,7 +163,7 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
         if not isinstance(actual_value, (list, set, tuple)):
             return CheckResult.error(
                 message=(
-                    f"Expected a list, set, or tuple at key '{self.key}' when match is "
+                    f"Expected a list, set, or tuple at key '{self.target_key}' when match is "
                     f"{self.match!r}, but got {type(actual_value).__name__}."
                 ),
                 details=details,
@@ -213,7 +215,7 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
     @override
     async def run(self, trace: TraceType) -> CheckResult:
         """Execute the check against the provided trace."""
-        actual_value = resolve(trace, self.key)
+        actual_value = resolve(trace, self.target_key)
         expected_value = provided_or_resolve(
             trace,
             key=self.expected_value_key,
@@ -233,7 +235,7 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
 
         if isinstance(actual_value, NoMatch):
             return CheckResult.error(
-                message=f"No value found for key '{self.key}', expected a value {self._comparison_message} {repr(self.expected_value)}.",
+                message=f"No value found for key '{self.target_key}', expected a value {self._comparison_message} {repr(self.expected_value)}.",
                 details=details,
             )
 
@@ -279,10 +281,11 @@ class LessThan[InputType, OutputType, TraceType: Trace, ExpectedType](  # pyrigh
     ----------
     expected_value : ExpectedType
         The expected value to compare against the extracted values
-    key : JSONPathStr
+    target_key : JSONPathStr
         JSONPath expression to extract the actual value from the trace.
         Defaults to "trace.last.outputs" which extracts the last
-        interaction's outputs.
+        interaction's outputs. Also accepts ``actual_key`` and the legacy
+        ``key`` on input.
     """
 
     @override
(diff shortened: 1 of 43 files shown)


libs/giskard-checks/src/giskard/checks/builtin/comparison.py
```diff
@@ -35,12 +35,14 @@ class ComparisonCheck[InputType, OutputType, TraceType: Trace, ExpectedType](  #
     - Result formatting
     """
 
-    key: JSONPathStr = Field(
+    target_key: JSONPathStr = Field(
         default="trace.last.outputs",
+        validation_alias=AliasChoices("target_key", "actual_key", "key"),
```
davidberenstein1957: do we need to redfinine target_key here?

davidberenstein1957: why do we keep the deprecated keys? can't we just do a cleann switch for the official release?

libs/giskard-checks/src/giskard/checks/builtin/comparison.py
```diff
@@ -324,10 +327,11 @@ class GreaterThan[InputType, OutputType, TraceType: Trace, ExpectedType](  # pyr
     ----------
     expected_value : ExpectedType
         The expected value to compare against the extracted values
-    key : JSONPathStr
+    target_key : JSONPathStr
         JSONPath expression to extract the actual value from the trace.
         Defaults to "trace.last.outputs" which extracts the last
-        interaction's outputs.
+        interaction's outputs. Also accepts ``actual_key`` and the legacy
```
davidberenstein1957: not sure if we should keep the "legacy" key, and if we do, we can just mention we also expcept aliases, perhaps?

libs/giskard-checks/src/giskard/checks/builtin/json_valid.py
```diff
@@ -30,9 +30,14 @@ class JsonValid[InputType, OutputType, TraceType: Trace](  # pyright: ignore[rep
 
     model_config = ConfigDict(populate_by_name=True, serialize_by_alias=True)
 
-    key: JSONPathStr = Field(
+    target_key: JSONPathStr = Field(
         default="trace.last.outputs",
-        description="JSONPath expression to extract the value to validate.",
+        validation_alias=AliasChoices("target_key", "value_key", "key"),
```
davidberenstein1957: same as above in terms of phrasing

libs/giskard-checks/src/giskard/checks/builtin/semantic_similarity.py
```diff
@@ -88,15 +89,21 @@ class SemanticSimilarity[InputType, OutputType, TraceType: Trace](  # pyright: i
...
+    target_key: JSONPathStr = Field(
         default="trace.last.outputs",
-        description="The key to extract the actual answer from the trace",
+        validation_alias=AliasChoices("target_key", "answer_key", "actual_answer_key"),
+        description=(
+            "The key to extract the actual answer from the trace. Also accepts "
+            "'answer_key' and the legacy 'actual_answer_key' on input; always "
+            "serialized as 'target_key'."
+        ),
     )
```
davidberenstein1957, line 101: is ther some kind of default way to deal with the renaming for examples with multiple key-value-pairs?

libs/giskard-checks/src/giskard/checks/builtin/semantic_similarity.py
```diff
@@ -115,58 +122,58 @@ async def run(self, trace: TraceType) -> CheckResult:
         CheckResult
             The result of the check evaluation.
         """
-        reference_text = provided_or_resolve(
+        reference = provided_or_resolve(
```
davidberenstein1957, line 119: why do we rename this and not just keep reference_text?

libs/giskard-checks/src/giskard/checks/judges/contradiction.py
```diff
@@ -18,18 +18,32 @@ class Contradiction[InputType, OutputType, TraceType: Trace](  # pyright: ignore
...
     )
-    answer_key: JSONPathStr = Field(
+    target_key: JSONPathStr = Field(
         default="trace.last.outputs",
-        description="Key to extract the answer from the trace",
+        validation_alias=AliasChoices("target_key", "answer_key"),
+        description=(
+            "Key to extract the answer from the trace. Also accepts the legacy "
+            "'answer_key' on input; always serialized as 'target_key'."
+        ),
```
davidberenstein1957: is there a way we can cleanly unify the key overview and naming in some way, since we re-use it in a lot of places and it seems to differe slightly throughout the cod.e

libs/giskard-checks/tests/builtin/test_answer_relevance.py
```diff
@@ -291,7 +291,7 @@ async def test_custom_keys(self):
         check = AnswerRelevance(
             generator=generator,
             question_key="trace.interactions[0].inputs.query",
-            answer_key="trace.interactions[0].outputs.response",
+            target_key="trace.interactions[0].outputs.response",
```
davidberenstein1957, line 294: should we also unify trace.interactions[0].outputs.response to trace.last.outputs.response? also, perhaps our docs should have a dedicated section on trace keys?

libs/giskard-checks/tests/builtin/test_comparison.py
```diff
@@ -521,7 +521,7 @@ async def test_typeerror_missing_method(self):
 
     def test_serialises_with_greater_than_equals_kind(self):
         """Serialized kind remains greater_than_equals and round-trips."""
-        check = GreaterThanEquals(expected_value=10, key="trace.last.outputs")
+        check = GreaterThanEquals(expected_value=10, target_key="trace.last.outputs")
```
davidberenstein1957: do we still need to define this explicitely? we default to thses values now, right?

libs/giskard-checks/tests/builtin/test_comparison.py
```diff
@@ -714,7 +714,7 @@ async def test_float_not_equals_success(self):
         trace = await Trace.from_interactions(Interaction(inputs="test", outputs=3.14))
         check = NotEquals(
             expected_value=5.0,
-            key="trace.interactions[-1].outputs",
+            target_key="trace.interactions[-1].outputs",
```
davidberenstein1957: also here, do we still need and want to push trace.interactions[-1].outputs? Perhaps we can just simplify and default to trace.last.outputs, and create one single test to verify if trace.interactions[-1].outputs still works?

libs/giskard-checks/tests/core/test_key_naming_convention.py
```diff
@@ -0,0 +1,188 @@
...
+    ("field", "static"),
+    [
+        ("reference_text_key", "reference_text"),
+        ("keyword_key", "keyword"),
+        ("pattern_key", "pattern"),
+        ("context_key", "context"),
+        ("question_key", "question"),
+        ("expected_value_key", "expected_value"),
+    ],
+)
```
davidberenstein1957: this seems a bit manual and might be resolved by a unified definition as I proposed in another part of the review. we could then simply import that here? Anyway not sure if it is needed but it should be cleaner to do this upstream IMO.

davidberenstein1957 (commented):
Hi,

LFTM but left some comments regarding code simplicity. They are not required but IMO it should clean a lot of code and make it more simpler to evaluate in the long run. Given the v3 release, it would now be a time to fix it potentially.

libs/giskard-checks/src/giskard/checks/builtin/semantic_similarity.py
kevinmessiaen: > why do we rename this and not just keep reference_text?

`reference_text` is the static comparison value, while this path selects the value under test from the trace. `target_key` makes that role consistent across checks; `reference_text_key` remains the path for the separate reference source.

libs/giskard-checks/tests/builtin/test_answer_relevance.py
kevinmessiaen: We test a different path on purpose here

PR #2786, opened 2026-08-21
fix(checks): reject empty text matchers 🤖🤖🤖🤖

### LINKED ISSUE #2785: StringMatching and RegexMatching pass unconditionally when the keyword or pattern is empty
## Summary

`StringMatching` with an empty or whitespace-only `keyword`, and `RegexMatching` with an empty `pattern`, return **PASS** against any text. The check reports success without testing anything.

I'm raising this as an issue rather than a PR because two existing tests assert the current behaviour, so this is a design question for you rather than an obvious defect.

## Reproduction

```python
from giskard.checks import StringMatching, RegexMatching, Trace

await StringMatching(text="The capital of France is Paris.", keyword="").run(Trace())
# CheckStatus.PASS  -- "The answer contains the keyword ''."

await StringMatching(text="The capital of France is Paris.", keyword="   ").run(Trace())
# CheckStatus.PASS  -- whitespace normalisation reduces "   " to ""

await StringMatching(text="The capital of France is Paris.", keyword="\n").run(Trace())
# CheckStatus.PASS

await RegexMatching(text="anything at all", pattern="").run(Trace())
# CheckStatus.PASS  -- an empty regex matches everything
```

`""` is a substring of every string and an empty regex matches everything, so these are Python's semantics working as designed. The question is whether a check framework should surface that as a green result.

## Why it matters in practice

Nobody writes `keyword=""` on purpose. But `keyword_key` and `pattern_key` resolve the matcher from a trace field:

```python
StringMatching(target_key="trace.last.outputs.answer",
               keyword_key="trace.last.inputs.expected_keyword")
```

If `expected_keyword` is blank or whitespace in a row of test data, that check passes unconditionally and silently. In a suite of hundreds, a vacuously green check is indistinguishable from a real one, and it's the failure mode you least want in a testing tool: it reports assurance it did not earn.

The two checks also disagree on whitespace. `StringMatching(keyword="   ")` passes because it normalises whitespace before matching; `RegexMatching(pattern="   ")` fails because it does not.

## Suggested behaviour

Return `CheckStatus.ERROR` for an empty matcher, consistent with how `_extract_and_validate` already treats a non-string matcher or an unresolvable key — both configuration problems rather than assertion outcomes.

I'd leave `RegexMatching(pattern="   ")` as a genuine FAIL, since matching literal spaces is a legitimate intent.

## Existing tests that assert the current behaviour

```python
# tests/builtin/test_string_matching.py
async def test_empty_keyword() -> None:
    check = StringMatching(text="Hello", keyword="")
    result = await check.run(Trace())
    # Empty string should be found in any text
    assert result.status == CheckStatus.PASS

# tests/builtin/test_regex_matching.py
async def test_empty_pattern_regex_mode() -> None:
    ...
    # Empty regex matches any string
    assert result.status == CheckStatus.PASS
```

These read to me like characterisation tests pinning observed behaviour rather than a deliberate product decision, but you'd know. If it was deliberate, feel free to close this.

## Branch ready if you want it

Implemented, with the two tests above updated and a parametrised case added for whitespace-only keywords:

https://github.com/WatchTree-19/giskard-oss/tree/proposal-empty-matcher-vacuous-pass

All 64 tests in `test_string_matching.py` and `test_regex_matching.py` pass on it, and the wider `tests/builtin` suite is unchanged apart from those two assertions. Two pre-existing failures in `test_rego_policy.py` are unrelated — they fail on a clean checkout too.

Happy to open it as a PR if you'd like it, or to drop it.

--

Contributing as WatchTree-19; you merged my Gemini `finish_reason` enum-value fix a while back.

mikemikimike:
## Description

Reject empty keyword and pattern values before text matching. This prevents Python substring matching and regex search from treating empty matchers as universal matches. The shared validation path covers both direct values and values resolved from a trace.

## Related Issue

Fixes #2785

## Type of Change

## Coding agents

I have read `AUTONOMOUS.md` and followed its PR requirements.

## Checklist

- [x] I've written tests for the changed behavior.
- [x] I've written the docstring in NumPy format for modified methods and classes where applicable.
- [x] No lockfile update is needed because `pyproject.toml` was not modified.

## Validation

- `uv run --frozen --directory libs/giskard-checks pytest tests/builtin/test_string_matching.py tests/builtin/test_regex_matching.py -m "not functional" -q` (63 passed)
- `uv run --frozen ruff format --check .`
- `uv run --frozen ruff check .`
- `uv tool run basedpyright --level error libs/giskard-checks/src libs/giskard-checks/tests`
- `git diff --check`

The full `giskard-checks` unit suite ran 935 passed / 16 skipped / 3 failed. The failures are existing RegoPolicy tests that require `celine-regorus`; that optional dependency intentionally has no Windows wheel, so it cannot be installed on this host.

### DIFF
--- a/libs/giskard-checks/src/giskard/checks/builtin/text_matching.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/text_matching.py
@@ -109,6 +109,16 @@ class TextBasedCheck[InputType, OutputType, TraceType: Trace](  # pyright: ignor
                 ),
             )
 
+        if matcher == "":
+            return (
+                None,
+                None,
+                CheckResult.error(
+                    message=f"Value for {matcher_name} must not be empty.",
+                    details=details,
+                ),
+            )
+
         # Validate text
         if isinstance(text, NoMatch):
             return (
--- a/libs/giskard-checks/src/giskard/checks/builtin/text_matching.py
+++ b/libs/giskard-checks/src/giskard/checks/builtin/text_matching.py
@@ -109,12 +109,12 @@ class TextBasedCheck[InputType, OutputType, TraceType: Trace](  # pyright: ignor
                 ),
             )
 
-        if matcher == "":
+        if not matcher.strip():
             return (
                 None,
                 None,
                 CheckResult.error(
-                    message=f"Value for {matcher_name} must not be empty.",
+                    message=f"Value for {matcher_name} must not be empty or whitespace-only.",
                     details=details,
                 ),
             )
--- a/libs/giskard-checks/tests/builtin/test_regex_matching.py
+++ b/libs/giskard-checks/tests/builtin/test_regex_matching.py
@@ -303,6 +303,15 @@ async def test_empty_pattern_from_trace_returns_error() -> None:
     assert "pattern must not be empty" in result.message.lower()
 
 
+async def test_whitespace_only_pattern_returns_error() -> None:
+    """Test that a whitespace-only regex pattern is rejected."""
+    check = RegexMatching(text="The capital of France is Paris.", pattern=" ")
+    result = await check.run(Trace())
+    assert result.status == CheckStatus.ERROR
+    assert result.message is not None
+    assert "pattern must not be empty" in result.message.lower()
+
+
 async def test_missing_pattern_validation() -> None:
     """Test that either pattern or pattern_key must be provided."""
     with pytest.raises(ValueError, match="pattern"):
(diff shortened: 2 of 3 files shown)


kevinmessiaen:
Hey @mikemikimike 

Thanks for the PR, it only partially fixes #2785

The case with empty string is fine, it now raises. However the following case still pass where it should fail:

`StringMatching(text="The capital of France is Paris.", keyword="   ")`

mikemikimike:
Addressed in the current head `6a00dd92`: validation now rejects whitespace-only matchers via `matcher.strip()`, not just empty strings. Added regression coverage for `StringMatching(keyword=" ")` (and the equivalent regex case); both return an error with the expected message.

Issue #2735, opened 2026-08-12
detect-secrets is not enforced in CI and the baseline is unaudited


kevinmessiaen:
`.secrets.baseline` contains **21 entries across 5 files, every one with** `is_secret: null` — meaning nobody has triaged whether any of them is a real secret. The baseline was generated 2026-05-22 and has not been audited since.

Compounding it: `detect-secrets` runs only via pre-commit, which is installed locally (`Makefile:25-30`) but **never enforced in CI**. So the hook is advisory — a contributor who skips it, or who has not run `make install`, gets no gate at all.

## Why it matters for a 1.0

A baseline of unaudited entries is indistinguishable from a baseline of accepted false positives. If a real credential were among them it would be silently allowlisted, and the tool would keep reporting clean.

## Suggested actions

1. Audit the 21 entries (`detect-secrets audit .secrets.baseline`) so each is explicitly marked true/false. This is the substantive half.
2. Add a pre-commit run to CI so the hook is actually a gate. `pre-commit/action` or a plain `uv run pre-commit run --all-files` step.
3. Consider whether the baseline should be regenerated after the audit.

## Note

`pip-audit` is separately clean and **is** wired into `make check` via the `security` target — this issue is specifically about the secrets hook, not dependency CVEs.

PR #2808, opened 2026-08-29
ci: enforce detect-secrets in CI 🤖🤖🤖🤖

CharoenwitKunna:
## Summary

- Add pinned `detect-secrets` tooling and a reusable `check-secrets` Make target.
- Include the target in `make check`, which is the existing CI lint gate.
- Regenerate the baseline with the audited documentation false positive in `AGENTS.md`.

All existing findings are intentional test fixtures or environment-variable references; intentional exceptions use `pragma: allowlist secret`.

## Validation

- `uv tool run detect-secrets==1.5.0 scan --baseline .secrets.baseline`
- `git diff --check`

Full Make targets could not run locally because `make` is unavailable in the environment.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -133,6 +133,15 @@
     }
   ],
   "results": {
+    "AGENTS.md": [
+      {
+        "type": "Secret Keyword",
+        "filename": "AGENTS.md",
+        "hashed_secret": "eb071a8890bb7296cd433abd885cb1fa814b3dea",
+        "is_verified": false,
+        "line_number": 59
+      }
+    ],
     "libs/giskard-llm/tests/functional/test_completion.py": [
       {
         "type": "Secret Keyword",
@@ -291,5 +300,5 @@
       }
     ]
   },
-  "generated_at": "2026-05-22T08:27:18Z"
+  "generated_at": "2026-08-29T15:12:01Z"
 }
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,11 @@ install: ## Install project dependencies
 # `make check` red on a tree nobody touched. Pinned for the same reason as
 # LICENSECHECK_VERSION below.
 RUFF_VERSION := 0.16.1
+DETECT_SECRETS_VERSION := 1.5.0
 
 install-tools: ## Install development tools
 	uv tool install ruff==$(RUFF_VERSION)
+	uv tool install detect-secrets==$(DETECT_SECRETS_VERSION)
 	uv tool install vermin
 	uv tool install basedpyright
 	uv tool install pre-commit --with pre-commit-uv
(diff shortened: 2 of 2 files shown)


Santoshkumarpuppala:
I ran the command this adds, and I don't think the target can fail — which would make `make check` gain a step that never turns CI red.

**Measured.** In a git repo with an unbaselined AWS secret key committed, using this PR's exact command and pin:

```
$ uv tool run detect-secrets==1.5.0 scan --baseline .secrets.baseline
EXIT CODE = 0
```

The baseline was **rewritten in place** and the planted key was **absorbed into it**, both entries carrying `is_verified: false` and no `is_secret` field at all.

**Control, so that exit 0 means something.** The same tool on the same file finds it: `detect-secrets scan planted.py` returns 2 findings (`Base64 High Entropy String`, `Secret Keyword`). So the zero is "the subcommand's contract is to update the baseline, not to report", not "the scanner missed it".

(First attempt at this measurement was wrong and worth flagging so you can reproduce cleanly: I ran it in a bare directory, the scan walked no git-tracked files, and the baseline went 21 entries → 0. It has to be a git repo with the files committed.)

**Two consequences.**

*Scope.* `make check` runs at `.github/workflows/ci.yml:33` and again at `.github/workflows/release.yml:102`, so the inert step lands in the release path too.

*It reopens the hole #2735 was filed to close.* That issue's first ask is an audit of the baseline — `scan --baseline` answers it by appending unaudited entries. `.secrets.baseline` on main has 21 entries across 5 files and **not one carries `is_secret`**; the command that this target runs is the one that keeps it that way.

Nothing here is about `AGENTS.md:59` — that line genuinely is prose (`Lint/test/build/run all work without secrets: make lint...`) and baselining it is correct. The problem is that the mechanism returns the same exit 0 for that harmless case and for a real credential.

**The fix keeps your Makefile shape and changes one word** — `scan` → `detect-secrets-hook`:

```make
check-secrets: ## Check for secrets not covered by the audited baseline
	uv tool run --from detect-secrets==$(DETECT_SECRETS_VERSION) \
	  detect-secrets-hook --baseline .secrets.baseline $$(git ls-files)
```

Control-tested both directions, baseline byte-identical after each run:

| tree | exit |
|---|---|
| secret present | **1** |
| secret removed | **0** |

It's also the entry point `.pre-commit-config.yaml:34-39` already pins at `v1.5.0`, so the Makefile target and the pre-commit hook would finally run the same thing rather than diverging.

One caveat I hit while testing, so it doesn't surprise you in CI: `detect-secrets-hook` exits **3** when the baseline is stale relative to the tree and it updates it. That's arguably the right signal, but it is a third non-zero code to expect if the baseline ever drifts.

PR #2823, opened 2026-09-09
fix(agents): preserve api_base and forbid unknown GenerationParams 🤖🤖🤖🤖

### LINKED ISSUE #2802: GenerationParams silently drops unsupported fields (e.g. api_base) instead of raising
### Description

`GenerationParams` (used by both `LiteLLMGenerator` and `GiskardLLMGenerator`) has no `api_base` field. This makes it impossible to point a generator at a custom OpenAI-compatible endpoint (for example a local LM Studio or Ollama server) without setting a provider-specific environment variable (e.g. `LM_STUDIO_API_BASE`, `OPENAI_BASE_URL`).

Passing `api_base` does not raise an error, it is silently dropped, which makes the problem hard to notice.

### Steps to reproduce

```python
from giskard.agents.generators import LiteLLMGenerator

g = LiteLLMGenerator(model="openai/qwen2.5-coder-0.5b-instruct")
g2 = g.with_params(api_base="http://127.0.0.1:1234/v1")

print(g2.params.model_dump(exclude={"tools"}, exclude_unset=True))
# {}  <- api_base silently disappears
```

`LiteLLMGenerator._call_model` builds `wire_params` from `params.model_dump(exclude={"tools"}, exclude_unset=True)` and forwards it to `litellm.acompletion(messages=..., model=self.model, **wire_params)`. Since `GenerationParams` does not declare `api_base` and its pydantic config uses the default `extra="ignore"`, any `api_base` passed via `.with_params(...)` or the constructor is dropped before it reaches `litellm.acompletion`, even though `litellm.acompletion` accepts `api_base` natively. Same issue applies to `GiskardLLMGenerator` (`Generator`), which builds `wire_params` the same way before calling `giskard.llm.acompletion`.

### Expected behavior

Either:
- `GenerationParams` gains an optional `api_base: str | None` field that is forwarded to the underlying completion call, or
- passing an unsupported field raises a validation error instead of silently dropping it, so the mistake is caught immediately.

### Actual behavior

`api_base` is accepted by `.with_params(**kwargs)` (no error) but silently discarded, so the request goes out with no custom base URL and either falls back to the provider's default endpoint or fails with an opaque error depending on the provider.

### Workaround

Currently the only way to point at a custom endpoint is a provider-specific env var (e.g. `LM_STUDIO_API_BASE`), or subclassing `GenerationParams` to declare the field yourself:

```python
from giskard.agents.generators._types import GenerationParams

class GenerationParamsWithApiBase(GenerationParams):
    api_base: str | None = None

generator = LiteLLMGenerator(model="openai/qwen2.5-coder-0.5b-instruct")
generator.params = GenerationParamsWithApiBase(api_base="http://127.0.0.1:1234/v1")
```

### Environment

- `giskard-agents` 1.0.2
- `giskard-core` 1.0.1
- `giskard-checks` 1.0.3
- `giskard-llm` 1.0.0
- `giskard` 3.0.0
- `litellm` 1.98.0
- `pydantic` 2.13.4
- Python 3.14.2, macOS (Darwin 25.5.0)

siddiqueirshad:
## Description

`GenerationParams` previously used pydantic's default `extra="ignore"`, so fields like `api_base` passed through `.with_params(...)` or construction were silently discarded before they could reach `litellm.acompletion` / `giskard.llm.acompletion`. That made custom OpenAI-compatible endpoints (LM Studio, Ollama, etc.) fail without a clear signal.

This change:

- Adds optional `api_base` so supported backends receive the custom endpoint via `model_dump(..., exclude_unset=True)`.
- Sets `extra="forbid"` on `GenerationParams`, matching `BaseGenerator` / middleware, so typos and unsupported kwargs raise `ValidationError`.
- Updates `BaseGenerator.with_params` to reject unknown kwargs before applying updates.

## Related Issue

Closes #2802

## Type of Change

## Coding agents

Read [AUTONOMOUS.md](https://github.com/Giskard-AI/giskard-oss/blob/main/AUTONOMOUS.md). Title includes the required agent marker.

## Checklist

- [x] I've updated the `uv.lock` running `uv lock` (only applicable when `pyproject.toml` has been modified) — N/A, no dependency changes.

## Verification

- `uv run --directory libs/giskard-agents python -m pytest tests src -m "not functional" -q` → **151 passed, 2 skipped, 16 deselected**
- `uv tool run ruff==0.16.1 format` / `check` on touched files → clean
- `uv tool run basedpyright --level error` on touched source files → 0 errors
- Manual repro: `.with_params(api_base=...)` now dumps `api_base`; `.with_params(typo_param=1)` raises `ValidationError`

### DIFF
--- a/libs/giskard-agents/src/giskard/agents/generators/_types.py
+++ b/libs/giskard-agents/src/giskard/agents/generators/_types.py
@@ -4,7 +4,9 @@ Kept in a separate module to break the circular dependency between
 ``base`` (BaseGenerator) and ``middleware`` (CompletionMiddleware).
 """
 
-from pydantic import BaseModel, Field
+from typing import ClassVar
+
+from pydantic import BaseModel, ConfigDict, Field
 
 from ..tools import Tool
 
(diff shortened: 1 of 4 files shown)


siddiqueirshad:
Hi @lansolo99, when you get a chance, could you please take a look at this PR? Thanks!

